PR #5, opened 2015-12-22
Jhbate sprint25

jh-bate:
@darinkrauss - is there a neater way to do this? the merge is too large


darinkrauss:
Try two branches.

First branch is for only the Godeps directory. Do a PR for that based upon master (just did the merge into master).

Second branch is for all the other changes. Do a PR for that based upon the first branch (not master).

Will that work?

jh-bate:
thanks @darinkrauss I will try that

PR #6, opened 2016-03-15
Cleanup Makefile and general project setup

darinkrauss:
@jh-bate


.travis.yml
```diff
@@ -1,8 +1,5 @@
 language: go
 
-services:
-  - mongodb
```
jh-bate: so pretty sure this is needed

jh-bate:
one query other than that LGTM

.travis.yml
darinkrauss: It is, but my earlier integration added a second copy of it, so this deletes the extra copy. :)

jh-bate: ahh - sweet then.

PR #9, opened 2016-03-22
enable query params on get

jh-bate:
@darinkrauss essentially the tide-whisperer replacement. I was doing some of the store changes prior so thought I would put this up there as an initial implementation. 

Note: I really need to reset the mongo collection between each test run and hence there are a couple of "loose" tests that I am not happy with yet.


data/base.go
```diff
@@ -69,19 +69,22 @@ const (
 	//UserIDField is the userID
 	UserIDField = "userId"
 	//GroupIDField id the groupID
-	GroupIDField = "groupId"
-
-	deviceIDField = "deviceId"
-	uploadIDField = "uploadId"
-
+	GroupIDField         = "groupId"
```
darinkrauss: I would make two sets of constants if going this way, one for the URL parameters and one for the data model.  We don't want to mix-n-match in case we ever need to change one.  They are for two different things.

dataservices/dataservices.go
```diff
@@ -175,6 +176,30 @@ func process(iter store.Iterator) data.Dataset {
 	return all
 }
 
+func buildQuery(params url.Values) store.Query {
+
+	types := strings.Split(params.Get("type"), ",")
```
darinkrauss: Are we using constants here or not?

dataservices/dataservices.go
```diff
@@ -175,6 +176,30 @@ func process(iter store.Iterator) data.Dataset {
...
+	subTypes := strings.Split(params.Get("subType"), ",")
```
darinkrauss: As noted in previous PR, use `type`, `subType`, `startDate` and `endDate` for variable names.

dataservices/dataservices.go
darinkrauss: Too bad we don't have some sort of Compact on the array to remove nils and whitespace-only and empty strings.

store/store.go
```diff
@@ -20,11 +20,17 @@ var (
 type Store interface {
 	Save(d interface{}) error
 	Update(selector interface{}, d interface{}) error
-	Delete(find Fields) error
-	Read(find Fields, remove Fields, result interface{}) error
-	ReadAll(find Fields, remove Fields) Iterator
+	Delete(id Field) error
```
darinkrauss: For any of these `Fields` we should have the interface be as specific as possible.  For example, I suspect id will always be a string, so that's what the argument type should be.

darinkrauss:
In general, looks good, but I assume all of those other comments from #8.

PR #12, opened 2016-03-23
All data-validation changes (PR only)

darinkrauss:


darinkrauss:
Might want to run `make pre-commit` as there are a couple of extra imports.

_config/dataservices.json
```diff
@@ -0,0 +1,3 @@
+{
```
darinkrauss: For the configs that do NOT have secrets in them, I recommend we put all environment configuration in the code repository (rather than the config repository, which would only be used for secrets). This would mean that config changes would be rare.  So, the structure of a json config file with environment specific settings might be:

```
{
  "dev": {
    "port": ":8077"
  },
  "stg": {
    "port": "whatever"
  },
  ...
}
```

The config reading code could then be "smart" about choosing the right configuration based upon the environment name.  We could also have an implicit or explicit "global" or "common" area which would apply to all environments.

Thoughts?

config/config.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/jinzhu/configor"
+)
+
+//FromJSON will load the requested JSON config into the given config interface
```
darinkrauss: Let's bail on all package/function level comments unless it is absolutely needed.  I believe I've removed those from the code quality reporting in the Makefile.

config/config.go
```diff
@@ -0,0 +1,37 @@
...
+func FromJSON(config interface{}, name string) error {
```
darinkrauss: This code will likely change with my comments about per-environment configuration in each config file above.

config/config.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// we need to get the absolute path and also test it because of the
+	// difference of how this code runs in different environments
+	const configPath = "./_config/"
+	absPath, _ := filepath.Abs(configPath + name)
+	_, err := os.Open(absPath)
```
darinkrauss: Can we use `configor.Load` to implicitly test existence of the file?  Or perhaps `if _, err := os.Stat("/path/to/whatever"); os.IsNotExist(err) {`. If we want to stick with `os.Open` we should use the *File returned and close it (otherwise it is a leaked open file).

config/config.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		const configLibPath = "./../_config/"
+		absPath, _ = filepath.Abs(configLibPath + name)
+	}
+
+	return configor.Load(&config, absPath)
+}
+
+//FromEnv will load the requested config value
+func FromEnv(name string) (string, error) {
```
darinkrauss: So, this only loads a string config from env? Do we have an JSON structs that are defined in the ENV?  (Probably not, but I know other repos have them like Jellyfish, for example.)

config/config_test.go
```diff
@@ -0,0 +1,53 @@
+package config_test
+
+import (
+	"os"
+
+	. "github.com/tidepool-org/platform/config"
```
darinkrauss: If we are going to use a different package (config_test), then I think we should NOT use the dot import. We want to show we are explicitly testing the `config` package. While I think the dot imports make a limited sense for ginkgo and gomega, I think they should be strongly avoided for our code.  For example, it isn't immediately clear that the code:

```
err := FromJSON(&random, "random.json")
```

is actually testing the config package.  However, if it were:

```
err := config.FromJSON(&random, "random.json")
```

then it would.

config/config_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	var _ = Describe("FromJson", func() {
+
+		It("should load the given config file", func() {
+			var mgoConfig store.MongoConfig
+			FromJSON(&mgoConfig, "mongo.json")
+			Expect(mgoConfig).To(Not(BeNil()))
+			Expect(mgoConfig.URL).To(Not(BeEmpty()))
+		})
+		It("should error if the config doen't exist", func() {
```
darinkrauss: Let's use a blank line between each test for clarity.

config/config.go
jh-bate: yeah - but not before I had commented every single one :)

config/config_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/ginkgo"
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/gomega"
+)
+
+var _ = Describe("Config", func() {
+
+	var _ = Describe("FromJson", func() {
+
+		It("should load the given config file", func() {
```
darinkrauss: Let's leave off the "should".  So maybe, "loads the given config file" or "returns an error if the config file does not exist".

config/config_test.go
```diff
@@ -0,0 +1,53 @@
...
+			var mgoConfig store.MongoConfig
```
darinkrauss: No abbreviations, right?  :)  So, either `mongoConfig` or just `config`.

config/config_test.go
```diff
@@ -0,0 +1,53 @@
...
+		})
+		It("should error if the config doen't exist", func() {
+			var random interface{}
+			err := FromJSON(&random, "random.json")
+			Expect(random).To(BeNil())
+			Expect(err).ToNot(BeNil())
+		})
+	})
+
+	var _ = Describe("FromEnv", func() {
```
darinkrauss: Do we need the `var _ = ` here for it to work?  (I don't know.)  If not, let's get rid of it.

config/config_test.go
jh-bate: yeah the dot import is the default with ginkgo and gomega but I have the same feeling. On one of the PR's I changed this in the data repo and asked your opinion on it. I am in favour of changing

config/config_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+		It("should load the given config value from env", func() {
+			const platformKey, platformValue = "CONFIG_TEST", "yay I exist!"
+			os.Setenv(platformKey, platformValue)
+
+			platfromValue, _ := FromEnv(platformKey)
```
darinkrauss: platform (not platfrom). Also, should test err return value for expectations as well.

config/config_test.go
```diff
@@ -0,0 +1,53 @@
...
+			Expect(platfromValue).To(Equal(platformValue))
+
+			os.Unsetenv(platformKey)
+		})
+
+		It("should error if the value doesn't exist", func() {
+			const otherKey = "OTHER"
+			os.Unsetenv(otherKey) // make sure it doesn't exist
+
+			_, err := FromEnv("OTHER")
```
darinkrauss: Perhaps use `otherKey` rather than "OTHER"?  (Or just use "OTHER" for both).

Also, should validate expectations for the value returned is an empty string.

config/config.go
jh-bate: None at the moment - I thought I would just start with what I needed

data/Builder.go
```diff
@@ -0,0 +1,110 @@
+package data
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+//Datum represent one data point
```
darinkrauss: Ditch all of the comments.

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+type Datum map[string]interface{}
+
+//Dataset represents an array of data points
+type Dataset []Datum
+
+//GetSelector returns the selector fields for a given platform datatype, or nil if there is no match
+func GetSelector(data interface{}) interface{} {
```
darinkrauss: Is this used?  If not, delete it.

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+	"errors"
+	"fmt"
+	"strings"
+)
+
+//Datum represent one data point
+type Datum map[string]interface{}
+
+//Dataset represents an array of data points
+type Dataset []Datum
```
darinkrauss: I would call this either `Data`, `DataArray`, or `DataList`.  Data is the plural of Datum, after all.  Also, it isn't an Set (which is unordered and guarantees uniqueness).

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+	case *DeviceEvent:
+		return data.(*DeviceEvent).Selector()
+	default:
+		return nil
+	}
+}
+
+//Builder interface that the TypeBuilder implements
+type Builder interface {
+	BuildFromRaw(raw []byte) (interface{}, *Error)
```
darinkrauss: BuildFromBytes?

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+	BuildFromData(data map[string]interface{}) (interface{}, *Error)
```
darinkrauss: Use `Datum` as type rather than `map[string]interface{}`.  What does this return, BTW, I see it returns an interface{} but should we have a real type for that?

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+
+//NewTypeBuilder returns an instance of TypeBuilder
+func NewTypeBuilder(inject map[string]interface{}) Builder {
+	return &TypeBuilder{
+		inject: inject,
+	}
+}
+
+//BuildFromDataSet will build the matching type(s) from the given Dataset
+func (typeBuilder *TypeBuilder) BuildFromDataSet(dataSet Dataset) ([]interface{}, *ErrorSet) {
```
darinkrauss: I would order these function in the same order they appear in the interface above.

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+	for i := range dataSet {
+		item, err := typeBuilder.BuildFromData(dataSet[i])
+		if err != nil && !err.IsEmpty() {
+			if buildError == nil {
+				buildError = NewErrorSet()
+			}
+			buildError.AppendError(err)
+			continue
+		}
+		set = append(set, item)
```
darinkrauss: If we have an error, then an item will not be placed into the final array and the incoming array and outgoing array will not be coordinated.  What if we actually just stored the error back into the array after building.  So, the outgoing array would be an array of both built data as well as errors?

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return set, buildError
+}
+
+//BuildFromRaw will build the matching type(s) from the given raw data
+func (typeBuilder *TypeBuilder) BuildFromRaw(raw []byte) (interface{}, *Error) {
+
+	var data map[string]interface{}
+
+	if err := json.NewDecoder(strings.NewReader(string(raw))).Decode(&data); err != nil {
```
darinkrauss: Maybe just use `bytes.NewReader(raw)` rather than `strings.NewReader(string(raw))`?

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+
+//BuildFromDataSet will build the matching type(s) from the given Dataset
+func (typeBuilder *TypeBuilder) BuildFromDataSet(dataSet Dataset) ([]interface{}, *ErrorSet) {
+
+	var set []interface{}
+	var buildError *ErrorSet
+
+	for i := range dataSet {
+		item, err := typeBuilder.BuildFromData(dataSet[i])
+		if err != nil && !err.IsEmpty() {
```
darinkrauss: Should we ever be getting an empty error here?

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+		typeField = "type"
+	)
+
+	if data[typeField] != nil {
+
+		for k, v := range typeBuilder.inject {
+			data[k] = v
+		}
+
+		if strings.ToLower(data[typeField].(string)) == strings.ToLower(BasalName) {
```
darinkrauss: Rather than a switch statement we should consider a factory where each data type can register itself in an `init()` function.  That way, this would remain entirely generic and could support any type registered.  In fact, I can imagine a registered type whose build function actually delegates to another based upon sub-type.

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+			return BuildBasal(data)
+		} else if strings.ToLower(data[typeField].(string)) == strings.ToLower(DeviceEventName) {
+			return BuildDeviceEvent(data)
+		}
+		e := NewError(data)
+		e.AppendError(fmt.Errorf("we can't deal with `type`=%s", data[typeField].(string)))
```
darinkrauss: We should coordinate on error messages since these will technically be part of the official open API.

config/config_test.go
jh-bate: but its for the `mgo` package :) yeah sure thing

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+//BuildFromData will build the matching type from the given raw data
+func (typeBuilder *TypeBuilder) BuildFromData(data map[string]interface{}) (interface{}, *Error) {
+
+	const (
+		typeField = "type"
+	)
+
+	if data[typeField] != nil {
```
darinkrauss: Might want to do the type field conversion to a string at this level so someone who passes a type field that is a boolean fails properly.

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+
+		for k, v := range typeBuilder.inject {
```
darinkrauss: Not sure I understand this yet. (Maybe I need to read more of the code.) What are we injecting here?  Is it okay to inject and *replace* what is currently in the data?

data/DataError.go
```diff
@@ -0,0 +1,88 @@
+package data
+
+import "fmt"
+
+//ErrorSet contains a list of Error's
+type ErrorSet struct {
```
darinkrauss: Not Set. It is an ordered array or list.

data/DataError.go
```diff
@@ -0,0 +1,88 @@
...
+	return &ErrorSet{}
+}
+
+//AppendError appends an error to the ErrorSet
+func (e *ErrorSet) AppendError(err *Error) {
+	if err == nil || err.IsEmpty() {
+		return
+	}
+	e.errors = append(e.errors, err)
+	return
```
darinkrauss: Don't need return here.  In fact, I'd change this to:

```
if err != nil && !err.IsEmpty() {
  e.errors = append(e.errors, err)
}
```

data/DataError.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+//Error returns a string representation of the errors
+func (e *ErrorSet) Error() string {
+
+	errorsStr := ""
+	if e != nil {
```
darinkrauss: I don't think `e` can be nil. It is the struct itself.

data/DataError.go
```diff
@@ -0,0 +1,88 @@
...
+		for i := range e.errors {
+			if e.errors[i] != nil {
```
darinkrauss: Can an individual error ever be nil?

data/DataError.go
```diff
@@ -0,0 +1,88 @@
...
+				errorsStr = fmt.Sprintln(errorsStr, e.errors[i].Error())
```
darinkrauss: Do we want this on multiple lines?  Semi-colon delimited, single line?  Particularly for logging?

data/DataError.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+
+	return errorsStr
+}
+
+//Error contains a list or errors and asscociated data
+type Error struct {
+	errors []error
+	data   map[string]interface{}
```
darinkrauss: What do we need the data here for?  Is the array for the potential multiple validation errors?

data/DataError.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+//NewError creates and instance of Error
+func NewError(data map[string]interface{}) *Error {
+	return &Error{data: data}
+}
+
+//AppendError appends an error to the Error
+func (e *Error) AppendError(err error) {
```
darinkrauss: As above:

```
if err != nil {
  e.errors = append(e.errors, err)
}
```

data/Builder.go
jh-bate: Cool - yeah I spent ages coming up with crap names so just left it like this but happy to change it .... did remember one thing that I think pushed me down this path though `data.Data` would be how its referenced

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+			buildError.AppendError(err)
+			continue
+		}
+		set = append(set, item)
+	}
+	return set, buildError
+}
+
+//BuildFromRaw will build the matching type(s) from the given raw data
+func (typeBuilder *TypeBuilder) BuildFromRaw(raw []byte) (interface{}, *Error) {
```
darinkrauss: I notice in other places you use a single letter ('e' for *Error below) for the struct "self". Here you are using 'typeBuilder'.  We should standardize on one code style or the other. I'm fine with single letter (even though it technically violates the "self-commenting code", but given the circumstances I think it is acceptable.

data/DataError.go
```diff
@@ -0,0 +1,88 @@
...
+	if err == nil {
+		return
+	}
+	e.errors = append(e.errors, err)
+	return
+}
+
+//AppendFieldError appends a error and associated field name
+func (e *Error) AppendFieldError(name string, detail interface{}) {
```
darinkrauss: Do we want 'generic' errors on fields?  Or perhaps more specific & descriptive errors.  If so, then perhaps this generic field error should go away?

data/DataError.go
```diff
@@ -0,0 +1,88 @@
...
+
+	e.errors = append(e.errors,
+		fmt.Errorf("encountered an error on type field %s when given %v ", name, detail),
+	)
+	return
+}
+
+//IsEmpty return true if there are no errors contained
+func (e *Error) IsEmpty() bool {
+	return len(e.errors) == 0
```
darinkrauss: Will we ever have this?

config/config.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/jinzhu/configor"
+)
+
+//FromJSON will load the requested JSON config into the given config interface
+func FromJSON(config interface{}, name string) error {
+
+	// we need to get the absolute path and also test it because of the
+	// difference of how this code runs in different environments
+	const configPath = "./_config/"
+	absPath, _ := filepath.Abs(configPath + name)
```
darinkrauss: We should check and fail on all errors.  Also, we should use `filepath.Join` to join paths.  So, `filepath.Abs(filepath.Join("_config", name))`.

data/Builder.go
```diff
@@ -0,0 +1,110 @@
...
+//BuildFromDataSet will build the matching type(s) from the given Dataset
+func (typeBuilder *TypeBuilder) BuildFromDataSet(dataSet Dataset) ([]interface{}, *ErrorSet) {
+
+	var set []interface{}
+	var buildError *ErrorSet
+
+	for i := range dataSet {
+		item, err := typeBuilder.BuildFromData(dataSet[i])
+		if err != nil && !err.IsEmpty() {
+			if buildError == nil {
```
darinkrauss: Rather than check this every time through the loop, perhaps create at beginning?  Then, if we are going to stick with Error.IsEmpty perhaps add an ErrorSet.IsEmpty and use that at the end of this function on whether to return the ErrorSet or nil (if the ErrorSet is empty).

data/DataError.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+//Error returns a string representation of the errors
+func (e *Error) Error() string {
+
+	errorsStr := ""
+
+	for i := range e.errors {
+		if errorsStr == "" {
```
darinkrauss: Not sure I understand the point of this.  It looks as if the entire function just uses the first error string found (due to the break in the first if statement).  What is your intent here?

Also, another alternative to building the string this way (and ErrorSet.Error()) would be to walk this list of e.errors and build an array of them.  Then use `strings.Join`?

data/DataError.go
```diff
@@ -0,0 +1,88 @@
...
+			errorsStr = e.errors[i].Error()
+			break
+		}
+		errorsStr = fmt.Sprintln(errorsStr, e.errors[i].Error())
+	}
+
+	return fmt.Sprintf("processing %v found: %s", e.data, errorsStr)
```
darinkrauss: We probably don't want any error string if there aren't any errors contained within, right?

Also, I think the whole "processing %v found" isn't necessary (except maybe debugging).

data/_fixtures/test_data_stream.json
```diff
@@ -0,0 +1,80 @@
...
+    "reason": {
+      "suspended": "manual"
+    },
+    "payload": {
+      "logIndices": [
+        0
+      ]
+    }
+  }
+]
```
darinkrauss: New line at end.

Also, we should all sorts of bogus data here.  Missing required fields, extra fields, fields that have the wrong type, fields that have collections that expect simple types or visa versa, etc.

_config/dataservices.json
jh-bate: there is no issue with having the actual ports we use checked in?

data/Builder.go
jh-bate: i like it

data/basal.go
```diff
@@ -0,0 +1,166 @@
+package data
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	valid "github.com/tidepool-org/platform/Godeps/_workspace/src/gopkg.in/bluesuncorp/validator.v8"
```
darinkrauss: Is this necessary here?  Doesn't look like it is used.  `validator` is used in the init, but where is that pulled in from?  Perhaps it is just the package rename to `valid` that is confusing me.

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/tidepool-org/platform/validate"
+)
+
+func init() {
+	validator.RegisterValidation(rateTag, BasalRateValidator)
```
darinkrauss: I'm inclined to not use const/variables for this type of thing.  Instead using just `"rate"` so it is really clear what is going on.

data/base.go
```diff
@@ -0,0 +1,286 @@
...
+	"github.com/tidepool-org/platform/Godeps/_workspace/src/labix.org/v2/mgo/bson"
+
+	"github.com/tidepool-org/platform/validate"
+)
+
+//used for all data types
+var validator = validate.NewPlatformValidator()
+
+func init() {
+	validator.RegisterValidation(timeStrTag, TimeStringValidator)
```
darinkrauss: Ahhh, no abbreviations. :smile: `timeStringTag` and `timeObjectTag` (and their subsequent values).

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+	validator.RegisterValidation(durationTag, BasalDurationValidator)
+	validator.RegisterValidation(deliveryTypeTag, BasalDeliveryTypeValidator)
+	validator.RegisterValidation(insulinTag, BasalInsulinValidator)
+	validator.RegisterValidation(valueTag, BasalValueValidator)
+}
```
darinkrauss: Will some of these validators be promoted to more general usage if they are needed by other types?

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+	BasalName = "basal"
+
+	deliveryTypeField = "deliveryType"
+	scheduleNameField = "scheduleName"
+	insulinField      = "insulin"
+	valueField        = "value"
+	rateField         = "rate"
+	durationField     = "duration"
+
+	deliveryTypeTag validate.ValidationTag = "basaldeliverytype"
```
darinkrauss: Should we use camelCase values here?  `basalDeliveryType`?  We use that for our json and bson tags.  Or is there a limitation with validator?

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+	validator.RegisterValidation(valueTag, BasalValueValidator)
+}
+
+//Basal represents a basal device data record
+type Basal struct {
+	DeliveryType string          `json:"deliveryType" bson:"deliveryType" valid:"basaldeliverytype"`
+	ScheduleName string          `json:"scheduleName" bson:"scheduleName" valid:"omitempty,required"`
+	Rate         float64         `json:"rate" bson:"rate" valid:"omitempty,basalrate"`
+	Duration     int             `json:"duration" bson:"duration" valid:"omitempty,basalduration"`
+	Insulin      string          `json:"insulin" bson:"insulin,omitempty" valid:"omitempty,basalinsulin"`
```
darinkrauss: Should we omitempty on json output for insulin, value, an suppressed since they might not be in the bson?

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+	Value        int             `json:"value" bson:"value,omitempty" valid:"omitempty,basalvalue"`
+	Suppressed   *SupressedBasal `json:"suppressed" bson:"suppressed,omitempty" valid:"omitempty,required"`
+	Base         `bson:",inline"`
+}
+
+//SupressedBasal represents a suppressed basal portion of a basal
+type SupressedBasal struct {
```
darinkrauss: Two `p`s in SuppressedBasal

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+//Basal represents a basal device data record
+type Basal struct {
+	DeliveryType string          `json:"deliveryType" bson:"deliveryType" valid:"basaldeliverytype"`
+	ScheduleName string          `json:"scheduleName" bson:"scheduleName" valid:"omitempty,required"`
+	Rate         float64         `json:"rate" bson:"rate" valid:"omitempty,basalrate"`
+	Duration     int             `json:"duration" bson:"duration" valid:"omitempty,basalduration"`
+	Insulin      string          `json:"insulin" bson:"insulin,omitempty" valid:"omitempty,basalinsulin"`
+	Value        int             `json:"value" bson:"value,omitempty" valid:"omitempty,basalvalue"`
+	Suppressed   *SupressedBasal `json:"suppressed" bson:"suppressed,omitempty" valid:"omitempty,required"`
+	Base         `bson:",inline"`
```
darinkrauss: I haven't seen the `",inline"` specification before. What does this do?

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+//SupressedBasal represents a suppressed basal portion of a basal
+type SupressedBasal struct {
+	Type         string  `json:"type" bson:"type" valid:"required"`
+	DeliveryType string  `json:"deliveryType" bson:"deliveryType" valid:"basaldeliverytype"`
+	ScheduleName string  `json:"scheduleName" bson:"scheduleName" valid:"omitempty,required"`
```
darinkrauss: Do we need to coordinate on which fields are omitempty for `json`, `bson`, and `valid`?

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+	rateField         = "rate"
+	durationField     = "duration"
+
+	deliveryTypeTag validate.ValidationTag = "basaldeliverytype"
+	rateTag         validate.ValidationTag = "basalrate"
+	durationTag     validate.ValidationTag = "basalduration"
+	insulinTag      validate.ValidationTag = "basalinsulin"
+	valueTag        validate.ValidationTag = "basalvalue"
+
+	injectedDelivery, scheduledDelivery, suspendDelivery, tempDelivery = "injected", "scheduled", "suspend", "temp"
```
darinkrauss: Put each on their own line, please.

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+
+	levemirInsulin, lantusInsulin = "levemir", "lantus"
```
darinkrauss: Own line.

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+
+	rateValidationLowerLimit     = 0.0
+	durationValidationLowerLimit = 0
+	valueValidationLowerLimit    = 0
+	validationGreaterThanInt     = "Must be greater than %d"
```
darinkrauss: No need for this constant here.  Put it inline with all of the others.

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+		deliveryTypeTag: fmt.Sprintf("Must be one of %s,%s,%s,%s", injectedDelivery, scheduledDelivery, suspendDelivery, tempDelivery),
+		rateTag:         fmt.Sprintf("Must be greater than %f", rateValidationLowerLimit),
+		durationTag:     fmt.Sprintf(validationGreaterThanInt, durationValidationLowerLimit),
+		valueTag:        fmt.Sprintf(validationGreaterThanInt, valueValidationLowerLimit),
+		insulinTag:      fmt.Sprintf("Must be one of %s,%s", levemirInsulin, lantusInsulin),
+	}
+)
+
+//BuildBasal will build a Basal record
+func BuildBasal(obj map[string]interface{}) (*Basal, *Error) {
```
darinkrauss: No abbrevs.

data/base.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+//NewCaster creates a Cast
+func NewCaster(err *Error) *Cast {
+	return &Cast{err: err}
+}
+
+//ToString will return the given data as a string or add an error to the cast obj
+func (cast *Cast) ToString(fieldName string, data interface{}) string {
+	if data == nil {
```
darinkrauss: Nil is not the same thing as passing an empty string.  We need to keep the two separate.

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+
+	base, errs := BuildBase(obj)
+	cast := NewCaster(errs)
+
+	basal := &Basal{
+		DeliveryType: cast.ToString(deliveryTypeField, obj[deliveryTypeField]),
```
darinkrauss: Hmm, not sure if casting not returning an error (but adding an error and returning an empty string) is going to work well.

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+		basal.Rate = cast.ToFloat64(rateField, obj[rateField])
+	}
+	if obj[durationField] != nil {
+		basal.Duration = cast.ToInt(durationField, obj[durationField])
+	}
+	if obj[insulinField] != nil {
+		basal.Insulin = cast.ToString(insulinField, obj[insulinField])
+	}
+
+	if validationErrors := validator.Struct(basal); len(validationErrors) > 0 {
```
darinkrauss: Might want to only do validation if the previous casting hasn't returned any errors.  Could get a combination of confusing error messages.  You'll get an error about the casting failure and then a potential subsequent failure about the failed casted value (empty string, for example) not be valid either.

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+		errs.AppendError(validationErrors.GetError(basalFailureReasons))
+	}
+
+	if errs.IsEmpty() {
+		return basal, nil
```
darinkrauss: My general policy is to never return the object if you are returning and error.  The problem is that if the caller ignores the errors they still get a valid object back (even though it isn't really valid).  So, I'd like something like:

```
if !errs.IsEmpty() {
  return nil, errs
}
return basal, nil
```

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return basal, errs
+}
+
+func BasalRateValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
```
darinkrauss: For unused function arguments, use the `_` placeholder.

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+	if rate, ok := field.Interface().(float64); ok {
+		if rate > rateValidationLowerLimit {
```
darinkrauss: Might want to be very clear about whether the LowerLimit is inclusive or exclusive and adjust the code as necessary.

For example, you use `rate > rateValidationLowerLimit` and `rateValidationLowerLimit = 0.0`, which means that a rate of `0.0` is not valid. (Even though I know that a zero basal rate should be valid. :)  So, maybe `rateMinimumInclusive` and use `>=`.  Or, you could assume `Inclusive` on everything and just use `Exclusive` when it isn't?

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+		if duration > durationValidationLowerLimit {
+			return true
+		}
+	}
+	return false
+}
+
+func BasalDeliveryTypeValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	if deliveryType, ok := field.Interface().(string); ok {
+		if _, ok = allowedDeliveryTypes[strings.ToLower(deliveryType)]; ok {
```
darinkrauss: I wonder if it makes sense to normalize these types of strings earlier in the processing so we know it will always be lowercase?  Perhaps as part of the Caster?  For example, `ToLowercaseString`?

data/basal.go
```diff
@@ -0,0 +1,166 @@
...
+		if _, ok = allowedInsulins[strings.ToLower(insulin)]; ok {
+			return true
+		}
+	}
+	return false
+}
+
+func BasalValueValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	if value, ok := field.Interface().(int); ok {
+		if value > valueValidationLowerLimit {
```
darinkrauss: Do we need any sort of upper bounds?

What happens if the incoming data has 1.23 for the value.  When we type cast it what happens to the .23?  Is casting the correct solution for numbers?  Perhaps looking at the fieldType?  We don't want to blindly lose data and return success.  It would be far better to fail with an error message "expected integer, got float" (well, better written than that, but you get the point).

data/basal_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/tidepool-org/platform/validate"
+
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/ginkgo"
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/gomega"
+)
+
+var _ = Describe("Basal", func() {
+
+	const (
+		userid   = "b676436f60"
```
darinkrauss: Should standardize on ID being all capitalized.  I think golint will pick it up if you tried userId and want userID.  So, let's be consistent.

data/basal_test.go
```diff
@@ -0,0 +1,177 @@
...
+			"deviceId":         "IR1285-79-36047-15",
+			"type":             "basal",
+			"deliveryType":     "scheduled",
+			"scheduleName":     "DEFAULT",
+			"rate":             1.75,
+			"duration":         28800000,
+		}
+	)
+
+	Context("datum from obj", func() {
```
darinkrauss: Ahh, kill all abbrevs! :smiling_imp:

data/basal_test.go
```diff
@@ -0,0 +1,177 @@
...
+		It("should return a basal if the obj is valid", func() {
+			basal, _ := BuildBasal(basalObj)
```
darinkrauss: Test all return values.

data/basal_test.go
```diff
@@ -0,0 +1,177 @@
...
+		var (
+			validator = validate.NewPlatformValidator()
+		)
+
+		Context("BasalRateValidator", func() {
+			BeforeEach(func() {
+				validator.RegisterValidation(rateTag, BasalRateValidator)
+			})
+
+			Context("is invalid when", func() {
```
darinkrauss: Should test negative here, too.

data/base.go
```diff
@@ -0,0 +1,286 @@
...
+	"time"
+
+	valid "github.com/tidepool-org/platform/Godeps/_workspace/src/gopkg.in/bluesuncorp/validator.v8"
+	"github.com/tidepool-org/platform/Godeps/_workspace/src/labix.org/v2/mgo/bson"
+
+	"github.com/tidepool-org/platform/validate"
+)
+
+//used for all data types
+var validator = validate.NewPlatformValidator()
```
darinkrauss: Ahhh, THIS is where `validator` comes from!  Totally not clear in basal.go.  Perhaps wrap this in a getter like `GetPlatformValidator` so it is obvious in other files?

data/base.go
```diff
@@ -0,0 +1,286 @@
...
+//Base represents tha base types that all device data records contain
+type Base struct {
+	//required data
+	_ID      bson.ObjectId `bson:"_id" valid:"mongo,required"`
+	ID       string        `json:"id" bson:"id" valid:"required"`
+	UserID   string        `json:"userId" bson:"userId" valid:"required"`
+	DeviceID string        `json:"deviceId" bson:"deviceId" valid:"required"`
+	Time     string        `json:"time" bson:"time" valid:"timestr"`
+	Type     string        `json:"type" bson:"type" valid:"required"`
+	UploadID string        `json:"uploadId" bson:"uploadId" valid:"-"`
```
darinkrauss: We should have an uploadId validator as per my lengthy discussion with DuffLite.

data/base.go
```diff
@@ -0,0 +1,286 @@
...
+		"_active":        0,
+		"_schemaVersion": 0,
+		"_version":       0,
+		"createdTime":    0,
+	}
+)
+
+const (
+	//UserIDField is the userID
+	UserIDField = "userId"
```
darinkrauss: I wouldn't bother with putting all of the field names as constants.  They aren't likely to change.  I use constants generally for "magic" values and such (like tzValidationLowerLimit below).

data/base.go
```diff
@@ -0,0 +1,286 @@
...
+	timeField       = "time"
+	deviceTimeField = "deviceTime"
+
+	payloadField     = "payload"
+	annotationsField = "annotations"
+
+	//validation
+	tzValidationLowerLimit = -840
+	tzValidationUpperLimit = 720
+	timeStrValidationMsg   = "Times need to be ISO 8601 format and not in the future"
```
darinkrauss: Can we separate out the format validation from the future validation?

data/base.go
```diff
@@ -0,0 +1,286 @@
...
+
+	typeField       = "type"
+	timeField       = "time"
+	deviceTimeField = "deviceTime"
+
+	payloadField     = "payload"
+	annotationsField = "annotations"
+
+	//validation
+	tzValidationLowerLimit = -840
```
darinkrauss: `timezoneOffsetMinimum`?

data/base.go
```diff
@@ -0,0 +1,286 @@
...
+//BuildBase builds the base fields that all device data records contain
+func BuildBase(obj map[string]interface{}) (Base, *Error) {
+
+	errs := NewError(obj)
+	cast := NewCaster(errs)
+
+	base := Base{
+		_ID:      bson.NewObjectId(),
+		ID:       bson.NewObjectId().Hex(),
+		UserID:   cast.ToString(UserIDField, obj[UserIDField]),
```
darinkrauss: Perhaps consider a caster that pulls out of the object AND casts.

With pulling out using `obj[UserIDField]` you turn a missing field into an empty string, which are not the same thing.

data/base.go
```diff
@@ -0,0 +1,286 @@
...
+		DeviceID: cast.ToString(deviceIDField, obj[deviceIDField]),
+		UploadID: cast.ToString(uploadIDField, obj[uploadIDField]),
+		Time:     cast.ToString(timeField, obj[timeField]),
+		Type:     cast.ToString(typeField, obj[typeField]),
+		Payload:  obj[payloadField],
+		Internal: Internal{
+			GroupID:       cast.ToString(GroupIDField, obj[GroupIDField]),
+			ActiveFlag:    true,
+			SchemaVersion: 1, //TODO: configured ??
```
darinkrauss: Yes, configured.

data/base.go
```diff
@@ -0,0 +1,286 @@
...
+	aString, ok := data.(string)
+	if !ok {
+		cast.err.AppendFieldError(fieldName, data)
+	}
+	return aString
+}
+
+//ToFloat64 will return the given data as a float64 or add an error to the cast obj
+func (cast *Cast) ToFloat64(fieldName string, data interface{}) float64 {
+	if data == nil {
```
darinkrauss: Same thing about nil not being the same thing as the type's default value.  This is relevant to all of the types below.

data/data_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package data
+
+import (
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/ginkgo"
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/gomega"
+
+	"testing"
```
darinkrauss: This should be above the other two packages.  stdlib packages first, then external packages, and internal packages each separated by a blank line (to help import sorting).

data/datalog.go
```diff
@@ -0,0 +1,7 @@
+package data
+
+import "github.com/tidepool-org/platform/logger"
+
+var (
+	log = logger.Log.GetNamed("data")
```
darinkrauss: You can collapse this down to:

```
var log = logger.Log.GetNamed("data")
```

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/tidepool-org/platform/data"
+	"github.com/tidepool-org/platform/logger"
+	"github.com/tidepool-org/platform/store"
+	"github.com/tidepool-org/platform/user"
+	"github.com/tidepool-org/platform/version"
+
+	"github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/ant0ine/go-json-rest/rest"
```
darinkrauss: Put Godeps stuff above internal stuff.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+	api              *rest.Api
+	dataStore        store.Store
+	validateToken    user.ChainedMiddleware
+	attachPermissons user.ChainedMiddleware
+	resolveGroupID   user.ChainedMiddleware
+}
+
+//NewDataServiceClient returns an initialised client
+func NewDataServiceClient() *DataServiceClient {
+	log.Info(version.Long())
```
darinkrauss: You can drop this.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+
+	return &DataServiceClient{
+		api: rest.NewApi(),
+		//TODO: from config
+		dataStore:        store.NewMongoStore("deviceData"),
+		validateToken:    user.NewAuthorizationMiddleware(userClient).ValidateToken,
+		attachPermissons: user.NewMetadataMiddleware(userClient).GetPermissons,
+		resolveGroupID:   user.NewMetadataMiddleware(userClient).GetGroupID,
+	}
+
```
darinkrauss: Kill extra blank lines.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/ant0ine/go-json-rest/rest"
+)
+
+const (
+	missingPermissionsError = "missing required permissions"
+	missingDataError        = "missing data to process"
+	gettingDataError        = "there was an error getting your data"
+
+	dataservicesName = "dataservices"
```
darinkrauss: No need for constants here.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/tidepool-org/platform/logger"
+	"github.com/tidepool-org/platform/store"
+	"github.com/tidepool-org/platform/user"
+	"github.com/tidepool-org/platform/version"
+
+	"github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/ant0ine/go-json-rest/rest"
+)
+
+const (
+	missingPermissionsError = "missing required permissions"
```
darinkrauss: We might want to refactor out all error message constants into one file for each package.  This will make it easy to use in the API docs.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+		validateToken:    user.NewAuthorizationMiddleware(userClient).ValidateToken,
+		attachPermissons: user.NewMetadataMiddleware(userClient).GetPermissons,
+		resolveGroupID:   user.NewMetadataMiddleware(userClient).GetGroupID,
+	}
+
+}
+
+//Run will run the service
+func (client *DataServiceClient) Run(URL string) error {
+	client.api.Use(rest.DefaultDevStack...)
```
darinkrauss: Any return value to check here?

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+		rest.Get("/version", client.GetVersion),
+		rest.Get("/data/:userid/:datumid", client.validateToken(client.resolveGroupID((client.GetData)))),
+		rest.Post("/dataset/:userid", client.validateToken(client.attachPermissons(client.resolveGroupID(client.PostDataset)))),
+		rest.Get("/dataset/:userid", client.validateToken(client.attachPermissons(client.resolveGroupID(client.GetDataset)))),
+		rest.Post("/blob/:userid", client.validateToken(client.attachPermissons(client.resolveGroupID(client.PostBlob)))),
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+	client.api.SetApp(router)
```
darinkrauss: Any return value to check here?

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+//PostDataset will process a posted dataset for the requested user if permissons are sufficient
+func (client *DataServiceClient) PostDataset(w rest.ResponseWriter, r *rest.Request) {
+
+	userid := r.PathParam(useridParamName)
+
+	log.AddTrace(userid)
+
+	if checkPermisson(r, user.Permission{}) {
+
+		groupID := r.Env[user.GROUPID]
```
darinkrauss: I don't understand this.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+
+		if r.ContentLength == 0 || groupID == "" {
```
darinkrauss: I'd separate these two out.  A missing groupID is not missing data.  It is an internal error (since it should have failed earlier in the middleware chain.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+	w.WriteJson(version.Long())
+	return
+}
+
+//PostDataset will process a posted dataset for the requested user if permissons are sufficient
+func (client *DataServiceClient) PostDataset(w rest.ResponseWriter, r *rest.Request) {
+
+	userid := r.PathParam(useridParamName)
+
+	log.AddTrace(userid)
```
darinkrauss: Logging should be more informative if it is going to be for debugging.  If for logging the request, then it should be in a middleware layer where the info about the request *and* response should be logged.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+			return
+		}
+
+		platformData, err := data.NewTypeBuilder(map[string]interface{}{data.UserIDField: userid, data.GroupIDField: groupID}).BuildFromDataSet(dataSet)
+		processedDataset.Dataset = platformData
+		processedDataset.Errors = err.Error()
+
+		if err.Error() != "" {
+			w.WriteHeader(http.StatusBadRequest)
+			w.WriteJson(&processedDataset)
```
darinkrauss: We shouldn't be returning any data here, I don't think  Certainly not on an error.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+			return
+		}
+
+		//TODO: should this be a bulk insert?
```
darinkrauss: Yes, if we can, for performance.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+		w.WriteHeader(http.StatusNotImplemented)
+		return
+	}
+	rest.Error(w, missingPermissionsError, http.StatusUnauthorized)
+	return
+
+}
+
+//process the found data and send the appropriate response
+func process(iter store.Iterator) data.Dataset {
```
darinkrauss: Kind of killing the purpose of using an iterator here.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+
+	if checkPermisson(r, user.Permission{}) {
+
+		var found struct {
+			data.Dataset `json:"Dataset"`
+			Errors       string `json:"Errors"`
+		}
+
+		userid := r.PathParam(useridParamName)
+		log.Info(useridParamName, userid)
```
darinkrauss: Again, more descriptive log statements.  And, only if absolutely useful in the log.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+
+		types := strings.Split(r.URL.Query().Get("type"), ",")
```
darinkrauss: I'd call these variables `type, `subType`, `startDate`, and `endDate` to be more accurate.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+
+		iter := client.dataStore.ReadAll(
+			store.Fields{data.UserIDField: userid},
+			data.InternalFields,
+		)
+		defer iter.Close()
+
+		found.Dataset = process(iter)
+
+		w.WriteJson(&found)
```
darinkrauss: Probably need to walk the iterator and write your own JSON partials.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+		return
+	}
+	rest.Error(w, missingPermissionsError, http.StatusUnauthorized)
```
darinkrauss: I'd fail this up top where you do the check.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
...
+
+		userid := r.PathParam(useridParamName)
+		datumid := r.PathParam("datumid")
+
+		log.Info("userid and datum", userid, datumid)
+
+		foundDatum.Datum = data.Datum{}
+		foundDatum.Errors = ""
+
+		w.WriteJson(&foundDatum)
```
darinkrauss: Do you have an example of what the JSON output from all of the APIs looks like?  I'm particularly interested in the Errors and how that plays out.

dataservices/dataservices_test.go
```diff
@@ -0,0 +1,309 @@
+package main_test
+
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+	"strings"
+
+	. "github.com/tidepool-org/platform/dataservices"
```
darinkrauss: Don't use dot import here.

dataservices/dataservices_test.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/tidepool-org/platform/service"
+	"github.com/tidepool-org/platform/user"
+	"github.com/tidepool-org/platform/version"
+
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/ginkgo"
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/gomega"
+)
+
+var _ = Describe("The Dataservices client", func() {
```
darinkrauss: We need to figure a good way to do full integration testing including the route and middleware.  Will likely need to mock the user/permissions client.

logger/logger.go
```diff
@@ -0,0 +1,101 @@
+package logger
+
+import (
+	"os"
+
+	logrus "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/Sirupsen/logrus"
```
darinkrauss: This doesn't naturally come out as 'logrus'?

logger/logger.go
```diff
@@ -0,0 +1,101 @@
...
+
+//GetNamed return a named instance of the logger
+func (log *PlatformLog) GetNamed(name string) Logger {
+	named := setup()
+	named.WithField("name", name)
+	return named
+}
+
+//AddTrace will add a trace id to the logs
+func (log *PlatformLog) AddTrace(id string) {
```
darinkrauss: This adds a single field for ALL log statements.  The problem is that it uses the global `Log` and so simultaneous requests will conflict with each other.  If you attached it to just a specific instance of a logger that was only used for a single request (and then discarded) then this might work.

Also consider that we will likely have at least two "tracing" mechanisms.  One for each request and one from the client for each "session" of requests.

logger/logger.go
```diff
@@ -0,0 +1,101 @@
...
+	AddTrace(id string)
+	AddTraceUUID()
+	Debug(args ...interface{})
+	Info(args ...interface{})
+	Warn(args ...interface{})
+	Error(args ...interface{})
+	Fatal(args ...interface{})
+}
+
+const traceID = "traceID"
```
darinkrauss: I wouldn't use this constant here.

logger/logger_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package logger_test
+
+import (
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/ginkgo"
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/gomega"
+
+	"testing"
```
darinkrauss: stdlib first.

logger/logger_test.go
```diff
@@ -0,0 +1,55 @@
+package logger_test
+
+import (
+	. "github.com/tidepool-org/platform/logger"
```
darinkrauss: No dot imports.

dataservices/dataservices.go
```diff
@@ -0,0 +1,239 @@
+package main
```
darinkrauss: Might want to pull out some of this out of the main package into a dataservices package.  The file with the main package should be tiny, I think. Basically, the main function that calls into the dataservices package to do everything else.

service/testutil.go
```diff
@@ -0,0 +1,232 @@
+package service
```
darinkrauss: I think it best to take all of the testing stuff and put it into a `test` or `testing` package by itself, not in the `service` package.

service/testutil.go
```diff
@@ -0,0 +1,232 @@
...
+	panicOnError(err)
+
+	r, err := http.NewRequest(method, urlStr, bodyBuf)
+
+	panicOnError(err)
+
+	r.Header.Set("Content-Type", bodyWriter.FormDataContentType())
+	r.Header.Set("Accept-Encoding", "gzip")
+
+	bodyWriter.Close()
```
darinkrauss: Probably should `defer bodyWriter.Close() right after the `NewWriter` call above.

service/testutil.go
```diff
@@ -0,0 +1,232 @@
...
+	return strings.Contains(r.Body.String(), expectedToContain)
+}
+
+// DecodeJSONPayload decodes the recorded payload to JSON
+func DecodeJSONPayload(r *httptest.ResponseRecorder, v interface{}) error {
+	content, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return err
+	}
+	err = json.Unmarshal(content, v)
```
darinkrauss: Might want to unmarshal into a local variable and then assign to v at end after checking for no error.

store/store.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+//Store is a generic store interface for all operations that we need
+// Note: there will be a specific implementation for the store
+// type we are using, which could also be a mock
+type Store interface {
+	Save(d interface{}) error
+	Update(selector interface{}, d interface{}) error
+	Delete(find Fields) error
+	Read(find Fields, remove Fields, result interface{}) error
```
darinkrauss: Perhaps rather than removed fields what about allowed field?  Prevents future accidental leaks later on.

store/store.go
```diff
@@ -0,0 +1,180 @@
...
+type MongoStore struct {
+	Session        *mgo.Session
+	CollectionName string
+	Config         MongoConfig
+}
+
+//MongoConfig is the required config for the MongoStore
+type MongoConfig struct {
+	URL     string `json:"connectionUrl"`
+	DbName  string `json:"databaseName"`
```
darinkrauss: DatabaseName

store/store.go
```diff
@@ -0,0 +1,180 @@
...
+//MongoStore is the mongo implementation of Store
+type MongoStore struct {
+	Session        *mgo.Session
+	CollectionName string
+	Config         MongoConfig
+}
+
+//MongoConfig is the required config for the MongoStore
+type MongoConfig struct {
+	URL     string `json:"connectionUrl"`
```
darinkrauss: connectionUrl to just url?

store/store.go
```diff
@@ -0,0 +1,180 @@
...
+	DbName  string `json:"databaseName"`
+	Timeout int    `json:"timeout"`
+}
+
+//NewMongoStore returns an initailised instance of MongoStore
+func NewMongoStore(name string) *MongoStore {
+
+	store := &MongoStore{CollectionName: name}
+	config.FromJSON(&store.Config, "mongo.json")
```
darinkrauss: Inject config in here rather than reading it here would make for easier testing?

store/store.go
```diff
@@ -0,0 +1,180 @@
...
+	mapped := bson.M{}
+	for key, val := range fields {
+		mapped[key] = val
+	}
+	return mapped
+}
+
+//Cleanup will cleanup the collection, used for testing purposes
+func (mongoStore *MongoStore) Cleanup() {
+	cpy := mongoStore.Session.Copy()
```
darinkrauss: cpy => copy

store/store.go
```diff
@@ -0,0 +1,180 @@
...
+	defer cpy.Close()
+
+	cpy.DB(mongoStore.Config.DbName).C(mongoStore.CollectionName).DropCollection()
+}
+
+//Save will save the specified data
+func (mongoStore *MongoStore) Save(d interface{}) error {
+	cpy := mongoStore.Session.Copy()
```
darinkrauss: I see a lot of:

```
	cpy := mongoStore.Session.Copy()
	defer cpy.Close()

	cpy.DB(mongoStore.Config.DbName).C(mongoStore.CollectionName)
```

Can we DRY it up?

store/store.go
```diff
@@ -0,0 +1,180 @@
...
+	defer cpy.Close()
+
+	if err := cpy.DB(mongoStore.Config.DbName).C(mongoStore.CollectionName).Insert(d); err != nil {
```
darinkrauss: I think you can just 

```
return cpy.DB(mongoStore.Config.DbName).C(mongoStore.CollectionName).Insert(d)
```

Same for some other code below.

user/client.go
```diff
@@ -0,0 +1,343 @@
...
+	go func() {
+		for {
+			timer := time.After(time.Duration(client.config.TokenRefreshDuration))
+			select {
+			case twoWay := <-client.closed:
+				twoWay <- true
+				return
+			case <-timer:
+				if err := client.serverLogin(); err != nil {
+					log.Error("Error when refreshing server login:", err.Error())
```
darinkrauss: What happens here if this fails? It looks like it only will retry after another token refresh duration.

user/middleware.go
```diff
@@ -0,0 +1,103 @@
...
+	Client Client
+}
+
+//NewAuthorizationMiddleware creates an initialised AuthorizationMiddleware
+func NewAuthorizationMiddleware(userClient Client) *AuthorizationMiddleware {
+	return &AuthorizationMiddleware{Client: userClient}
+}
+
+//ValidateToken returns if valid or sends http.StatusUnauthorized if invalid
+func (mw *AuthorizationMiddleware) ValidateToken(h rest.HandlerFunc) rest.HandlerFunc {
```
darinkrauss: h => handler

user/middleware.go
```diff
@@ -0,0 +1,103 @@
...
+
+		token := r.Header.Get(xTidepoolSessionToken)
+		userid := r.PathParam("userid")
+
+		if tokenData := mw.Client.CheckToken(token); tokenData != nil {
+			if tokenData.IsServer || tokenData.UserID == userid {
+				h(w, r)
+				return
+			}
+			log.Info("id's don't match and not server token", tokenData.UserID, userid)
```
darinkrauss: This should be an Error (as it shouldn't happen with our clients).

user/middleware.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+//MetadataMiddleware middleware is used for getting user permissons
+type MetadataMiddleware struct {
+	Client Client
+}
+
+//PERMISSIONS constant for accessing users permissions that are attached to request.Env
+const PERMISSIONS = "permissons"
```
darinkrauss: "permissons" => "permissions"

validate/validate.go
```diff
@@ -0,0 +1,134 @@
...
+// the FieldError found within the ValidationErrors map
+func (e Errors) Error() string {
+	const fieldErrMsg = "Key: '%s' Error:Field validation for '%s' failed on the '%s' tag"
+	if len(e) > 0 {
+
+		buff := bytes.NewBufferString("")
+
+		for key, err := range e {
+			buff.WriteString(fmt.Sprintf(fieldErrMsg, key, err.Field, err.Tag))
+			buff.WriteString("\n")
```
darinkrauss: Where does this ultimately write to?  Do we want multi-line output?

validate/validate.go
```diff
@@ -0,0 +1,134 @@
...
+		return strings.TrimSpace(buff.String())
+	}
+	return ""
+}
+
+//GetError returns a formatted error message for the user
+func (e Errors) GetError(reasons ErrorReasons) error {
+
+	const (
+		fieldErrorMsg = "Error:Field validation for '%s' failed with '%s' when given '%v' for type '%s' "
```
darinkrauss: Probably don't need the "Error:" prefix. Kinda implied.

validate/validate_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	. "github.com/tidepool-org/platform/validate"
+)
+
+var _ = Describe("Validate", func() {
+
+	Context("basic struct", func() {
+		type testStruct struct {
+			ID      string      `json:"id"  valid:"required"`
+			Type    string      `json:"type"  valid:"required"`
```
darinkrauss: extra spaces

validate/validate_test.go
```diff
@@ -0,0 +1,143 @@
...
+			Offset  int         `json:"offset,omitempty"  valid:"omitempty,min=-10,max=10"`
+			Payload interface{} `json:"payload,omitempty"  valid:"-"`
+		}
+		var (
+			validator = NewPlatformValidator()
+		)
+		Context("is valid", func() {
+			It("all feilds set correctly", func() {
```
darinkrauss: feilds => fields

darinkrauss:
@jh-bate I ran out of steam, but will look at it again when you are ready again for another review.

_config/dataservices.json
darinkrauss: No, I don't think so. The ports should only be accessible to the firewall not to the public.

data/Builder.go
jh-bate: things like the `_groupId`. We store for each data point but we only resolve once it has come in via the api

data/DataError.go
jh-bate: ok `ErrorArray` in line with other changes

data/Builder.go
jh-bate: lets go for single letter then. Then we can use the editor to full effect also 

screen shot 2016-03-24 at 3 22 29 pm

config/config.go
jh-bate: true

data/basal.go
jh-bate: yeah `validator` is used else where and then conflicts with this import hence the `valid` name ... still thinking

data/basal.go
jh-bate: yes can be - but will do so once the need arises

data/basal.go
jh-bate: yeah probably

data/basal.go
jh-bate: yeah -kinda where I really want the data model docs to be honest

data/base.go
jh-bate: yup

data/base.go
jh-bate: yeah - awaiting docs / spec as-per most of the validation.

data/base.go
jh-bate: in sperate PR

dataservices/dataservices.go
jh-bate: the middleware is what I asked you to do - this is just a pass through at the moment

dataservices/dataservices.go
jh-bate: true

dataservices/dataservices.go
jh-bate: talk tomorrow

dataservices/dataservices.go
jh-bate: check seperate PR

data/DataError.go
jh-bate: its a pointer

darinkrauss: Okay, at first I thought that because the pointer was nil there was no way to call the function.  But, with golang, that's not true (and that's really f*ed up).  Technically, in *every* function for a struct pointer you need to check against the pointer being nil. Frankly, I'm surprised this is the case because I see no reason to allow this, its error prone, and the Golang designers should know better. I'd really like to understand their rationale for supporting this (rather than just panicking if you call a function on a nil pointer (eg. `e.Error()` when e is nil)).

data/basal.go
darinkrauss: I see the use of `validator` below in the `init()` function.  As I mention later, I think the raw use of that global variable (that is defined elsewhere) should probably be wrapper in a getter (eg. `GetPlatformValidator()`) so you can rename the internal variable something other than `validator`. Then you can use the `validator` package as it is.

data/basal.go
darinkrauss: :+1:

dataservices/dataservices.go
darinkrauss: Ah, right. Haven't gotten to it yet.

PR #14, opened 2016-03-29
Review from initial

darinkrauss:


_config/dataservices.json
```diff
@@ -0,0 +1,3 @@
+{
```
darinkrauss: Just a reminder that we should figure out our config strategy here. What goes in config file committed to repo versus environment (secrets)? How do we handle different configuration for different environments in the config file committed to repo?

config/config.go
```diff
@@ -0,0 +1,35 @@
...
+	"path/filepath"
+
+	"github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/jinzhu/configor"
+)
+
+func FromJSON(config interface{}, name string) error {
+
+	// we need to get the absolute path and also test it because of the
+	// difference of how this code runs in different environments
+	absPath, _ := filepath.Abs(filepath.Join("./_config/", name))
```
darinkrauss: I don't think we need the "./" prefix here or a couple lines below.  Also, what about checking for errors here?

config/config.go
```diff
@@ -0,0 +1,35 @@
...
+
+	if _, err := os.Stat(absPath); os.IsNotExist(err) {
+		absPath, _ = filepath.Abs(filepath.Join("./../_config/", name))
+		if _, err := os.Stat(absPath); os.IsNotExist(err) {
```
darinkrauss: What happens if it is any other error than IsNotExist? Should we fail on any error out of Stat?

config/config_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package config_test
+
+import (
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/ginkgo"
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/gomega"
+
+	"testing"
```
darinkrauss: stdlib imports should be before 3rd party package imports

config/config_test.go
```diff
@@ -0,0 +1,55 @@
...
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/ginkgo"
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/gomega"
+)
+
+var _ = Describe("Config", func() {
+
+	Describe("FromJson", func() {
+
+		It("loads the given config file", func() {
+			var mgoConfig store.MongoConfig
```
darinkrauss: Still `mgoConfig`, eh? :)

config/config_test.go
```diff
@@ -0,0 +1,55 @@
...
+		})
+	})
+
+	Describe("FromEnv", func() {
+
+		It("loads the given config value from env", func() {
+			const platformKey, platformValue = "CONFIG_TEST", "yay I exist!"
+			os.Setenv(platformKey, platformValue)
+
+			platfromValue, _ := config.FromEnv(platformKey)
```
darinkrauss: Check error return here as well.

config/config_test.go
```diff
@@ -0,0 +1,55 @@
...
+			Expect(platfromValue).To(Equal(platformValue))
+
+			os.Unsetenv(platformKey)
+		})
+
+		It("returns error if the value doesn't exist", func() {
+			const otherKey = "OTHER"
+			os.Unsetenv(otherKey) // make sure it doesn't exist
+
+			_, err := config.FromEnv("OTHER")
```
darinkrauss: Check result return here as well.

config/config_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+			platfromValue, _ := config.FromEnv(platformKey)
+			Expect(platfromValue).To(Equal(platformValue))
+
+			os.Unsetenv(platformKey)
+		})
+
+		It("returns error if the value doesn't exist", func() {
+			const otherKey = "OTHER"
+			os.Unsetenv(otherKey) // make sure it doesn't exist
```
darinkrauss: From previous comment:

You use other `otherKey` and `"OTHER"` below.  If using a constant (which doesn't really need one here) you should use it for all locations.

data/Builder.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/tidepool-org/platform/validate"
+)
+
+type Datum map[string]interface{}
+
+type DatumArray []Datum
+
+type Builder interface {
+	BuildFromDatum(datum Datum) interface{}
```
darinkrauss: What is the return value `interface{}` represent?  Should we have a type for it so the function signature is clear?

data/Builder.go
```diff
@@ -0,0 +1,72 @@
...
+
+	const typeField = "type"
+
+	if datum[typeField] != nil {
+
+		for k, v := range t.inject {
+			datum[k] = v
+		}
+
+		if strings.ToLower(datum[typeField].(string)) == strings.ToLower(BasalName) {
```
darinkrauss: Are we going to stick with a big `if` statement?  Or move to a "registered" factory pattern?  If the `if` statement, I suggest using a `switch` statement instead.

Also, I'm going to refine my stance on "lenient on input, strict on output". I think "keywords" (the `type` value in this case) HAS to be case-sensitive.  There's just too much potential for error by allowing case-insensitivity of everything.   If we don't fail on case issues early on, we either need to 1) force them to one case before storing; or 2) lowercase ANY check EVERYWHERE. Since #2 is really easy to forget and error prone, and #1 technically changes the data from what the client sent, I recommend just requiring the case to be correct.

data/Builder.go
```diff
@@ -0,0 +1,72 @@
...
+			return BuildBasal(datum, validate.ErrorProcessing{BasePath: fmt.Sprintf("%d/%s", t.Index, strings.ToLower(BasalName)), ErrorsArray: t.Errors})
+		} else if strings.ToLower(datum[typeField].(string)) == strings.ToLower(DeviceEventName) {
+			return BuildDeviceEvent(datum, validate.ErrorProcessing{BasePath: fmt.Sprintf("%d/%s", t.Index, strings.ToLower(BasalName)), ErrorsArray: t.Errors})
```
darinkrauss: I think `strings.ToLower(BasalName)` is wrong.

Also, any way to DRY this up?

data/base.go
```diff
@@ -0,0 +1,256 @@
...
+	Time     *string       `json:"time" bson:"time" valid:"timestr"`
+	Type     *string       `json:"type" bson:"type" valid:"required"`
+	UploadID *string       `json:"uploadId" bson:"uploadId" valid:"-"`
+
+	//optional data
+	DeviceTime       *string       `json:"deviceTime,omitempty" bson:"deviceTime,omitempty" valid:"omitempty,timestr"`
+	TimezoneOffset   *int          `json:"timezoneOffset,omitempty" bson:"timezoneOffset,omitempty" valid:"omitempty,timezoneoffset"`
+	ConversionOffset *int          `json:"conversionOffset,omitempty" bson:"conversionOffset,omitempty" valid:"omitempty,required"`
+	ClockDriftOffset *int          `json:"clockDriftOffset,omitempty" bson:"clockDriftOffset,omitempty" valid:"omitempty,required"`
+	Payload          interface{}   `json:"payload,omitempty" bson:"payload,omitempty" valid:"omitempty,payload"`
```
darinkrauss: Might want pointer here and for Annotations since a nil is different than an empty map or empty array.

data/base.go
```diff
@@ -0,0 +1,256 @@
...
+	return isTimeObjectValid(timeObject)
+}
+
+func ToString(fieldName string, data interface{}, errs validate.ErrorProcessing) *string {
+	if data == nil {
+		return nil
+	}
+	aString, ok := data.(string)
+	if !ok {
+		errs.Append(validate.NewPointerError(errs.BasePath+"/"+fieldName, "Invalid type", "should be of type string"))
```
darinkrauss: I'm seeing a lot of `errs.Append(validate.NewPointerError(errs.BasePath+"/"+`

Would it make sense to DRY that up?

data/base.go
```diff
@@ -0,0 +1,256 @@
...
+
+func ToInt(fieldName string, data interface{}, errs validate.ErrorProcessing) *int {
+	if data == nil {
+		return nil
+	}
+	theInt, ok := data.(int)
+
+	if !ok {
+		if theInt == 0 {
+			return &theInt
```
darinkrauss: Not sure I understand this logic.  I believe 0 is the value returned for `data.(int)` if ANY error occurs. It is the default value for an `int`.

config/config_test.go
jh-bate: thats because I started doing the naming updates on the https://github.com/tidepool-org/platform/tree/data-validation branch and then when I got to the the errors branched off. But yeah it will change from being an abbreviation

config/config_test.go
jh-bate: sorry same issue as above - I have made those changes but should merge them into here

data/Builder.go
jh-bate: yeah I guess so - its just one built datapoint

data/Builder.go
jh-bate: yeah ok - will remove the `strings.ToLower`

data/base.go
jh-bate: true - will update

data/base.go
jh-bate: yeah - I just found that out before :) I need to update how this will work

_config/dataservices.json
jh-bate: yeah I kinda broached that subject in the deployment branch review

config/config.go
jh-bate: yeah - they are un-needed. I have added in those error checks too

config/config_test.go
jh-bate: no i tell a lie - fixed it up now :)

config/config_test.go
jh-bate: another lie - have now fixed it up

data/Builder.go
jh-bate: updated now

data/Builder.go
jh-bate: have also cleaned up the building of data types

data/Builder.go
jh-bate: yeah - have cleaned this up now

data/base.go
jh-bate: attempted to, its a little better now

config/config.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/jinzhu/configor"
+)
+
+func FromJSON(config interface{}, name string) error {
+
+	// we need to get the absolute path and also test it because of the
+	// difference of how this code runs in different environments
+	absPath, _ := filepath.Abs(filepath.Join("./_config/", name))
+
+	if _, err := os.Stat(absPath); os.IsNotExist(err) {
```
darinkrauss: What happens if there is any other error than IsNotExist? We should probably fail here out of the function.

config/config.go
```diff
@@ -0,0 +1,35 @@
...
+
+	return configor.Load(&config, absPath)
+}
+
+func FromEnv(name string) (string, error) {
+
+	value := os.Getenv(name)
+
+	if value == "" {
+		return "", fmt.Errorf("$%s must be set", name)
```
darinkrauss: So, I guess we can't know the difference between an empty string and not-defined?  Yes, I guess that is the same as the shell.

config/config_test.go
```diff
@@ -0,0 +1,55 @@
...
+			Expect(random).To(BeNil())
+			Expect(err).ToNot(BeNil())
+			Expect(err.Error()).To(ContainSubstring("random.json: no such file or directory"))
+		})
+	})
+
+	Describe("FromEnv", func() {
+
+		It("loads the given config value from env", func() {
+			const platformKey, platformValue = "CONFIG_TEST", "yay I exist!"
```
darinkrauss: I prefer separate lines for variable declarations/definitions.  It is much clearer, particularly if there are more than a couple variables.

data/Builder.go
```diff
@@ -0,0 +1,72 @@
...
+
+type DatumArray []Datum
+
+type Builder interface {
+	BuildFromDatum(datum Datum) interface{}
+	BuildFromDatumArray(datumArray DatumArray) ([]interface{}, *validate.ErrorsArray)
+}
+
+type TypeBuilder struct {
+	inject map[string]interface{}
```
darinkrauss: Every time I read the 'inject' I have to think twice about what it is for.  Might want to rename to something like "CommonDatum".  Just a thought.

data/Builder.go
```diff
@@ -0,0 +1,72 @@
...
+	Errors *validate.ErrorsArray
+	Index  int
+}
+
+func NewTypeBuilder(inject map[string]interface{}) Builder {
+	return &TypeBuilder{
+		inject: inject,
+		Errors: validate.NewErrorsArray(),
+		Index:  0,
```
darinkrauss: I think we should not use 'Index', but instead something like 'SourceRoot' as string to be more generic. I can imagine a scenario where we have an embedded array within a datum that we are validating.  Then the 'SourceRoot' would be something like '0/myArray/1', but 'Index' wouldn't work.

data/Builder.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+func (t *TypeBuilder) BuildFromDatumArray(datumArray DatumArray) ([]interface{}, *validate.ErrorsArray) {
+
+	var set []interface{}
```
darinkrauss: Its not a `set` it is an array. ;)

data/Builder.go
```diff
@@ -0,0 +1,72 @@
...
+
+	for i := range datumArray {
+		if item := t.BuildFromDatum(datumArray[i]); item != nil {
```
darinkrauss: What happens if `item` is nil? Can this ever happen? If so, does it mean that there is one or more errors? If so, we're just going to return those errors to the user and the resultsArray will be ignored.  So, I see no reason not to just simplify this down to:

```
resultsArray = append(resultsArray, t.BuildFromDatum(datumArray[i]))
```

What do you think?

data/Builder.go
```diff
@@ -0,0 +1,72 @@
...
+
+		if strings.ToLower(datum[typeField].(string)) == strings.ToLower(BasalName) {
+			return BuildBasal(datum, validate.ErrorProcessing{BasePath: fmt.Sprintf("%d/%s", t.Index, strings.ToLower(BasalName)), ErrorsArray: t.Errors})
+		} else if strings.ToLower(datum[typeField].(string)) == strings.ToLower(DeviceEventName) {
+			return BuildDeviceEvent(datum, validate.ErrorProcessing{BasePath: fmt.Sprintf("%d/%s", t.Index, strings.ToLower(BasalName)), ErrorsArray: t.Errors})
+		}
+		t.Errors.Append(validate.NewPointerError(fmt.Sprintf("%d/type", t.Index), "Invalid type", "The type must be one of `basal`, `deviceEvent`"))
+		return nil
+	}
+	t.Errors.Append(validate.NewParameterError(fmt.Sprintf("%d", t.Index), "Invalid data", "Missing required `type` field."))
```
darinkrauss: I know this is picky, but we should decide if we are going to include the trailing period in error messages or not.  I'd prefer not, but whatever we choose we should be consistent.  Same with embedded quotations (either double-quotes, single-quotes, or backticks).

Also, the "Invalid type" and "Invalid data" strings are one place where constants would be useful since they will 1) be used all over the place, and 2) more importantly, may be used for error checking on the client so being exact with the string is important (case sensitivity, spelling, etc.)

data/_fixtures/test_data_stream.json
```diff
@@ -0,0 +1,80 @@
...
+    "reason": {
+      "suspended": "manual"
+    },
+    "payload": {
+      "logIndices": [
+        0
+      ]
+    }
+  }
+]
```
darinkrauss: Newline at end.

data/basal.go
```diff
@@ -0,0 +1,141 @@
+package data
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	valid "github.com/tidepool-org/platform/Godeps/_workspace/src/gopkg.in/bluesuncorp/validator.v8"
```
darinkrauss: Have you had a chance to clean up with `validator` global variable into a Getter and then be able to get rid of the special `valid` import here? (From the previous PR.)

data/basal.go
```diff
@@ -0,0 +1,141 @@
...
+
+func init() {
+	validator.RegisterValidation(rateTag, BasalRateValidator)
+	validator.RegisterValidation(durationTag, BasalDurationValidator)
+	validator.RegisterValidation(deliveryTypeTag, BasalDeliveryTypeValidator)
+	validator.RegisterValidation(insulinTag, BasalInsulinValidator)
+	validator.RegisterValidation(valueTag, BasalValueValidator)
+}
+
+type Basal struct {
```
darinkrauss: A couple comments about data types. 

After working on the de-duplication problem and the variability of raw device data (and the potential for variability of Tidepool data, particularly when 3rd parties start uploading data), I now think we should be as strict as possible. (Screw being lenient, we need to make sure our data is as pristine as possible. Bad data is a huge tech debt to deal with.)  I think it does make sense to be lenient in some areas of our API, but not the data.  For example, we could be lenient on signup when the user enters an email address (we should accept any case, but force it to lowercase before creating the account).

Also, I think we should be strict not only about the format of an individual fields, but also how the entire data point looks.  For example, for a Basal data point, many of these fields should be absolutely required.  Also, we should enforce extra validations upon data fields are interdependent (for example, if we allowed both 'mg/dL' and 'mmol' units then the range of acceptable values should be tied to the units).

My 2 cents... I guess I'm just getting paranoid about our data.  After all, our data is everything. Without clean data our usefulness is reduced and dirty data is just a huge tech debt.  This was reinforced with all the data de-dup work.

data/basal.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+type SuppressedBasal struct {
+	Type         *string  `json:"type" bson:"type" valid:"required"`
+	DeliveryType *string  `json:"deliveryType" bson:"deliveryType" valid:"basaldeliverytype"`
+	ScheduleName *string  `json:"scheduleName" bson:"scheduleName" valid:"omitempty,required"`
+	Rate         *float64 `json:"rate" bson:"rate" valid:"omitempty,basalrate"`
+}
+
+const (
```
darinkrauss: I believe I had some comments in the last PR about these constants.

data/basal.go
```diff
@@ -0,0 +1,141 @@
...
+		valueTag:        fmt.Sprintf("Must be greater than %d", valueValidationLowerLimit),
+		insulinTag:      fmt.Sprintf("Must be one of %s,%s", levemirInsulin, lantusInsulin),
+	}
+)
+
+//BuildBasal will build a Basal record
+func BuildBasal(datum Datum, errs validate.ErrorProcessing) *Basal {
+
+	basal := &Basal{
+		ScheduleName: ToString(scheduleNameField, datum[scheduleNameField], errs),
```
darinkrauss: I see where you are going with this.  It does tighten up the code, but somehow it feels a little much to me.  Not that I have a better idea, though.  :)  Perhaps partly because the function is called `ToString`, but is doing far more that just converting to a string.  Also, the duplication of the field name twice seems repetitive.  Perhaps something like:

```
ScheduleName: FetchAndValidateStringField(datum, "scheduleName", errs)
```

Furthermore, what if we added the Fetch function to the TypeBuilder and passed that into the BuildBasal function rather than the errs.  For example:

```
func BuildBasal(datum Datum, builder *validate.TypeBuilder) *Basal {

	basal := &Basal{
		ScheduleName: builder.FetchAndValidateStringField(datum, "scheduleName"),
```

Not sure if that is better, but something to think about.

validate/validate.go
```diff
@@ -0,0 +1,56 @@
...
+func (pv *PlatformValidator) SetErrorReasons(reasons ErrorReasons) *PlatformValidator {
+	pv.reasons = reasons
+	return pv
+}
+
+func (pv *PlatformValidator) toErrorsArray(ve validator.ValidationErrors, errorProcessing ErrorProcessing) {
+	for _, v := range ve {
+		if reason, ok := pv.reasons[ValidationTag(v.Tag)]; ok {
+			errorProcessing.Append(NewPointerError(
+				fmt.Sprintf("%s/%s", errorProcessing.BasePath, v.Type),
```
darinkrauss: Is this v.Type or v.Tag? Not sure I understand the difference.

data/basal.go
```diff
@@ -0,0 +1,141 @@
...
+		DeliveryType: ToString(deliveryTypeField, datum[deliveryTypeField], errs),
+		Rate:         ToFloat64(rateField, datum[rateField], errs),
+		Duration:     ToInt(durationField, datum[durationField], errs),
+		Insulin:      ToString(insulinField, datum[insulinField], errs),
+		Base:         BuildBase(datum, errs),
+	}
+
+	validator.SetErrorReasons(basalFailureReasons).Struct(basal, errs)
```
darinkrauss: Rather than a separate `SetErrorReaons` I would go with:

```
validator.ValidateStruct(basal, basalFailureReasons, errs)
```

data/basal.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return basal
+}
+
+func BasalRateValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	if rate, ok := field.Interface().(float64); ok {
+		if rate > rateValidationLowerLimit {
```
darinkrauss: I believe the last PR has some comments regarding whether the limits were inclusive or not.

data/basal.go
```diff
@@ -0,0 +1,141 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func BasalDurationValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	if duration, ok := field.Interface().(int); ok {
+		if duration > durationValidationLowerLimit {
```
darinkrauss: I think I'd switch the logic around somewhat for all of these

```
if duration, ok := field.Interface().(int); !ok {
  return false
} else {
  return duration > durationValidationLowerLimit
}
```

Maybe?

data/basal.go
```diff
@@ -0,0 +1,141 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func BasalDeliveryTypeValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	if deliveryType, ok := field.Interface().(string); ok {
+		if _, ok = allowedDeliveryTypes[strings.ToLower(deliveryType)]; ok {
```
darinkrauss: Ditch the `strings.ToLower` for all of these.  Lets be more strict.

data/basal_test.go
```diff
@@ -0,0 +1,248 @@
+package data
+
+import (
+	"github.com/tidepool-org/platform/validate"
+
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/ginkgo"
+	. "github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/onsi/gomega"
+)
+
+var _ = Describe("Basal", func() {
```
darinkrauss: No `var _ =` ?

data/base.go
```diff
@@ -0,0 +1,256 @@
...
+	GroupID       string `json:"-" bson:"_groupId" valid:"required"`
+	ActiveFlag    bool   `json:"-" bson:"_active" valid:"required"`
+	SchemaVersion int    `json:"-" bson:"_schemaVersion" valid:"required,min=0"`
+	Version       int    `json:"-" bson:"_version,omitempty" valid:"-"`
+}
+
+var (
+	//InternalFields are what we only use internally in the service and don't wish to return to any clients
+	InternalFields = map[string]interface{}{
+		"_groupId":       0,
```
darinkrauss: These are all supposed to be strings, right? If so, might want to initialize to `""`.

data/base.go
```diff
@@ -0,0 +1,256 @@
...
+
+	base := Base{
+		_ID:              bson.NewObjectId(),
+		ID:               bson.NewObjectId().Hex(),
+		UserID:           ToString(UserIDField, datum[UserIDField], errs),
+		DeviceID:         ToString(deviceIDField, datum[deviceIDField], errs),
+		UploadID:         ToString(uploadIDField, datum[uploadIDField], errs),
+		Time:             ToString(timeField, datum[timeField], errs),
+		Type:             ToString(typeField, datum[typeField], errs),
+		Payload:          datum[payloadField],
```
darinkrauss: Probably want some sort of TODO to provide a minimum validation for the payload field.

data/base.go
```diff
@@ -0,0 +1,256 @@
...
+	return base
+}
+
+var validationFailureReasons = validate.ErrorReasons{
+	timeStringTag:     timeStrValidationMsg,
+	timeZoneOffsetTag: fmt.Sprintf("TimezoneOffset needs to be in minutes and greater than %d and less than %d", tzValidationLowerLimit, tzValidationUpperLimit),
+}
+
+func PayloadValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	//a place holder for more through validation
```
darinkrauss: Make this a TODO or FIXME comment so we catch it.

data/base.go
```diff
@@ -0,0 +1,256 @@
...
+	if _, ok := field.Interface().(interface{}); ok {
+		return true
+	}
+	return false
+}
+
+func AnnotationsValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	//a place holder for more through validation
```
darinkrauss: Make this a TODO or FIXME comment so we catch it.

data/base.go
```diff
@@ -0,0 +1,256 @@
...
+	if _, ok := field.Interface().([]interface{}); ok {
+		return true
+	}
+	return false
+}
+
+func TimezoneOffsetValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	if offset, ok := field.Interface().(int); ok {
```
darinkrauss: I think we need more validation here.  TODO comment?

data/base.go
```diff
@@ -0,0 +1,256 @@
...
+
+func TimeObjectValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
+	if timeObject, ok := field.Interface().(time.Time); ok {
+		return isTimeObjectValid(timeObject)
+	}
+	return false
+}
+
+func isTimeObjectValid(timeObject time.Time) bool {
+	return !timeObject.IsZero() && timeObject.Before(time.Now())
```
darinkrauss: Technically, a time object is valid even if it is after now. We might want to clarify this in the naming, "PastTimeObjectValidator"?

data/base.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func TimeStringValidator(v *valid.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string) bool {
```
darinkrauss: Same thing here with the "Past..."

data/Builder.go
jh-bate: yeah good idea will rename it

data/Builder.go
jh-bate: yeah - good point

data/Builder.go
jh-bate: ok 
- no trailing periods
- use single quotes 

- "Invalid type" and "Invalid data" are constants declared in the data package

data/basal.go
jh-bate: Updated now

data/basal.go
jh-bate: yeah I know what you mean and I also had a few changes to this - maybe we just mull it over a little and then change

data/basal.go
jh-bate: yeah makes sense

data/basal.go
jh-bate: yeah done that

data/base.go
jh-bate: no - this is for the mongo filter they are `0` or `1`

data/base.go
jh-bate: done

data/base.go
darinkrauss: Ah, now I see. It is for the select.  Sure, all good.

data/basal_test.go
jh-bate: I believe that the outer describe is the same on all the tests - when you do a `ginkgo generate basal` it will be created

data/base.go
jh-bate: Agreed - updated

PR #16, opened 2016-04-04
Data validation pr 2

darinkrauss:
@jh-bate For final review without Godeps


config/config.go
```diff
@@ -0,0 +1,41 @@
+package config
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/tidepool-org/platform/Godeps/_workspace/src/github.com/jinzhu/configor"
```
darinkrauss: Just noticed this. Do we actually have to write it this way? I thought if we used `godep go build` then it sets the GOPATH up correctly for the build, so the import can be just `"github.com/jinzhu/configor"`, as normal. That's why we'd used `godep go build`.

config/config.go
```diff
@@ -0,0 +1,41 @@
...
+func FromJSON(config interface{}, name string) error {
+
+	// we need to get the absolute path and also test it because of the
+	// difference of how this code runs in different environments
+	absPath, err := filepath.Abs(filepath.Join("_config/", name))
+	if err != nil {
+		return err
+	}
+
+	if _, err := os.Stat(absPath); os.IsNotExist(err) {
```
darinkrauss: What about errors other than `os.IsNotExist` here?

config/config.go
```diff
@@ -0,0 +1,41 @@
...
+		absPath, err = filepath.Abs(filepath.Join("../_config/", name))
+		if err != nil {
+			return err
+		}
+		if _, err := os.Stat(absPath); os.IsNotExist(err) {
```
darinkrauss: Same here.

data/Builder.go
```diff
@@ -0,0 +1,82 @@
...
+
+const (
+	InvalidTypeTitle = "Invalid type"
+	InvalidDataTitle = "Invalid data"
+)
+
+type Datum map[string]interface{}
+type DatumArray []Datum
+
+type BuiltDatum interface{}
```
darinkrauss: Are we going to have to use `BuiltDatum` all over the place. I think I would change `Datum` to `RawDatum` and `BuiltDatum` to `Datum`.  More clear.  Since it is a global change, I can do it if you want (with a regex find/replace).

Or, we can keep it as-is.

PR #17, opened 2016-05-03
update tests to reflect what is needed for parsing animas data

jh-bate:
@darinkrauss this should be all the cleanup needed


darinkrauss:
Looks good. 🚢  However, I still need to add `.SetValueAllowedToBeEmpty(true)` to the two blood glucose validators in `buildBloodGlucoseTarget` in `calculator.go`. (And comment out the associated tests.)

PR #19, opened 2016-05-04
Device event validation

jh-bate:
@darinkrauss validation and tests updated based on docs.

### DIFF
--- a/data/types/base.go
+++ b/data/types/base.go
@@ -33,12 +33,12 @@ type (
 		UploadID *string       `json:"uploadId" bson:"uploadId" valid:"-"`
 
 		//optional data
-		DeviceTime       *string        `json:"deviceTime,omitempty" bson:"deviceTime,omitempty" valid:"omitempty,timestr"`
-		TimezoneOffset   *int           `json:"timezoneOffset,omitempty" bson:"timezoneOffset,omitempty" valid:"omitempty,timezoneoffset"`
-		ConversionOffset *int           `json:"conversionOffset,omitempty" bson:"conversionOffset,omitempty" valid:"omitempty,required"`
-		ClockDriftOffset *int           `json:"clockDriftOffset,omitempty" bson:"clockDriftOffset,omitempty" valid:"omitempty,required"`
-		Payload          *interface{}   `json:"payload,omitempty" bson:"payload,omitempty" valid:"omitempty,payload"`
-		Annotations      *[]interface{} `json:"annotations,omitempty" bson:"annotations,omitempty" valid:"omitempty,annotations"`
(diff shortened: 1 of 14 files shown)


data/types/base.go
```diff
@@ -283,6 +283,18 @@ func (d Datum) ToStringArray(fieldName string, errs validate.ErrorProcessing) *[
 	return &stringArray
 }
 
+func (d Datum) ToMap(fieldName string, errs validate.ErrorProcessing) *map[string]interface{} {
```
darinkrauss: I don't think you need to return a pointer to a map here. You should be able to return nil for a map (or array or interface).

data/types/device/alarm.go
```diff
@@ -15,10 +15,15 @@ func init() {
 
 type Alarm struct {
 	AlarmType *string `json:"alarmType" bson:"alarmType" valid:"devicealarmtype"`
+	Status    *string `json:"status,omitempty" bson:"status,omitempty" valid:"-"`
```
darinkrauss: Do we want to standardize on where the `omitempty` occurs in the tag?  With a quick find I see that the vast majority have it second except for a few validators where it is first (such as `valid:"omitempty,required"`).

data/types/device/prime.go
```diff
@@ -21,11 +20,18 @@ type Prime struct {
...
-		Message:           "Must be greater than 0.0",
-		AllowedFloatRange: &types.AllowedFloatRange{LowerLimit: 0.0},
+		Message:           "Must be >= 0.0 and <= 3.0",
+		AllowedFloatRange: &types.AllowedFloatRange{LowerLimit: 0.0, UpperLimit: 3.0},
+	}
+
+	tubingVolumeField = types.FloatDatumField{
+		DatumField:        &types.DatumField{Name: "volume"},
+		Tag:               "devicevolume",
+		Message:           "Must be >= 0.0 and <= 100.0",
```
darinkrauss: We should decide whether we want these characters in the strings (that will be escaped in JSON).  Just makes it harder to read the output.

data/types/device/alarm.go
jh-bate: its a requirement for validators to have it first and the other tags it is the opposite

data/types/device/prime.go
jh-bate: yeah agreed - I was going to do a bulk update when we looked at all message we are returning to client of the service

darinkrauss:
A couple comments, but otherwise looks good. :shipit:

data/types/base.go
jh-bate: true - i have update the code to reflect this

PR #20, opened 2016-05-05
Date fix

jh-bate:
update all date time validation based on docs 

 - in prep work moved validation into own times (time and bloodglucose) 
 - moving towards camelCase for validation tag names
 - bolus validation moved to bolus section as not required elsewhere 
 - tests updated


data/types/base_test.go
```diff
@@ -62,7 +62,7 @@ var _ = Describe("Base", func() {
 							types.BuildBase(baseObj, helper.ErrorProcessing),
 							types.ExpectedErrorDetails{
 								Path:   "0/time",
-								Detail: "Times need to be ISO 8601 format and not in the future given 'Monday, 02 Jan 2016'",
+								Detail: "An ISO 8601-formatted timestamp including either a timezone offset from UTC OR converted to UTC with a final Z for 'Zulu' time. e.g.2013-05-04T03:58:44.584Z OR 2013-05-04T03:58:44-08:00 given 'Monday, 02 Jan 2016'",
```
darinkrauss: I don't think we need to be quite so descriptive here. As long as the error message is unique the client can always read the data model docs for more info.

darinkrauss:
One minor nit that you can address later.  Otherwise, looks good. 🚢  I'm going to fast-forward master to this.

PR #21, opened 2016-05-05
New client work, data updates

darinkrauss:
@jh-bate Feel free to review or not. We can do a more global review later.


jh-bate:
@darinkrauss had a quick run through, looks good enough for now. I will do a quick fix for the conversion  issue i mentioned today

PR #22, opened 2016-05-06
fix conversion issue when value is allowed to be empty

jh-bate:


jh-bate:
@darinkrauss the fix mentioned to you, will merge for now

darinkrauss:
LTGM 🚢

PR #24, opened 2016-05-09
Refactor service errors

darinkrauss:
- Remove separate errors packages
- Move from error maps to error functions (better type safety, no clone)

@jh-bate Consolidating into a more usable style in preparation for pvn, but even if we don't use pvn this is still a better solution than the map lookup


jh-bate:
I agree -

PR #31, opened 2016-05-25
Carelink fixes

jh-bate:
fixes and updates based on testing with @gniezen and @darinkrauss

### DIFF
--- a/data/types/base/pump/basal_schedule.go
+++ b/data/types/base/pump/basal_schedule.go
@@ -51,7 +51,7 @@ func ParseBasalSchedulesMap(parser data.ObjectParser) *map[string]*[]*BasalSched
 
 	if parser.Object() != nil {
 		basalScheduleMap = map[string]*[]*BasalSchedule{}
-		for key, _ := range *parser.Object() {
+		for key := range *parser.Object() {
 			basalScheduleMap[key] = ParseBasalScheduleArray(parser.NewChildArrayParser(key))
 		}
 	}
--- a/data/types/base/pump/basal_schedule.go
+++ b/data/types/base/pump/basal_schedule.go
@@ -24,7 +24,7 @@ func (b *BasalSchedule) Validate(validator data.Validator) {
 func (b *BasalSchedule) Normalize(normalizer data.Normalizer) {
 }
 
-func ParseBasalSchedule(parser data.ObjectParser) *BasalSchedule {
+func parseScheduleItem(parser data.ObjectParser) *BasalSchedule {
 	var basalSchedule *BasalSchedule
 	if parser.Object() != nil {
 		basalSchedule = NewBasalSchedule()
@@ -33,14 +33,27 @@ func ParseBasalSchedule(parser data.ObjectParser) *BasalSchedule {
 	return basalSchedule
 }
 
-func ParseBasalScheduleArray(parser data.ArrayParser) *map[string][]*BasalSchedule {
-	var basalScheduleArray *map[string][]*BasalSchedule
+func ParseBasalScheduleArray(parser data.ArrayParser) *[]*BasalSchedule {
+	var basalScheduleArray *[]*BasalSchedule
 	if parser.Array() != nil {
-		basalScheduleArray = &map[string][]*BasalSchedule{}
+		basalScheduleArray = &[]*BasalSchedule{}
 
-		// for index := range *parser.Array() {
-		// 	*basalScheduleArray[key] = ParseBasalSchedule(parser.NewChildObjectParser(key))
-		// }
+		for index := range *parser.Array() {
+			*basalScheduleArray = append(*basalScheduleArray, parseScheduleItem(parser.NewChildObjectParser(index)))
+		}
 	}
 	return basalScheduleArray
+}
+
+func ParseBasalSchedulesMap(parser data.ObjectParser) *map[string]*[]*BasalSchedule {
+
+	var basalScheduleMap map[string]*[]*BasalSchedule
+
+	if parser.Object() != nil {
+		basalScheduleMap = map[string]*[]*BasalSchedule{}
+		for key, _ := range *parser.Object() {
+			basalScheduleMap[key] = ParseBasalScheduleArray(parser.NewChildArrayParser(key))
+		}
+	}
+	return &basalScheduleMap
 }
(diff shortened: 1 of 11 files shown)


data/types/base/selfmonitored/self_monitored.go
```diff
@@ -9,8 +9,9 @@ import (
 type BloodGlucose struct {
 	base.Base `bson:",inline"`
 
-	Value *float64 `json:"value,omitempty" bson:"value,omitempty"`
-	Units *string  `json:"units,omitempty" bson:"units,omitempty"`
+	Value   *float64 `json:"value,omitempty" bson:"value,omitempty"`
+	Units   *string  `json:"units,omitempty" bson:"units,omitempty"`
+	SubType *string  `json:"subType,omitempty" bson:"subType,omitempty"`
```
darinkrauss: If this adds a subType, you'll need to add the Meta object (look at basal, bolus, or device for the Meta()).  The base.Meta() only has type, which is why basal, bolus, and device all have their own Meta() function that adds the subType (or deviceType).

data/types/base/pump/settings.go
```diff
@@ -87,6 +85,17 @@ func (s *Settings) Validate(validator data.Validator) {
 		}
 	}
 
+	if s.BasalSchedules != nil {
```
darinkrauss: You need to add an additional level of validator since this appears to be an array WITHIN an array.  So:
```
	if s.BasalSchedules != nil {
		basalSchedulesValidator := validator.NewChildValidator("basalSchedules")
		for basalScheduleName, basalSchedule := range *s.BasalSchedules {
			// TODO: Do we want to validate the basalScheduleName?  Length > 0?
			if basalSchedule != nil {
				basalScheduleValidator := basalSchedulesValidator.NewChildValidator(basalScheduleName)
				for basalScheduleItemIndex, basalScheduleItem := range *basalSchedule {
					basalScheduleItem.Validate(basalSchedulesValidator.NewChildValidator(basalScheduleItemIndex))
				}
			}
		}
	}
```

data/types/base/pump/insulin_sensitivity.go
```diff
@@ -33,6 +33,12 @@ func (i *InsulinSensitivity) Validate(validator data.Validator) {
 }
 
 func (i *InsulinSensitivity) Normalize(normalizer data.Normalizer) {
+	if i.amountUnits == nil {
```
darinkrauss: For this type of thing were you are "secretly" passing an extra parameter down to Normalize (or Validate/Parse) there's no reason why you can't simply add it to the Normalize function, such as:

```
func (i *InsulinSensitivity) Normalize(normalizer data.Normalizer, units string)
```

After all, the only objects that MUST conform to the standard Normalize signature are the data types (data.Datum). Any sub-structs can use anything they want.

data/types/base/selfmonitored/self_monitored.go
jh-bate: i think it is actually more the overloaded use of `subType` ... or thats how I read it. Other subtypes create different types of records with different fields, this one does not. http://developer.tidepool.io/data-model/device-data/types/smbg.html#subtype

data/types/base/pump/basal_schedule.go
```diff
@@ -24,7 +24,7 @@ func (b *BasalSchedule) Validate(validator data.Validator) {
 func (b *BasalSchedule) Normalize(normalizer data.Normalizer) {
 }
 
-func ParseBasalSchedule(parser data.ObjectParser) *BasalSchedule {
+func parseScheduleItem(parser data.ObjectParser) *BasalSchedule {
 	var basalSchedule *BasalSchedule
 	if parser.Object() != nil {
```
darinkrauss: Since this function can return nil, do we validate that it isn't nil during the Validate phase?

data/types/base/pump/basal_schedule.go
```diff
@@ -33,14 +33,27 @@ func ParseBasalSchedule(parser data.ObjectParser) *BasalSchedule {
 	return basalSchedule
 }
 
-func ParseBasalScheduleArray(parser data.ArrayParser) *map[string][]*BasalSchedule {
-	var basalScheduleArray *map[string][]*BasalSchedule
+func ParseBasalScheduleArray(parser data.ArrayParser) *[]*BasalSchedule {
+	var basalScheduleArray *[]*BasalSchedule
 	if parser.Array() != nil {
```
darinkrauss: Since this function can return nil, do we validate that it isn't nil during the Validate phase?

data/types/base/pump/basal_schedule.go
```diff
@@ -33,14 +33,27 @@ func ParseBasalSchedule(parser data.ObjectParser) *BasalSchedule {
...
+		}
 	}
 	return basalScheduleArray
 }
+
+func ParseBasalSchedulesMap(parser data.ObjectParser) *map[string]*[]*BasalSchedule {
+
+	var basalScheduleMap map[string]*[]*BasalSchedule
+
+	if parser.Object() != nil {
```
darinkrauss: Since this function can return nil, do we validate that it isn't nil during the Validate phase?

data/types/base/pump/basal_schedule.go
```diff
@@ -33,14 +33,27 @@ func ParseBasalSchedule(parser data.ObjectParser) *BasalSchedule {
...
+		basalScheduleMap = map[string]*[]*BasalSchedule{}
+		for key := range *parser.Object() {
+			basalScheduleMap[key] = ParseBasalScheduleArray(parser.NewChildArrayParser(key))
+		}
+	}
+	return &basalScheduleMap
```
darinkrauss: Since the basalScheduleMap is only declared above (not defined), isn't this potentially taking the address of nil?

data/types/base/selfmonitored/self_monitored.go
darinkrauss: Okay, sound reasonable, though confusing for the overloaded use.

darinkrauss:
Overall, looks good, but some comments that I think need to be addressed.

data/types/base/selfmonitored/self_monitored.go
jh-bate: yeah I agree - there are a few inconsistencies with the data-model. Maybe we could rename the struct field and then migrate the data-model type over time?

data/types/base/pump/insulin_sensitivity.go
jh-bate: ahh true! yeah I was following the datum interface hence the extra field

data/types/base/selfmonitored/self_monitored.go
darinkrauss: Sure, sounds reasonable.

PR #33, opened 2016-05-27
fixes for units conversion and also cleanup of bg data

jh-bate:
@darinkrauss once I change those switch statements the tests started failing. This should address the issue 

- fixed switch statements that where incorrectly using pointers
- fix associated tests to ensure that raw objects are initialised with correct units and associated value ranges for either mg/dL or mmol/L

- cleanup bg range checks 
 -  `bloodglucose.AllowedMmolLRange()`
 - `bloodglucose.AllowedMgdLRange()`

- cleanup bg allowed units checks
 - `bloodglucose.AllowedUnits`

- fix validation for `normal` bolus 

> The Animas driver does remove zero boluses before passing it on to the platform API. I sent a zero bolus to the platform API to check if it fails (it should), and it didn’t. For normal boluses, normal >0


darinkrauss:
@jh-bate Overall, looks good. I'm going to add some defensive code (like checking that s.units is != nil before dereferencing). Then I'll push it into the new-validation branch.

PR #34, opened 2016-06-01
General performance/logging improvements

darinkrauss:
@jh-bate Attempt to fix the slow performance on `stg`. Three separate, very focused commits.


jh-bate:
LGTM
 - discussed logging level and where data goes (encrypted app server)
 - also changes to query should enable use of existing indexes

PR #35, opened 2016-06-03
Various refinements

darinkrauss:
@jh-bate Each commit is pretty focused as to what it does. The large commit (https://github.com/tidepool-org/platform/commit/7d066ea982c238b13cd9aa96067c30b478446ab4) is the  one that deals with unparsed values and either ignores, logs, or errors on them.

### DIFF
--- a/data/context/standard_test.go
+++ b/data/context/standard_test.go
@@ -6,19 +6,28 @@ import (
 
 	"github.com/tidepool-org/platform/data"
 	"github.com/tidepool-org/platform/data/context"
+	"github.com/tidepool-org/platform/log/test"
 	"github.com/tidepool-org/platform/service"
 )
 
 var _ = Describe("Standard", func() {
-	Context("New", func() {
+	It("NewStandard returns an error if logger is nil", func() {
+		standard, err := context.NewStandard(nil)
+		Expect(standard).To(BeNil())
+		Expect(err).To(HaveOccurred())
+	})
+
+	Context("NewStandard", func() {
 		var standard *context.Standard
+		var err error
 
 		BeforeEach(func() {
-			standard = context.NewStandard()
+			standard, err = context.NewStandard(test.NewLogger())
 		})
 
 		It("exists", func() {
 			Expect(standard).ToNot(BeNil())
+			Expect(err).To(Succeed())
 		})
 
 		It("has a contained Errors that is empty", func() {
@@ -30,11 +39,21 @@ var _ = Describe("Standard", func() {
 			Expect(standard.Errors()).To(BeEmpty())
 		})
 
+		It("Logger returns a logger", func() {
+			Expect(standard.Logger()).ToNot(BeNil())
+		})
+
 		Context("SetMeta", func() {
 			It("sets the meta on the context", func() {
 				meta := "metametameta"
 				standard.SetMeta(meta)
 				Expect(standard.Meta()).To(BeIdenticalTo(meta))
+			})
+		})
+
+		Context("ResolveReference", func() {
+			It("correctly returns the resolved reference", func() {
+				Expect(standard.ResolveReference("reference")).To(Equal("/reference"))
 			})
 		})
 
@@ -152,6 +171,16 @@ var _ = Describe("Standard", func() {
 				Expect(child).ToNot(BeNil())
 			})
 
+			It("Logger returns a logger", func() {
+				Expect(child.Logger()).ToNot(BeNil())
+			})
+
+			Context("ResolveReference", func() {
+				It("correctly returns the resolved reference", func() {
+					Expect(child.ResolveReference("reference")).To(Equal("/child/reference"))
+				})
+			})
+
 			Context("AppendError with a first error", func() {
 				var firstError *service.Error
 
@@ -212,6 +241,16 @@ var _ = Describe("Standard", func() {
 
 				It("exists", func() {
 					Expect(grandchild).ToNot(BeNil())
+				})
+
+				It("Logger returns a logger", func() {
+					Expect(grandchild.Logger()).ToNot(BeNil())
+				})
+
+				Context("ResolveReference", func() {
+					It("correctly returns the resolved reference", func() {
+						Expect(grandchild.ResolveReference("reference")).To(Equal("/child/grandchild/reference"))
+					})
 				})
 
 				Context("AppendError with a first error", func() {
--- a/data/parser/standard_array.go
+++ b/data/parser/standard_array.go
@@ -15,23 +15,37 @@ import (
 
 	"github.com/tidepool-org/platform/app"
 	"github.com/tidepool-org/platform/data"
+	"github.com/tidepool-org/platform/log"
 	"github.com/tidepool-org/platform/service"
 )
 
 type StandardArray struct {
-	context data.Context
-	array   *[]interface{}
+	context         data.Context
+	array           *[]interface{}
+	parsed          []bool
+	notParsedPolicy NotParsedPolicy
 }
 
-func NewStandardArray(context data.Context, array *[]interface{}) (*StandardArray, error) {
+func NewStandardArray(context data.Context, array *[]interface{}, notParsedPolicy NotParsedPolicy) (*StandardArray, error) {
 	if context == nil {
 		return nil, app.Error("parser", "context is missing")
 	}
 
+	var parsed []bool
+	if array != nil {
+		parsed = make([]bool, len(*array))
+	}
+
 	return &StandardArray{
-		context: context,
-		array:   array,
+		context:         context,
+		array:           array,
+		parsed:          parsed,
+		notParsedPolicy: notParsedPolicy,
 	}, nil
+}
+
+func (s *StandardArray) Logger() log.Logger {
+	return s.context.Logger()
 }
 
 func (s *StandardArray) SetMeta(meta interface{}) {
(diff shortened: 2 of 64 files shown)


data/types/base/base.go
```diff
@@ -17,42 +17,47 @@ import (
...
+func New(baseType string) (*Base, error) {
+	if baseType == "" {
 		return nil, app.Error("base", "type is missing")
 	}
 
 	return &Base{
 		Active:        false,
-		CreatedTime:   time.Now().Format(time.RFC3339),
+		CreatedTime:   time.Now().UTC().Format(time.RFC3339),
+		GUID:          app.NewUUID(),
```
jh-bate: so from memory we started adding GUID as that is what we really wanted instead of the generated ID. Do we still need both in this new world? Or is it just backwards compatibility?

data/types/base/base.go
```diff
@@ -63,41 +68,29 @@ func (b *Base) Meta() interface{} {
...
 	return nil
 }
 
 func (b *Base) Validate(validator data.Validator) error {
-	// validator.ValidateString("createdTime", &b.CreatedTime).LengthGreaterThanOrEqualTo(1)
-	// validator.ValidateString("type", &b.Type).LengthGreaterThanOrEqualTo(1)
-
-	// validator.ValidateString("userId", b.UserID).Exists().LengthGreaterThanOrEqualTo(10)	// TODO_DATA: Validation is for parsed data only
-	validator.ValidateString("deviceId", b.DeviceID).Exists().LengthGreaterThanOrEqualTo(1)
+	// validator.ValidateInterfaceArray("annotations", b.Annotations)    // TODO: Any validations? Optional? Size?
```
jh-bate: so we aim to confirm these `TODO:`'s as part of the testing and before the push to prod?

data/types/base/base_test.go
```diff
@@ -48,26 +48,26 @@ import (
 // TODO_DATA: Need to find another way to test base not using sample/sub
```
jh-bate: maybe we test it as part of another type and just concentrate on the base fields? better than no tests

data/context/standard_test.go
```diff
@@ -6,19 +6,28 @@ import (
...
+	"github.com/tidepool-org/platform/log/test"
 	"github.com/tidepool-org/platform/service"
 )
 
 var _ = Describe("Standard", func() {
-	Context("New", func() {
+	It("NewStandard returns an error if logger is nil", func() {
+		standard, err := context.NewStandard(nil)
+		Expect(standard).To(BeNil())
+		Expect(err).To(HaveOccurred())
```
jh-bate: no need to test the content of that error?

data/context/standard_test.go
```diff
@@ -152,6 +171,16 @@ var _ = Describe("Standard", func() {
 				Expect(child).ToNot(BeNil())
 			})
 
+			It("Logger returns a logger", func() {
+				Expect(child.Logger()).ToNot(BeNil())
```
jh-bate: i guess your only testing that it is not nil, and not that it is actually a logger ??

data/parser/standard_array.go
```diff
@@ -246,23 +276,46 @@ func (s *StandardArray) ParseInterfaceArray(index int) *[]interface{} {
...
+	}
+
+	switch s.notParsedPolicy {
+	case WarnLoggerNotParsed:
+		for index := range *s.array {
+			if !s.parsed[index] {
+				s.Logger().WithField("reference", s.context.ResolveReference(index)).Warn("Reference not parsed")
+			}
+		}
+	case AppendErrorNotParsed:
```
jh-bate: i thought with switch's you always had to specify a default??

data/parser/standard_array_test.go
```diff
@@ -512,14 +532,85 @@ var _ = Describe("StandardArray", func() {
...
+					}, parser.WarnLoggerNotParsed)
+				})
+
+				It("without anything parsed has no errors", func() {
+					standardArray.ProcessNotParsed()
+					Expect(standardContext.Errors()).To(BeEmpty())
+				})
+			})
+
+			Context("with ParsedPolicy as WarnLoggerNotParsed", func() {
```
jh-bate: am I reading this wrong? should it not be `AppendErrorNotParsed`

data/parser/standard_object.go
```diff
@@ -148,12 +172,12 @@ func (s *StandardObject) ParseStringArray(key string) *[]string {
...
 		arrayValue, arrayValueOk := rawValue.([]interface{})
 		if !arrayValueOk {
-			s.context.AppendError(key, ErrorTypeNotArray(rawValue))
+			s.AppendError(key, ErrorTypeNotArray(rawValue))
 			return nil
 		}
 
 		stringArrayValue = []string{}
-		parser, _ := NewStandardArray(s.context.NewChildContext(key), &arrayValue)
+		parser, _ := NewStandardArray(s.context.NewChildContext(key), &arrayValue, IgnoreNotParsed)
```
jh-bate: ignored value?

data/parser/standard_object.go
```diff
@@ -199,11 +227,11 @@ func (s *StandardObject) ParseObjectArray(key string) *[]map[string]interface{}
 	if !objectArrayValueOk {
 		arrayValue, arrayValueOk := rawValue.([]interface{})
 		if !arrayValueOk {
-			s.context.AppendError(key, ErrorTypeNotArray(rawValue))
+			s.AppendError(key, ErrorTypeNotArray(rawValue))
 			return nil
 		}
 
-		parser, _ := NewStandardArray(s.context.NewChildContext(key), &arrayValue)
+		parser, _ := NewStandardArray(s.context.NewChildContext(key), &arrayValue, IgnoreNotParsed)
```
jh-bate: ignored value?

data/parser/standard_object_test.go
```diff
@@ -467,14 +487,80 @@ var _ = Describe("StandardObject", func() {
...
+					}, parser.WarnLoggerNotParsed)
+				})
+
+				It("without anything parsed has no errors", func() {
+					standardObject.ProcessNotParsed()
+					Expect(standardContext.Errors()).To(BeEmpty())
+				})
+			})
+
+			Context("with ParsedPolicy as WarnLoggerNotParsed", func() {
```
jh-bate: wrong name for `parser.AppendErrorNotParsed` ?

data/validator/standard_object_array_test.go
```diff
@@ -153,6 +167,54 @@ var _ = Describe("StandardObjectArray", func() {
 			})
 		})
 
+		Context("new validator with valid reference and empty object array value", func() {
+			var standardObjectArray *validator.StandardObjectArray
+			var result data.ObjectArray
+
+			BeforeEach(func() {
+				value := []map[string]interface{}{}
+				standardObjectArray = validator.NewStandardObjectArray(standardContext, "mummy", &value)
```
jh-bate: go `mummy` :)

jh-bate:
LGTM nothing major - just some comments and questions above

data/types/base/base.go
darinkrauss: Yes, there are a number of TODOs throughout the project. We'll need to at least revisit each one and decide whether it is something that needs to be addressed before prd or not.

data/types/base/base.go
darinkrauss: I added GUID back in because it is listed in the data model docs. The uploader has been adding it for a long while now. We are now generating it in the backend (and the uploader no longer sends it). The ID is definitely for backwards compatibility. At some point we should settle on one or the other and deprecate one of them.

data/types/base/base_test.go
darinkrauss: I think the best thing would be to create a concrete "test" type that uses base and test base through that. One of the TODOs that needs to be addressed before prd.

data/context/standard_test.go
darinkrauss: Ah, good point. However, I just looked and there are a whole bunch of these `HaveOccurred()`. I'll convert all of them at once in another PR.

data/context/standard_test.go
darinkrauss: Yes, but we really don't care what the concrete type is, just that it implements the interface. We know that it implements the interface because if it didn't Go wouldn't compile it.  Therefore, the way I see it, we don't ever need to check something like this. The only time we'd need to check a type is if we need to ensure a concrete type. For example, when parsing some raw upload type data, the `Parse` function returns `data.Datum`. Then, we'd want to validate that it actually returned a concrete `upload.Upload` type. Something more specific that `data.Datum`.

Does that make sense?

data/parser/standard_array.go
darinkrauss: Nope. The "default" default is just to fall through and do nothing.

data/parser/standard_array_test.go
darinkrauss: Actually this one is right, but the one below is incorrect. I will fix.

data/parser/standard_object.go
darinkrauss: Yes, I'm just utilizing the standard array parser here just to convert from an interface array to a string array. I don't want this parsing to indicate anything about whether it was actually parsed by the end code or not.

data/parser/standard_object.go
darinkrauss: Same as above.

data/parser/standard_object_test.go
darinkrauss: This is correct here, but incorrect down below. I will fix.

data/validator/standard_object_array_test.go
darinkrauss: Did you catch the theme here? 😁

PR #36, opened 2016-06-07
More specific error tests

darinkrauss:
@jh-bate I did this last week, but apparently forgot to push it up and have you do a PR. It is only changes to tests.

### DIFF
--- a/data/normalizer/standard_test.go
+++ b/data/normalizer/standard_test.go
@@ -25,7 +25,7 @@ var _ = Describe("Standard", func() {
 	It("New returns an error if context is nil", func() {
 		standard, err := normalizer.NewStandard(nil)
 		Expect(standard).To(BeNil())
-		Expect(err).To(HaveOccurred())
+		Expect(err).To(MatchError("normalizer: context is missing"))
 	})
 
 	Context("new normalizer", func() {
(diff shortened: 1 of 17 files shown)


data/normalizer/standard_test.go
```diff
@@ -36,9 +36,9 @@ var _ = Describe("Standard", func() {
 			var err error
 			standardContext, err = context.NewStandard(test.NewLogger())
 			Expect(standardContext).ToNot(BeNil())
-			Expect(err).ToNot(HaveOccurred())
+			Expect(err).To(Succeed())
```
jh-bate: so a `Succeed()` error just means it empty?

jh-bate:
LGTM, just one question

data/normalizer/standard_test.go
darinkrauss: Yes, `To(Succeed())` is the same as `ToNot(HaveOccurred())` and both mean that the error is nil.

jh-bate: `ToNot(HaveOccurred())` reads better to me

darinkrauss: Okay, potatoes vs. potAtoes to me. I'll change it when I'm in there next.

PR #37, opened 2016-06-09
Add expected normal and update tests

darinkrauss:
@jh-bate Primary functional change is adding the expected normal field for normal boluses. Needed to add a couple of new validations (NotExists/Empty with tests). Also updated the code in testing.go so it checks all errors (parsing and validation).

### DIFF
--- a/data/validator/standard_string_test.go
+++ b/data/validator/standard_string_test.go
@@ -46,11 +46,39 @@ var _ = Describe("StandardString", func() {
 				It("adds the expected error", func() {
 					Expect(standardContext.Errors()).To(HaveLen(1))
 					Expect(standardContext.Errors()[0]).ToNot(BeNil())
-					Expect(standardContext.Errors()[0].Code).To(Equal("value-does-not-exist"))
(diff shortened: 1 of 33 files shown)


data/validator/standard_string_test.go
```diff
@@ -510,6 +612,46 @@ var _ = Describe("StandardString", func() {
 				})
 			})
 
+			Context("NotExists", func() {
```
jh-bate: unsure if there is a way around it but I feel I have seen this ( `Context(""NotExists", func() ...` ) test / check several times before. Same with `Empty`

jh-bate:
LGTM - just wondered about the repeated tests but that might not be applicable to DRY up

data/validator/standard_string_test.go
darinkrauss: Yes, but the NotExists or Empty contexts are within other contexts (where the value being tested is different), so it isn't just a matter of repeated tests. Maybe Shared Examples would work. I'll look into it.

PR #38, opened 2016-06-10
Fix validation for normal, extended, and combination boluses; tests

darinkrauss:
@jh-bate @gniezen I think I have all of the boluses using correct validations and tests.


darinkrauss:
@jh-bate I'm going to merge this so I can deploy. Please provide review when you have a chance.

darinkrauss:
@jh-bate I already merged this one (so I could get a release deployed for Gerrit/Nick), but would happy to have feedback if you have any.

PR #42, opened 2016-06-13
remove redundant `Timezone` and `Reasons`

jh-bate:
@darinkrauss updates due to fields that are redundant for the timechange event 

cc @jebeck

### DIFF
--- a/data/types/base/device/timechange/change.go
+++ b/data/types/base/device/timechange/change.go
@@ -3,11 +3,9 @@ package timechange
 import "github.com/tidepool-org/platform/data"
 
 type Change struct {
-	From     *string   `json:"from,omitempty" bson:"from,omitempty"`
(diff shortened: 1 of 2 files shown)


darinkrauss:
@jh-bate Where are the `timezone` and `reasons` fields redundant from? Are these fields still in the raw data passed up? If so, then we'll at least need to parse then (and drop them on the floor if need be) so we don't get `unparsed` errors.

jebeck:
"redundant" isn't quite the right word. We don't ever push these fields up (I double-checked all the drivers). These fields were created in the data model as a "future-proofing" measure, thinking about a later time-editing tool, but they don't exist in any device data protocols today.

darinkrauss:
Okay, if they aren't used now, then LGTM. :shipit:

PR #44, opened 2016-06-28
Data store cleanup & related

darinkrauss:
@jh-bate This is what I was working on before the offsite. Originally this was just going to be the mongo package updates and tests, but needed to be expanded to 1) eventually allow reading data types directly from mgo, which required, 2) a new way to New and Init datum, which required, 3) various minor changes.

Highly recommend you look at this commit by commit. Each commit is very focused, but the overall branch is varied. Please let me know if you have any questions.


data/types/base/bolus/extended/extended.go
```diff
@@ -28,15 +28,28 @@ func SubType() string {
...
+	return &Extended{}
+}
+
+func Init() *Extended {
+	extended := New()
+	extended.Init()
+	return extended
+}
+
+func (e *Extended) Init() {
```
jh-bate: Just wondering should these be `private`? We call the `func Init() *Extended {...}` which internally calls `func (e *Extended) Init() {...}` but would we ever call `func (e *Extended) Init() {...}` on its own?

data/parser/parser_suite_test.go
```diff
@@ -5,9 +5,136 @@ import (
...
+	"github.com/tidepool-org/platform/log"
+	"github.com/tidepool-org/platform/service"
 )
 
 func TestSuite(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "data/parser")
 }
+
+type AppendErrorInput struct {
```
jh-bate: ahh OK had never thought of adding all this in the `suite` but makes the tests much cleaner and focused

data/parser/standard_array.go
```diff
@@ -182,7 +187,7 @@ func (s *StandardArray) ParseStringArray(index int) *[]string {
 		}
 
 		stringArrayValue = []string{}
-		parser, _ := NewStandardArray(s.context.NewChildContext(index), &arrayValue, IgnoreNotParsed)
+		parser, _ := NewStandardArray(s.context.NewChildContext(index), s.factory, &arrayValue, IgnoreNotParsed)
```
jh-bate: so we don't care about the second returned param? do we need it then?

data/parser/standard_object.go
```diff
@@ -177,7 +182,7 @@ func (s *StandardObject) ParseStringArray(key string) *[]string {
 		}
 
 		stringArrayValue = []string{}
-		parser, _ := NewStandardArray(s.context.NewChildContext(key), &arrayValue, IgnoreNotParsed)
+		parser, _ := NewStandardArray(s.context.NewChildContext(key), s.factory, &arrayValue, IgnoreNotParsed)
```
jh-bate: we don't care about errors here? only used in tests?

jh-bate:
In general looking very solid. A few comments / questions above and once those are addressed then LGTM

data/types/base/bolus/extended/extended.go
darinkrauss: We use the `NewDatum` global func and the `Init` struct func in the data factory. (It uses `NewDatum` first depending upon the type/subType and then invokes `Init` on the newly created object.) We use the `New` and `Init` global funcs in the tests.  So, yes, we use all of these outside of this package.

data/parser/parser_suite_test.go
darinkrauss: Anything that I reuse in more than one test file in the same suite I promote into the suite_test.go file.

data/parser/standard_array.go
darinkrauss: The second return param is an `error` that only occurs if one of the first two arguments is `nil`. As we are already in a working StandardArray object that has already validated those two arguments, we know the call to NewStandardArray cannot fail and so we can ignore the error as it will always be nil.  This is a special case of an object having special internal knowledge of itself and can take certain shortcuts. If `NewStandardArray` was called in any other location then it would be good practice to check the return error.

data/parser/standard_object.go
darinkrauss: Same response as in `standard_array.go`.

darinkrauss:
@jh-bate Responded to all of your comments. Let me know if it is good to go.

jh-bate:
@darinkrauss  :shipit:

PR #45, opened 2016-06-28
Use Mongo version 3+ with Travis; remove environment check

darinkrauss:
@jh-bate I contacted Travis and found out how to force version 3 of Mongo for the tests. Removed now unnecessary environment check in mongo store code.

### DIFF
--- a/data/store/mongo/mongo.go
+++ b/data/store/mongo/mongo.go
@@ -23,7 +23,6 @@ import (
 	"github.com/tidepool-org/platform/data"
 	"github.com/tidepool-org/platform/data/store"
 	"github.com/tidepool-org/platform/data/types/base/upload"
-	"github.com/tidepool-org/platform/environment"
 	"github.com/tidepool-org/platform/log"
 )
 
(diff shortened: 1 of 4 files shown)


data/store/mongo/mongo.go
```diff
@@ -94,11 +90,9 @@ func New(logger log.Logger, environmentReporter environment.Reporter, config *Co
 		return nil, app.ExtError(err, "mongo", "unable to determine build info")
 	}
 
-	if !environmentReporter.IsTest() {
-		if !buildInfo.VersionAtLeast(3) {
-			session.Close()
-			return nil, app.Errorf("mongo", "unsupported mongo build version %s", strconv.Quote(buildInfo.Version))
-		}
+	if !buildInfo.VersionAtLeast(3) {
```
jh-bate: should there be an explicit test for this? I couldn't see one below

jh-bate:
question about also including a test case for mongo version < 3. Once that is either resolved or clarified LGTM

data/store/mongo/mongo.go
darinkrauss: I couldn't find a way to explicitly test this. However, I can confirm this code does work as expected since it failed on Travis when it was using a version 2.* of Mongo (until I updated the Travis config to use a newly install version 3.* of Mongo).

jh-bate:
:shipit: merge away

darinkrauss:
Merged into master via `--ff-only`.

PR #46, opened 2016-07-01
Service package cleanup

darinkrauss:
@jh-bate The first 3 commits are pretty straightforward. The last commit is a refactor and tests for the service package. Some minor bug fixes and error checking. The biggest difference (and the cause for the "large" commit) is splitting the service package into service/context and service/middleware.


service/error.go
```diff
@@ -44,35 +44,3 @@ func (e *Error) WithMeta(meta interface{}) *Error {
 	e.Meta = meta
 	return e
 }
-
-// TODO: Deprecate below Errors struct
```
jh-bate: you answered my question - "why is there code with a Deprecate tag"  :) all clean now

service/context/standard_test.go
```diff
@@ -0,0 +1,337 @@
...
+						},
+					},
+					Meta: &context.Meta{
+						Trace: &context.Trace{
+							Request: "test-request",
+							Session: "test-session",
+						},
+					},
+				}
+				Expect(json.Marshal(jsonResponse)).To(MatchJSON("{" +
```
jh-bate: you should be able to do the JSON like so  `{"page": 1, "fruits": ["apple", "peach"]}`  - maybe this would clean it up a little?

service/middleware/access_log.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/ant0ine/go-json-rest/rest"
+
+	"github.com/tidepool-org/platform/service"
+)
+
+type AccessLog struct{}
+
+const (
+	_LogErrors           = "errors"
```
jh-bate: the `_` means its private?

jh-bate:
:shipit: - couple of comments but looking good

service/context/standard_test.go
darinkrauss: Oh, good call! I made the switch and it is much easier to read.

service/middleware/access_log.go
darinkrauss: Yes, I kept getting confused (and namespace collision errors) between private/package consts and variables versus function local variables. Adding an underscore (which forces it to be private) seemed to make it clear that it was not a function local variable. Not technically the 100% correct way in Go, but since I see value in distinguishing between function local variables and global const/vars this seemed to make sense.

darinkrauss:
@jh-bate I updated the MatchJSON to all use raw string literals.  Good call!

service/middleware/access_log.go
jh-bate: yeah to be honest it seems like a good idea to me

jh-bate:
all good @darinkrauss :shipit:

PR #48, opened 2016-07-21
Allow embedded deviceEvent/status in deviceEvent/alarm; cleanup calculator embedded bolus

darinkrauss:
Minor cleanup and additional error checking in embedded bolus in calculator.

Pull out embedded deviceEvent/status from deviceEvent/alarm, do proper validation and normalization, store reference link (id) to embedded deviceEvent/status into containing deviceEvent/alarm.

@jh-bate @gniezen `@anyone`?

### DIFF
--- a/data/types/base/device/alarm/alarm.go
+++ b/data/types/base/device/alarm/alarm.go
@@ -13,13 +13,18 @@ package alarm
 import (
 	"github.com/tidepool-org/platform/data"
 	"github.com/tidepool-org/platform/data/types/base/device"
+	"github.com/tidepool-org/platform/data/types/base/device/status"
+	"github.com/tidepool-org/platform/data/validator"
 )
 
 type Alarm struct {
 	device.Device `bson:",inline"`
 
 	AlarmType *string `json:"alarmType,omitempty" bson:"alarmType,omitempty"`
-	Status    *string `json:"status,omitempty" bson:"status,omitempty"`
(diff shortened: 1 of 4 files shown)


gniezen:
Oh wow, there's an actual username called `@anyone`. As we're planning on testing this next week, I'll review this while fixing up the Uploader code.

darinkrauss:
Oops, that was accidental... :)

data/types/base/device/alarm/alarm.go
```diff
@@ -54,7 +61,22 @@ func (a *Alarm) Parse(parser data.ObjectParser) error {
 	}
 
 	a.AlarmType = parser.ParseString("alarmType")
-	a.Status = parser.ParseString("status")
+
+	// TODO: This is a bit hacky to ensure we only parse true status data. Is there a better way?
```
gniezen: Probably something for @jh-bate to look into when he gets back?

gniezen:
LGTM. I've verified that removing a required field (e.g. `reason`) from the embedded status will not be accepted, and that alarms and status events appear in Blip with correct id linkage as expected.

data/types/base/device/alarm/alarm.go
darinkrauss: Actually, this is my hacky bit.  ;)  It is fine for now and is technically correct, but just doesn't feel right. I'll look at it again later and see if I can come up with something a bit better.

PR #50, opened 2016-08-01
Allow normal/extended to be equal to expectedNormal/Extended for boluses

darinkrauss:
@gniezen @jh-bate Looking for a quick review so I can release to `int` this afternoon.


gniezen:
LGTM. 🚢

PR #52, opened 2016-08-24
Works towards delete uploads

darinkrauss:
@jh-bate Please review commit-by-commit. Each individual commit should be pretty straightforward to grok.

Still need to do functional tests of the two new APIs. Many of the commits are a work towards that goal. In short, we need to be able to instantiate the entire dataservices in "test" mode *except* not actually running the server. The work is almost there, but ran into a snag with our current config loading mechanism. We can discuss over video later today.

Also, there is a new `migrate_gid_to_uid` tool that will need to be run after deploying the new `tide-whisperer` and `seagull`, but before deploying this new code. The tool adds the `_userId` field to all existing data (if it doesn't already exist) from the `_groupId`. It also does a bunch of sanity checking of the relevant data from the `users`, `seagull`, and `deviceData` collections.


darinkrauss:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;aQLE9plQ&#x2F;16-0-ux-5-dev-3-tst-platform-delete-errant-upload"> &lt;0:ux 5:dev 3:tst&gt; [platform] Delete errant upload</a>

userservices/client/standard_test.go
```diff
@@ -563,98 +567,100 @@ var _ = Describe("Standard", func() {
...
-				Context("with an successful response, but without the requested permission", func() {
+				Context("with an successful response with upload and view permissions", func() {
 					BeforeEach(func() {
 						server.AppendHandlers(
 							ghttp.CombineHandlers(
 								ghttp.VerifyRequest("GET", "/access/target-user-id/request-user-id"),
 								ghttp.VerifyHeaderKV("X-Tidepool-Session-Token", "test-session-token"),
 								ghttp.VerifyBody([]byte{}),
-								ghttp.RespondWith(http.StatusOK, "{\"upload\": {}}", nil)),
+								ghttp.RespondWith(http.StatusOK, "{\"upload\": {}, \"view\": {}}", nil)),
```
jh-bate: I think you can use the cleaner json string format here `{"upload":{},"view":{}}`

userservices/client/standard_test.go
```diff
@@ -563,98 +567,100 @@ var _ = Describe("Standard", func() {
...
-				Context("with an successful response, but without the requested permissions", func() {
+				Context("with an successful response with owner permissions that already includes upload permissions", func() {
 					BeforeEach(func() {
 						server.AppendHandlers(
 							ghttp.CombineHandlers(
 								ghttp.VerifyRequest("GET", "/access/target-user-id/request-user-id"),
 								ghttp.VerifyHeaderKV("X-Tidepool-Session-Token", "test-session-token"),
 								ghttp.VerifyBody([]byte{}),
-								ghttp.RespondWith(http.StatusOK, "{\"upload\": {}}", nil)),
+								ghttp.RespondWith(http.StatusOK, "{\"root\": {\"root-inner\": \"unused\"}, \"upload\": {}}", nil)),
```
jh-bate: ditto json comment above

userservices/client/standard_test.go
```diff
@@ -563,98 +567,100 @@ var _ = Describe("Standard", func() {
...
-				Context("with an successful response and the requested permissions", func() {
+				Context("with an successful response with owner permissions that already includes view permissions", func() {
 					BeforeEach(func() {
 						server.AppendHandlers(
 							ghttp.CombineHandlers(
 								ghttp.VerifyRequest("GET", "/access/target-user-id/request-user-id"),
 								ghttp.VerifyHeaderKV("X-Tidepool-Session-Token", "test-session-token"),
 								ghttp.VerifyBody([]byte{}),
-								ghttp.RespondWith(http.StatusOK, "{\"view\": {}, \"upload\": {}}", nil)),
+								ghttp.RespondWith(http.StatusOK, "{\"root\": {\"root-inner\": \"unused\"}, \"view\": {}}", nil)),
```
jh-bate: ditto json format

userservices/client/standard_test.go
```diff
@@ -563,98 +567,100 @@ var _ = Describe("Standard", func() {
...
-				Context("with an successful response and the root permission", func() {
+				Context("with an successful response with owner permissions that already includes upload and view permissions", func() {
 					BeforeEach(func() {
 						server.AppendHandlers(
 							ghttp.CombineHandlers(
 								ghttp.VerifyRequest("GET", "/access/target-user-id/request-user-id"),
 								ghttp.VerifyHeaderKV("X-Tidepool-Session-Token", "test-session-token"),
 								ghttp.VerifyBody([]byte{}),
-								ghttp.RespondWith(http.StatusOK, "{\"root\": {}}", nil)),
+								ghttp.RespondWith(http.StatusOK, "{\"root\": {\"root-inner\": \"unused\"}, \"upload\": {}, \"view\": {}}", nil)),
```
jh-bate: ditto json format

dataservices/dataservices.go
```diff
@@ -1,13 +1,13 @@
...
- * [ ] Reviewed for concurrency safety
- * [ ] Code complete
- * [ ] Full test coverage
+ * [x] Uses interfaces as appropriate
+ * [x] Private package variables use underscore prefix
+ * [x] All parameters validated
+ * [x] All errors handled
+ * [x] Reviewed for concurrency safety
+ * [x] Code complete
+ * [x] Full test coverage
```
jh-bate: so no tests but full coverage?

data/store/mongo/mongo_test.go
```diff
@@ -94,11 +94,9 @@ var _ = Describe("Mongo", func() {
 		var logger log.Logger
 		var mongoConfig *mongo.Config
 		var mongoStore *mongo.Store
-		var err error
```
jh-bate: what was the issue declaring it once like this?

userservices/client/standard_test.go
```diff
@@ -655,9 +643,7 @@ var _ = Describe("Standard", func() {
 					})
 
 					It("returns an error", func() {
-						permissions, err := standard.GetUserPermissions(context, "request-user-id", "target-user-id")
-						Expect(err).ToNot(HaveOccurred())
-						Expect(permissions).To(Equal(client.Permissions{
+						Expect(standard.GetUserPermissions(context, "request-user-id", "target-user-id")).To(Equal(client.Permissions{
```
jh-bate: counted `"request-user-id"` four times - move to a test constant ?

data/deduplicator/truncate/truncate_test.go
```diff
@@ -126,7 +125,7 @@ var _ = Describe("Truncate", func() {
 			})
 
 			It("returns false if the device id is empty", func() {
-				dataset.DeviceID = StringAsPointer("")
+				dataset.DeviceID = app.StringAsPointer("")
```
jh-bate: which is just `nil`?

store/mongo/mongo_suite_test.go
```diff
@@ -0,0 +1,69 @@
...
+func MongoTestSession() *mgo.Session {
+	return _nodeSession
+}
+
+func MongoTestDatabase() string {
+	return _database
+}
+
+var (
+	_globalSession *mgo.Session
```
jh-bate: trying to remember - why the leading `_`?

dataservices/service/server/config.go
```diff
@@ -58,3 +58,17 @@ func (c *Config) Validate() error {
 	}
 	return nil
 }
+
+func (c *Config) Clone() *Config {
```
jh-bate: more based on the commit title 

> Clone config structs when storing to prevent out-of-band modifications

what was the issue here ?

jh-bate:
@darinkrauss I have gone through this now. A few small comments and questions and once those are done LGTM

userservices/client/standard_test.go
darinkrauss: Doh! Completely zoned on using string literals here. Will update all code to use string literals where there are any escaped double-quotes.

userservices/client/standard_test.go
darinkrauss: Same as above.

userservices/client/standard_test.go
darinkrauss: Same as above.

userservices/client/standard_test.go
darinkrauss: Same as above.

dataservices/dataservices.go
darinkrauss: I'm using "Full test coverage" to mean "as full as is possible". For example, it is not possible to test private functions. In this very specific case, it is not possible to test the `main` function since it actually starts the server, which cannot be tested.

store/mongo/mongo_suite_test.go
darinkrauss: This isn't a Go thing, but just something I started to do for private global variables to distinguish them from local function variables. Impossible to know whether `database` is a local variable or private global variable. If there isn't a way to distinguish the two then it is very easy to accidentally shadow a private global variable with a local variable.  For example:

```
var myVar string = "global"

func SetMyVar(newVar string) {
  myVar := newVar
}
```

Obviously a too-simple example (and it warn on not using the local variable `myVar`), but hopefully you get the idea.

In any case, just something I started to keep me straight on private global variables. Besides, there are *very* few, if any, of these in the code since our code has to be safe in a concurrent environment.

darinkrauss:
@jh-bate Thanks! I'll have some additional commits coming to wrap up the testing.

data/store/mongo/mongo_test.go
darinkrauss: Not really an issue. Just getting in the habit of declaring variables as close to their use as possible. In this case, the err variable is essentially local to each specific test function so should be declared there. The other variables are used in the `BeforeEach`  as well as each specific test function so need to be declared in the outer scope.

userservices/client/standard_test.go
darinkrauss: Yea, I could, but I've tried to avoid test constants. In many cases constants in tests just obfuscate what the test is trying to accomplish. That is, I've tried to make tests as self-contained as possible. So, read just the code within the `It` block and any related `BeforeEach` blocks and it fully identifies the test. Using test constants would make one more place to look for the meaning of the test.

So, no strong reason, but just the way I've done things.

data/deduplicator/truncate/truncate_test.go
darinkrauss: No, a pointer to an empty string is not the same as nil. You can dereference a pointer to an empty string and get the string. If you dereference a nil, then you panic.

dataservices/service/server/config.go
darinkrauss: You create the config, pass the config into some object (MongoStore) that remembers the config, then you change the config. You've inadvertently changed the MongoStore by changing the config *after* giving it to the MongoStore. I ran into this while writing the `migrate_gid_to_uid` tool because I was using the same config (with the mongo address) for multiple mongo stores (user, seagull, data).

jh-bate:
LGTM @darinkrauss

darinkrauss:

darinkrauss:
Rebased onto `master` branch.

PR #53, opened 2016-08-29
Break deploy into separate packages for different services and tools

darinkrauss:
@jh-bate Instead of just a platform-wide artifact (eg. `platform-v1.2.3.tar.gz`) pushed to S3, there will be one per deploy (eg. `dataservices-v1.2.3.tar.gz`, `userservices-v1.2.3.tar.gz` and `tools-v1.2.3.tar.gz`). This allows us to deploy into separate `shio` slots for `dataservices`, `userservices`, and `tools`. This is the solution to deploying tools separate from the executable services (and enabling userservices).


darinkrauss:

jh-bate:
LGTM @darinkrauss

PR #54, opened 2016-08-30
Additional sanity checks and logging for migrate_gid_to_uid

darinkrauss:
@jh-bate What it says on the tin

### DIFF
--- a/tools/migrate_gid_to_uid/migrate_gid_to_uid.go
+++ b/tools/migrate_gid_to_uid/migrate_gid_to_uid.go
@@ -188,6 +188,7 @@ func initializeLogger(versionReporter version.Reporter, config *Config) (log.Log
 func buildMetaIDToUserIDMap(logger log.Logger, config *Config) (map[string]string, error) {
 	logger.Debug("Building meta id to user id map")
 
+	userIDMap := map[string]bool{}
 	metaIDToUserIDMap := map[string]string{}
 
 	logger.Debug("Creating users store")
@@ -222,13 +223,35 @@ func buildMetaIDToUserIDMap(logger log.Logger, config *Config) (map[string]strin
 		} `bson:"private"`
 	}
 	for iter.Next(&result) {
-		if userID := result.UserID; userID == "" {
-			logger.Warn("Missing user id in result from users query")
-		} else if metaID := result.Private.Meta.ID; metaID == "" {
-			logger.WithField("userId", userID).Warn("Missing private meta id in result from users query for user id")
-		} else {
-			metaIDToUserIDMap[metaID] = userID
+		userLogger := logger
+
+		userID := result.UserID
+		if userID == "" {
+			userLogger.Warn("Missing user id in result from users query")
+			continue
 		}
+
+		userLogger = userLogger.WithField("userId", userID)
+
+		if _, ok := userIDMap[userID]; ok {
+			userLogger.Error("Found multiple users with same user id")
+			continue
+		}
+		userIDMap[userID] = true
+
+		metaID := result.Private.Meta.ID
+		if metaID == "" {
+			userLogger.Warn("Missing private meta id in result from users query for user id")
+			continue
+		}
+
+		userLogger = userLogger.WithField("metaId", metaID)
+
+		if _, ok := metaIDToUserIDMap[metaID]; ok {
+			userLogger.Error("Found multiple users with same meta id")
+			continue
+		}
+		metaIDToUserIDMap[metaID] = userID
 	}
 	if err = iter.Close(); err != nil {
 		return nil, app.ExtError(err, "migrate_gid_to_uid", "unable to iterate users")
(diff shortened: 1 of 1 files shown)


tools/migrate_gid_to_uid/migrate_gid_to_uid.go
```diff
@@ -222,13 +223,35 @@ func buildMetaIDToUserIDMap(logger log.Logger, config *Config) (map[string]strin
...
-			logger.Warn("Missing user id in result from users query")
-		} else if metaID := result.Private.Meta.ID; metaID == "" {
-			logger.WithField("userId", userID).Warn("Missing private meta id in result from users query for user id")
-		} else {
-			metaIDToUserIDMap[metaID] = userID
+		userLogger := logger
+
+		userID := result.UserID
+		if userID == "" {
+			userLogger.Warn("Missing user id in result from users query")
```
jh-bate: is it an issue you warn there is no userID but continue and then next statement is setting up logger with nonexistent userID?

tools/migrate_gid_to_uid/migrate_gid_to_uid.go
```diff
@@ -222,13 +223,35 @@ func buildMetaIDToUserIDMap(logger log.Logger, config *Config) (map[string]strin
...
+			continue
+		}
+
+		userLogger = userLogger.WithField("userId", userID)
+
+		if _, ok := userIDMap[userID]; ok {
```
jh-bate: wont this also be `ok` if there is just one user with the same ID?

tools/migrate_gid_to_uid/migrate_gid_to_uid.go
```diff
@@ -272,28 +296,60 @@ func buildGroupIDToUserIDMap(logger log.Logger, config *Config, metaIDToUserIDMa
...
-			}
+		metaLogger := logger
+
+		metaID := result.MetaID
+		if metaID == "" {
+			metaLogger.Warn("Missing meta id in result from meta query")
+			continue
+		}
+
+		metaLogger = metaLogger.WithField("metaId", metaID)
```
jh-bate: same 'issue' as userID - if the metaID is missing and we then setup the logger with an empty metaID will that be confusing?

tools/migrate_gid_to_uid/migrate_gid_to_uid.go
```diff
@@ -272,28 +296,60 @@ func buildGroupIDToUserIDMap(logger log.Logger, config *Config, metaIDToUserIDMa
...
+
+		var value struct {
+			Private struct {
+				Uploads struct {
+					ID string `json:"id"`
+				} `json:"uploads"`
+			} `json:"private"`
+		}
+		if err = json.Unmarshal([]byte(result.Value), &value); err != nil {
+			metaLogger.WithError(err).Warn("Unable to unmarshal value from meta query for meta id")
```
jh-bate: it's an error but we warn? is that the intention?

tools/migrate_gid_to_uid/migrate_gid_to_uid.go
darinkrauss: In this case, `continue` means do the next iteration of the `for` loop. So, the code below is never executed if the `continue` is hit.

tools/migrate_gid_to_uid/migrate_gid_to_uid.go
darinkrauss: This is looking for whether we already have found the same `userID`. If we've already seen it, then we have a duplicate, so log the error and continue on with the next iteration of the `for` loop.

tools/migrate_gid_to_uid/migrate_gid_to_uid.go
darinkrauss: Same explanation as above. :)

tools/migrate_gid_to_uid/migrate_gid_to_uid.go
jh-bate: ahh yeah - I was thinking `break`

tools/migrate_gid_to_uid/migrate_gid_to_uid.go
jh-bate: scratch based on above comment with regards to `userID`

tools/migrate_gid_to_uid/migrate_gid_to_uid.go
darinkrauss: It is a warning simply because there is already "bogus" data in our deployed databases (presumably from earlier incarnations of the code). So, overall in this script I use `Error` to indicate that something has failed and the script wasn't able to do its work and needs to be resolved, while `Warn` is for when something is not right, but can happen and not something that needs to be resolved.

jh-bate:
LGTM @darinkrauss

darinkrauss:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;TUqPEQif&#x2F;24-0-ux-2-dev-1-tst-platform-add-user-id-to-all-device-data-records-in-addition-to-existing-group-id"> &lt;0:ux 2:dev 1:tst&gt; [platform] Add user id to all device data records (in addition to existing group id)</a>

darinkrauss:

PR #55, opened 2016-09-08
Delete uploads with pagination

darinkrauss:
@jh-bate Wraps up all of the work for deleting errant uploads. I recommend going through commit-by-commit. Each commit will be very focused and generally small. Many of the commits aren't directly related to deleting errant uploads, but are general refinements to the platform to support multiple services and tools per our discussion yesterday.

As a side note, with the split of the platform into three separate deployments, we'll need to update the styx config and the uploader. Fortunately, we can deploy the styx config change, then the three separate deployments, and then update the uploader. There are no deployment timing issues or dependencies. That is, the uploader can be updated soonish, but not immediately.


darinkrauss:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;aQLE9plQ&#x2F;16-0-ux-5-dev-3-tst-platform-delete-errant-upload"> &lt;0:ux 5:dev 3:tst&gt; [platform] Delete errant upload</a>

dataservices/service/service/standard.go
```diff
@@ -172,6 +172,30 @@ func (s *Standard) initializeLogger() error {
 	return nil
 }
 
+func (s *Standard) initializeUserServicesClient() error {
+	s.logger.Debug("Loading user services client config")
```
jh-bate: are these debug statements really of use?

dataservices/service/api/standard.go
```diff
@@ -23,24 +21,20 @@ import (
...
 	userServicesClient      client.Client
 	dataFactory             data.Factory
 	dataStore               store.Store
 	dataDeduplicatorFactory deduplicator.Factory
-	api                     *rest.Api
-	statusMiddleware        *rest.StatusMiddleware
 }
 
-func NewStandard(versionReporter version.Reporter, environmentReporter environment.Reporter, logger log.Logger, userServicesClient client.Client, dataFactory data.Factory, dataStore store.Store, dataDeduplicatorFactory deduplicator.Factory, routes []service.Route) (*Standard, error) {
+func NewStandard(versionReporter version.Reporter, environmentReporter environment.Reporter, logger log.Logger, userServicesClient client.Client, dataFactory data.Factory, dataStore store.Store, dataDeduplicatorFactory deduplicator.Factory) (*Standard, error) {
```
jh-bate: I don't think we need to adjust now but just wondering if it would be easier for these to list the params on a new line for readability?

dataservices/service/service/standard.go
```diff
@@ -109,184 +73,147 @@ func (s *Standard) Initialize() error {
...
-	s.versionReporter = versionReporter
-
-	return nil
-}
-
-func (s *Standard) initializeEnvironmentReporter() error {
-	environmentReporter, err := environment.NewDefaultReporter("TIDEPOOL")
-	if err != nil {
-		return app.ExtError(err, "dataservices", "unable to create environment reporter")
+func (s *Standard) Terminate() {
```
jh-bate: so Close is now Terminate?

dataservices/service/api/standard.go
darinkrauss: Yea, good idea. I've been avoiding that, but I think this line is getting a bit ridiculous.  Partly, I've been thinking about other ways to do this other than a long function signature.

dataservices/service/service/standard.go
darinkrauss: They can be useful if we ever have a startup problem on a deployed environment.

dataservices/service/service/standard.go
darinkrauss: Yea, to counter the "Initialize".  Silly cleanup, I guess.  :)

metricservices/client/client.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/tidepool-org/platform/userservices/client"
+)
+
+type Context interface {
+	Logger() log.Logger
+	Request() *rest.Request
+	AuthenticationDetails() client.AuthenticationDetails
+}
+
+type Client interface {
```
jh-bate: for some reason I thought the linting complained about things like `client. Client` ??

userservices/client/standard.go
```diff
@@ -218,16 +218,16 @@ func (s *Standard) GetUserGroupID(context service.Context, userID string) (strin
 // AND bad session token. Since a bad server token is unlikely (though possible) we MUST assume
 // that it means bad session token.
 
-func (s *Standard) sendRequest(context service.Context, method string, url string, responseObject interface{}) error {
+func (s *Standard) sendRequest(context service.Context, requestMethod string, requestURL string, responseObject interface{}) error {
```
jh-bate: why change from `method` to `requestMethod` ?? We are in the `sendRequest(...`

jh-bate:
a few comments but LGTM @darinkrauss

metricservices/client/client.go
darinkrauss: No, I don't think so. Looking through the Go docs and recommendations, what they discourage is something like `http.HttpServer`, instead preferring `http.Server`. In the case where the object is the same as the package (like `user.User`) then that is okay. At least, that is my interpretation.

userservices/client/standard.go
darinkrauss: Hmm, trying to remember. I believe it was for consistency sake with another sendRequest is either the metricservices or dataservices client.

PR #56, opened 2016-09-13
Delete user data; preparation for delete user

darinkrauss:
@jh-bate Per our conversation last week, this PR includes just deleting the data for a user (in dataservices); plus some minor preparation for full on delete user.

Each commit still is complete by itself and includes comments on exactly what it includes.


jh-bate:
@darinkrauss LGTM

PR #57, opened 2016-09-15
Dataservices client

darinkrauss:
@jh-bate Dataservices client so that the delete user API call can into the dataservices and delete all of the data for the user


darinkrauss:
Closed because the Travis build is messed.

PR #58, opened 2016-09-15
Dataservices client

darinkrauss:
@jh-bate Dataservices client so that the delete user API call can into the dataservices and delete all of the data for the user

### DIFF
--- a/dataservices/client/config.go
+++ b/dataservices/client/config.go
@@ -1,0 +1,45 @@
+package client
+
+/* CHECKLIST
+ * [ ] Uses interfaces as appropriate
+ * [ ] Private package variables use underscore prefix
+ * [ ] All parameters validated
+ * [ ] All errors handled
+ * [ ] Reviewed for concurrency safety
+ * [ ] Code complete
+ * [ ] Full test coverage
+ */
+
+import (
+	"net/url"
+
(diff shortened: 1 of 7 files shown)


dataservices/client/config.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type Config struct {
+	Address        string `json:"address"`
+	RequestTimeout int    `json:"requestTimeout"`
+}
+
+func (c *Config) Validate() error {
+	if c.Address == "" {
+		return app.Error("client", "address is missing")
```
jh-bate: do these give enough info so we know which client? .... I guess they do in the stack trace don't they

darinkrauss: At this level of code, no. However, the overall error reporting design is that if an error occurs, each "level" of code adds its own information.  

So, if there is a failure here, the resulting error message in the log might look like: 

`service: unable to create user services client; client: config is invalid; client: address is missing`

Furthermore, all log entries contain fields like: 

`"file":"userservices/service/service/standard.go","line": 134,"pid":12345,"process":"dataservices","version":"0.1.4+1669a7d"`

So, with all of this information (even if we have one massive log file for everything), you can isolate it down to the exact file/line the original error occurred and under what conditions.

darinkrauss:

PR #59, opened 2016-09-15
Various stores

darinkrauss:
@jh-bate Add all of the various stores so delete user can delete all pertinent user data

### DIFF
--- a/dataservices/service/api/standard.go
+++ b/dataservices/service/api/standard.go
@@ -16,13 +16,14 @@ import (
 	"github.com/tidepool-org/platform/app"
 	"github.com/tidepool-org/platform/data"
 	"github.com/tidepool-org/platform/data/deduplicator"
-	"github.com/tidepool-org/platform/data/store"
+	dataStore "github.com/tidepool-org/platform/data/store"
 	"github.com/tidepool-org/platform/dataservices/service"
 	"github.com/tidepool-org/platform/dataservices/service/context"
 	"github.com/tidepool-org/platform/environment"
 	"github.com/tidepool-org/platform/log"
 	metricservicesClient "github.com/tidepool-org/platform/metricservices/client"
 	"github.com/tidepool-org/platform/service/api"
+	taskStore "github.com/tidepool-org/platform/task/store"
 	userservicesClient "github.com/tidepool-org/platform/userservices/client"
 	"github.com/tidepool-org/platform/version"
 )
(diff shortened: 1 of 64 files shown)


dataservices/service/api/standard.go
```diff
@@ -32,11 +33,15 @@ type Standard struct {
...
 	userServicesClient      userservicesClient.Client
 	dataFactory             data.Factory
-	dataStore               store.Store
 	dataDeduplicatorFactory deduplicator.Factory
+	dataStore               dataStore.Store
+	taskStore               taskStore.Store
 }
 
-func NewStandard(versionReporter version.Reporter, environmentReporter environment.Reporter, logger log.Logger, metricServicesClient metricservicesClient.Client, userServicesClient userservicesClient.Client, dataFactory data.Factory, dataStore store.Store, dataDeduplicatorFactory deduplicator.Factory) (*Standard, error) {
+func NewStandard(versionReporter version.Reporter, environmentReporter environment.Reporter, logger log.Logger,
```
jh-bate: I still wonder if these would be cleanest with one pram per line 

```
   NewStandard(
      versionReporter version.Reporter, 
      environmentReporter environment.Reporter, 
      logger log.Logger,
      ....
```

darinkrauss: I tried that out, but it added a lot of lines to the code. Plus, if we went to one parameter per line, does that mean we need to change all of the code to reflect that.  I figured it made more sense to just start wrapping the line at some reasonable length. That way, if a function had only 2 or 3 parameters it could remain on the same line.

darinkrauss:
I think what each new store currently has is the absolute minimum, a custom store object that creates a custom session object that will (eventually) have custom functions for different data access.

Regarding the constructors with large number of parameters, I'm not particularly fond of them and have been trying to come up with a better way. We could do one parameter per line in the constructor, but my comments above address that.  In addition, we could do away with most of the parameters and have a bunch of setters for each of those related interfaces.  For example,

```
standard, err := standard.NewStandard(versionReporter, environmentReporter, logger)
if err != nil {
    return err
}
standard.SetMetricservicesClient(metricservicesClient)
standard.SetUserservicesClient(userservicesClient)
standard.SetDataFactory(dataFactory)
standard.SetDataStore(dataStore)
standard.SetDataDeduplicatorFactory(dataDeduplicatorFactory)
```

However, this pattern allows you to create an object that is 1/2-built and wouldn't function correctly (eg. doesn't have a data factory or data store). You'd almost need an `IsValid` call on the struct to ensure it is setup correctly before continuing on.  (Imagine where a new related interface is added to the struct, but we forget to inject it at creation time. With a constructor it will fail to compile, but with the above setters, not so.)

Another thought was to wrap all of the related objects into a config struct.  For example:

```
config := {
    versionReporter: versionReporter,
    environmentReporter: environmentReporter,
    logger: logger,
    metricservicesClient: metricServicesClient,
    userservicesClient: userservicesClient,
   ...
}
standard, err := standard.NewStandard(config)
if err != nil {
    return err
}
```

I think this comes with its own set of issues.

So far, I'm not happy with any of them so still thinking about it.

Do you have any ideas/thoughts?

jh-bate:
Yeah I agree with your comments above.

darinkrauss:

PR #64, opened 2016-09-27
Add Agent to global store session

darinkrauss:
@jh-bate The agent is the user ID that is authenticated (or empty string if server) on the request. The agent is used for all `createdUserID`, `modifiedUserID`, and `deleteUserID` fields, as appropriate.


darinkrauss:

PR #65, opened 2016-09-27
Destroy tasks with data; use ByID suffix for consistency/clarity

darinkrauss:
@jh-bate Rather than creating a separate endpoint for destroying tasks, I just piggy-backed it on to destroying data. Tasks (or data syncTasks) need to be generalized into either a "background job" mechanism or a message/event based system.

Also, for all store functions that operate on an object id, I changed the function name to have the "ByID" suffix so it is clear and consistent what the function does.

### DIFF
--- a/data/store/store.go
+++ b/data/store/store.go
@@ -27,15 +27,15 @@ type Store interface {
 type Session interface {
 	store.Session
 
-	GetDatasetsForUser(userID string, filter *Filter, pagination *Pagination) ([]*upload.Upload, error)
-	GetDataset(datasetID string) (*upload.Upload, error)
+	GetDatasetsForUserByID(userID string, filter *Filter, pagination *Pagination) ([]*upload.Upload, error)
+	GetDatasetByID(datasetID string) (*upload.Upload, error)
 	CreateDataset(dataset *upload.Upload) error
 	UpdateDataset(dataset *upload.Upload) error
 	DeleteDataset(dataset *upload.Upload) error
 	CreateDatasetData(dataset *upload.Upload, datasetData []data.Datum) error
 	ActivateDatasetData(dataset *upload.Upload) error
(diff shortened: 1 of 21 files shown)


data/store/store.go
```diff
@@ -27,15 +27,15 @@ type Store interface {
...
+	GetDatasetsForUserByID(userID string, filter *Filter, pagination *Pagination) ([]*upload.Upload, error)
+	GetDatasetByID(datasetID string) (*upload.Upload, error)
 	CreateDataset(dataset *upload.Upload) error
 	UpdateDataset(dataset *upload.Upload) error
 	DeleteDataset(dataset *upload.Upload) error
 	CreateDatasetData(dataset *upload.Upload, datasetData []data.Datum) error
 	ActivateDatasetData(dataset *upload.Upload) error
 	DeleteOtherDatasetData(dataset *upload.Upload) error
-	DeleteDataForUser(userID string) error
+	DestroyDataForUserByID(userID string) error
```
jh-bate: so `Destory` vs `Delete`  should it not all just be Delete for consistency?

dataservices/service/api/v1/users_data_delete_test.go
```diff
@@ -23,21 +23,24 @@ var _ = Describe("UsersDataDelete", func() {
 			context = NewTestContext()
 			context.RequestImpl.PathParams["userid"] = targetUserID
 			context.AuthenticationDetailsImpl.IsServerOutputs = []bool{true}
-			context.DataStoreSessionImpl.DeleteDataForUserOutputs = []error{nil}
+			context.DataStoreSessionImpl.DestroyDataForUserByIDOutputs = []error{nil}
```
jh-bate: if we are infact switching to `Destroy` rather than `Delete` then seems the file should then also become `users_data_destroy_test`

darinkrauss:

PR #66, opened 2016-09-28
GetDatasetByID returns nil (rather than error) if dataset not found

darinkrauss:
@jh-bate Per our Slack conversation, all Get(Object)ByID store functions (that intend to return a single object) now return nil for the object (and no error) if the object doesn't exist.

Also, removed all usage of the `One` mgo function as it returns error if there is no object. Instead, use the `All` function with `Limit(2)` to handle 0, 1, or > 1 objects in existence. If 0, then returns nil; if 1, then returns the object; if > 1, then returns the object and warns about multiples for the same id. (In a perfect world, where we had a unique index on the object ids, this could never happen, but we aren't there yet. ;) )

### DIFF
--- a/data/store/mongo/mongo.go
+++ b/data/store/mongo/mongo.go
@@ -109,20 +109,27 @@ func (s *Session) GetDatasetByID(datasetID string) (*upload.Upload, error) {
 
 	startTime := time.Now()
 
-	var dataset upload.Upload
+	datasets := []*upload.Upload{}
 	selector := bson.M{
 		"uploadId": datasetID,
 		"type":     "upload",
 	}
-	err := s.C().Find(selector).One(&dataset)
+	err := s.C().Find(selector).Limit(2).All(&datasets)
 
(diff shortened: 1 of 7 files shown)


data/store/mongo/mongo.go
```diff
@@ -109,20 +109,27 @@ func (s *Session) GetDatasetByID(datasetID string) (*upload.Upload, error) {
...
 	s.Logger().WithFields(loggerFields).WithError(err).Debug("GetDatasetByID")
 
 	if err != nil {
 		return nil, app.ExtError(err, "mongo", "unable to get dataset by id")
 	}
-	return &dataset, nil
+
+	if datasetsCount := len(datasets); datasetsCount == 0 {
+		return nil, nil
+	} else if datasetsCount > 1 {
```
jh-bate: I know its *unlikely* but if there are more than one datasets return we just warn but still return only the first? Is this OK?

darinkrauss:

PR #67, opened 2016-09-29
AES-256 encryption using passphrase for permissions group id generation

darinkrauss:
@jh-bate

### DIFF
--- a/app/crypto.go
+++ b/app/crypto.go
@@ -1,0 +1,79 @@
+package app
+
+/* CHECKLIST
+ * [x] Uses interfaces as appropriate
+ * [x] Private package variables use underscore prefix
+ * [x] All parameters validated
+ * [x] All errors handled
+ * [x] Reviewed for concurrency safety
+ * [x] Code complete
+ * [x] Full test coverage
+ */
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/md5"
+)
+
+func EncryptWithAES256UsingPassphrase(bytes []byte, passphrase []byte) ([]byte, error) {
+	if len(bytes) == 0 {
+		return nil, Error("app", "bytes is missing")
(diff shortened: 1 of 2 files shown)


app/crypto.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func passphraseToKey32AndIV16(passphrase []byte) ([]byte, []byte) {
+	keyAndIV := make([]byte, 48)
+	hashed := []byte{}
+	for i := 0; i < 3; i++ {
+		hashed = hash(hashed, passphrase)
+		copy(keyAndIV[i*16:], hashed)
+	}
+	return keyAndIV[:32], keyAndIV[32:]
```
jh-bate: another dumb question: so in order to guarantee we are in fact using the AES256 for encryption do you feel we assert well enough in the tests that we are getting 32 bytes here?

> NewCipher creates and returns a new cipher.Block. The key argument should be the AES key, either 16, 24, or 32 bytes to select AES-128, AES-192, or AES-256.

app/crypto.go
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/md5"
+)
+
+func EncryptWithAES256UsingPassphrase(bytes []byte, passphrase []byte) ([]byte, error) {
+	if len(bytes) == 0 {
```
jh-bate: after discussion with @darinkrauss the bytes being short is not an issue and also I  understand this is a utility that the passphrase would be coming from config and if we wanted to restrict we would place the check elsewhere

app/crypto.go
jh-bate: sounds good to me @darinkrauss after discussion

darinkrauss:

PR #76, opened 2016-10-05
Add new user-related features to tapi tool; minor refactoring

darinkrauss:
@jh-bate Primarily updating the tapi tool to add delete user functionality. Also moved over add/remove role (for clinics). Minor refactoring to improve trace request carry-through for outgoing requests.

### DIFF
--- a/service/middleware/trace.go
+++ b/service/middleware/trace.go
@@ -23,7 +23,7 @@ const (
 	_LogTraceRequest = "trace-request"
 	_LogTraceSession = "trace-session"
 
-	_TraceSessionMaximumLength = 64
+	_TraceMaximumLength = 64
 )
 
 func NewTrace() (*Trace, error) {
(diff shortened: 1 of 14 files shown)


darinkrauss:

service/middleware/trace.go
```diff
@@ -45,26 +45,34 @@ func (l *Trace) MiddlewareFunc(handler rest.HandlerFunc) rest.HandlerFunc {
 
 			newLogger := oldLogger
 
-			newTraceRequest := app.NewID()
+			newTraceRequest := request.Header.Get(service.HTTPHeaderTraceRequest)
+			if newTraceRequest != "" {
+				if len(newTraceRequest) > _TraceMaximumLength {
```
jh-bate: why could it be longer?

darinkrauss: Now that a request can get the "trace request" id from an HTTP header it could be sent in from an outside caller. That means it is technically an attack vector. It goes along with the "validate all input" line of thinking. Under normal circumstances it will never be greater than the maximum length, but this is protecting from non-normal circumstances.

PR #80, opened 2016-10-19
Clinic accounts should not be able to be deleted

darinkrauss:
@jh-bate Per discussion with Brandon and Howard, we don't want to accidentally delete a clinic account (as it would create a whole range of complicated issues). Therefore, I updated the code to prevent this. If a clinic account really needs to be deleted, then we'll have to delete all of the custodial accounts first, remove the clinic role, and then delete the account.


darinkrauss:

PR #81, opened 2016-10-19
Update tools to use correct environment variable for server secret

darinkrauss:
@jh-bate I will have to update the config as well, but wanted to make it consistent.


darinkrauss:
[[platform] [medtronic] Add validation for rest of data model fields](https://trello.com/c/IJNRnTIW/15-platform-medtronic-add-validation-for-rest-of-data-model-fields)

PR #82, opened 2016-10-19
Update tools to use correct environment variable for server secret

darinkrauss:
@jh-bate Screwed up the ordering of the merges. Can you reapprove this one?


darinkrauss:
[[platform] [medtronic] Add validation for rest of data model fields](https://trello.com/c/IJNRnTIW/15-platform-medtronic-add-validation-for-rest-of-data-model-fields)

PR #86, opened 2016-10-25
Clean log statements; use camelCase for keys; only report ids, not objects

darinkrauss:
@jh-bate Quick cleanup of logs for consistency. I noticed this while adding new log statements.  Use camelCase for keys since this is our JSON standard. Do not write full objects to log as it may contain user data, only write object ids.


darinkrauss:

PR #87, opened 2016-10-25
Refactor common data store session and datum mocks; minor field changes

darinkrauss:
@jh-bate Will need to add additional functions to the data.Datum interface so rather that doing it in three places, decided it was time to move the three separate data.Datum mocks to a common implementation. Use common mock data store session in API tests. Minor rename of mock input/output field names for consistence.


darinkrauss:

PR #96, opened 2016-10-25
Keep 5 decimal places on converting mg/dL to mmol/L (to do otherwise implies greater accuracy than is available)

darinkrauss:
@jh-bate What it says above.

### DIFF
--- a/data/blood/glucose/glucose.go
+++ b/data/blood/glucose/glucose.go
@@ -30,6 +30,7 @@ const (
 	MgdLUpperLimit float64 = 1000.0
 
 	MmolLToMgdLConversionFactor float64 = 18.01559
+	MmolLToMgdLPrecisionFactor  float64 = 100000.0
 )
 
 func Units() []string {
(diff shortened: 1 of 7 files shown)


data/blood/glucose/glucose.go
```diff
@@ -62,7 +63,9 @@ func NormalizeValueForUnits(value *float64, units *string) *float64 {
 	if value != nil && units != nil {
 		switch *units {
 		case MgdL, Mgdl:
-			return app.FloatAsPointer(*value / MmolLToMgdLConversionFactor)
+			intValue := int(*value/MmolLToMgdLConversionFactor*MmolLToMgdLPrecisionFactor + 0.5)
```
jh-bate: the `0.5`?

darinkrauss: It is just to force the integer value to round (rather than floor). Do you think it needs a constant?

jh-bate: makes sense - maybe just a short comment?

darinkrauss: Okay, will do.

darinkrauss:

PR #100, opened 2016-10-25
Add data identity fields to be used by hash generation for de-duplication

darinkrauss:
@jh-bate The identity fields are the values for each data that are used to generate a unique hash for de-duplication. These are similar to `jellyfish`, but order is slightly different and one or two additional fields for ensuring uniqueness. Add full tests for `bolus` and `device` data.


darinkrauss:

PR #111, opened 2017-03-01
Add migrate_pmid_to_uid tool (for Seagull metadata migration)

darinkrauss:
@jh-bate This script migrates the Seagull metadata database from using the `_id` (private.meta.id) to `userId` as its primary key. It would run before (and after to ensure everything was converted) deploying the next version of Seagull.

### DIFF
--- a/tools/migrate_pmid_to_uid/README.md
+++ b/tools/migrate_pmid_to_uid/README.md
@@ -1,0 +1,46 @@
+## migrate_pmid_to_uid
+
+This tool will migrate all Seagull metadata using the _id field and add a userId field containing the associated user id.
+
+This tool *MUST* be executed immediately before and immediately after upgrading `seagull` to version `v0.3.1`.
+
+To execute this tool against a local development setup:
+
+1. Prepare the environment and build the executables. For more information, please see the main README.md.
+
+  ```
+  source .env
+  make build
+  ```
+
+1. Run the tool against the Mongo database running on `localhost`.
+
+  ```
+  _bin/tools/migrate_pmid_to_uid/migrate_pmid_to_uid -a localhost
+  ```
+
+  There may be a number of warnings if the local development setup has partially created accounts and/or old data.
+
+  There should be a final log message along the lines of:
+
+  ```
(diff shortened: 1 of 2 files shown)


tools/migrate_pmid_to_uid/migrate_pmid_to_uid.go
```diff
@@ -0,0 +1,366 @@
+package main
+
+/* CHECKLIST
```
jh-bate: do we need to update the checklist or is it not relevant?

tools/migrate_pmid_to_uid/README.md
```diff
@@ -0,0 +1,46 @@
+## migrate_pmid_to_uid
```
jh-bate: do you think we should add the reason or link to the details as to why we are doing this? Or is it a case of once its done it no longer matters?

darinkrauss: It is a "done once, never again" migration.

However, what we really need is some sort of overall changelog that would detail these things for anyone with their own setup. Not specific per repo, I guess, as that would most likely not be seen. Basically, some reliable way to communicate with anyone using the code.

tools/migrate_pmid_to_uid/migrate_pmid_to_uid.go
darinkrauss: I'll leave it as-is. Probably doesn't belong in this file, but all of the files probably should be rev'd to remove the checklist.

jh-bate: ok

tools/migrate_pmid_to_uid/README.md
jh-bate: true. ok i guess that com's method goes on the platform backlog then :)

darinkrauss:
[[backend] Remove metadata dependency upon user private meta id](https://trello.com/c/7rsj25Mi/223-backend-remove-metadata-dependency-upon-user-private-meta-id)

PR #112, opened 2017-03-01
Update migrate_pmid_to_uid for additional debugging information

darinkrauss:
@jh-bate Added a bit of debugging information


darinkrauss:
[[backend] Remove metadata dependency upon user private meta id](https://trello.com/c/7rsj25Mi/127-backend-remove-metadata-dependency-upon-user-private-meta-id)

PR #113, opened 2017-03-02
Optionally add unique index after migration; update log messages

darinkrauss:
@jh-bate Fix a chicken-and-egg problem @krystophv ran into.


darinkrauss:
[[backend] Remove metadata dependency upon user private meta id](https://trello.com/c/7rsj25Mi/127-backend-remove-metadata-dependency-upon-user-private-meta-id)

PR #122, opened 2017-06-22
Fix hash deactivate old deduplicator for uploads with invalid timezone

darinkrauss:
@jh-bate


darinkrauss:
[Fix bug with Medtronic deduplication with uploads with incorrect timezone](https://trello.com/c/BM5WMsUa/194-fix-bug-with-medtronic-deduplication-with-uploads-with-incorrect-timezone)

PR #125, opened 2017-07-01
Add deduplicator version; update deduplicator name scheme

darinkrauss:
@jh-bate Moving towards a "org.tidepool.X" naming scheme throughout. Add versioning to allow greater flexibility in data mining and implementation. Update dependencies (including one that annoyingly changed the capitalization of their import path).


darinkrauss:
Aborting to create two separate PRs.

PR #127, opened 2017-07-01
Add deduplicator version; update deduplicator name scheme

darinkrauss:
@jh-bate Moving towards a "org.tidepool.X" naming scheme throughout. Add versioning to allow greater flexibility in data mining and implementation.

### DIFF
--- a/data/deduplicator.go
+++ b/data/deduplicator.go
@@ -2,12 +2,14 @@ package data
 
 import (
 	"strconv"
+	"strings"
 
 	"github.com/tidepool-org/platform/errors"
 )
 
 type Deduplicator interface {
 	Name() string
+	Version() string
 
 	RegisterDataset() error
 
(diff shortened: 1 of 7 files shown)


data/deduplicator.go
```diff
@@ -31,14 +34,19 @@ func (d *DeduplicatorDescriptor) IsRegisteredWithAnyDeduplicator() bool {
 }
 
 func (d *DeduplicatorDescriptor) IsRegisteredWithNamedDeduplicator(name string) bool {
-	return d.Name == name
+	// TODO: Backwards compatible until after data migration to update deduplicator name scheme
+	return d.Name == name || d.Name == strings.TrimPrefix(name, "org.tidepool.")
```
jh-bate: is it worth having `org.tidepool.` as a well known const? I can see we only use it once at the moment but just thinking...

darinkrauss: When we use it more than once, I'll add a constant for it.

darinkrauss:
[[platform] Add version to deduplicator data in upload record](https://trello.com/c/rdtYGDlu/219-platform-add-version-to-deduplicator-data-in-upload-record)

PR #129, opened 2017-07-01
Add header middleware; add MD5 hash of authentication token to log

darinkrauss:
@jh-bate Needed this for debugging a problem on production recently. Adding it so we have it in the future. The authentication token (X-Tidepool-Session-Token header value) is MD5 hashed first before adding to the log to 1) prevent unnecessary exposure of session tokens; and 2) shrink the size of the value added to the log. This means that you need to run the session token through an MD5 hasher before searching the logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,18 @@
 ## HEAD
 
+- Bump `hash_deactivate_old` data deduplicator to version 1.1.0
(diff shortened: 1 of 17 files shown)


darinkrauss:
[[platform] Update Medtronic Direct deduplicator to complete fix issues with "bad" upload (incorrect timezone)](https://trello.com/c/5HJ7qkoD/220-platform-update-medtronic-direct-deduplicator-to-complete-fix-issues-with-bad-upload-incorrect-timezone)

PR #130, opened 2017-07-05
Add data archive fields; revamp Mongo queries for consistent deduplication and delete for Medtronic

darinkrauss:
@jh-bate Add `archivedTime` and `archivedDatasetId` field to all data records. Allows exact history of deduplication with multiple uploads that can be unrolled. Allows fully accurate deduplication and upload deletion. Key changes are in the mongo queries. Upload deduplication is still fast with the same two queries (with a couple extra fields on update, but that doesn't affect performance). Upload deletion is now a couple more database calls, but still reasonable performance with latest indexes. (Also, upload deletion performance isn't that important.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,18 @@
 ## HEAD
 
+- Bump `hash_deactivate_old` data deduplicator to version 1.1.0
+- Update `hash_deactivate_old` data deduplicator to use archived dataset id and time fields to accurately:
+  - Deactivate deduplicate data on dataset addition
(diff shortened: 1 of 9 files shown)


darinkrauss:
[[platform] Update Medtronic Direct deduplicator to complete fix issues with "bad" upload (incorrect timezone)](https://trello.com/c/5HJ7qkoD/220-platform-update-medtronic-direct-deduplicator-to-complete-fix-issues-with-bad-upload-incorrect-timezone)

PR #131, opened 2017-08-01
Refactor config to only use environment; remove all deprecated code

darinkrauss:
@jh-bate All configuration has been moved into the environment, out of all of those files. The biggest new change is the config package. Most of the rest of the changes were updating the specific config structs to accommodate the new system and update the tests.

Also, rather than `source .env` when first creating a terminal session for platform, it is now `source .env.local` (as opposed to `.env.test` for the tests).

I'll be updating the builder configs when we go to deploy this, too.


darinkrauss:

darinkrauss:

PR #136, opened 2017-08-03
Move app package string functions to appropriate specific packages

darinkrauss:
@jh-bate Finally, move out the few string-related functions to their own functions where directly used. (Rather than obscure functionality behind generic string functions, write functions that are specific to need. So, `SplitAddresses` rather than `SplitStringAndRemoveWhitespace`.)


darinkrauss:

darinkrauss:

PR #138, opened 2017-08-03
Add ability to set and delete to config

darinkrauss:
@jh-bate Add aforementioned `Set` and `Delete` to config. Allows for overriding configuration from command line applications (later commit/PR).

### DIFF
--- a/config/env/reporter_test.go
+++ b/config/env/reporter_test.go
@@ -96,6 +96,27 @@ var _ = Describe("Reporter", func() {
 			})
 		})
 
+		Context("Set", func() {
+			It("sets the key with the value", func() {
+				Expect(syscall.Unsetenv("TIDEPOOL_TEST_JULIETTE")).To(Succeed())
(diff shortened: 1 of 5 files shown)


config/env/reporter_test.go
```diff
@@ -96,6 +96,27 @@ var _ = Describe("Reporter", func() {
 			})
 		})
 
+		Context("Set", func() {
+			It("sets the key with the value", func() {
+				Expect(syscall.Unsetenv("TIDEPOOL_TEST_JULIETTE")).To(Succeed())
+				reporter.Set("JULIETTE", "romeo")
```
jh-bate: so i re-read this a few times and then remember it is because we are namespacing. Should that be evident in the code with a comment?

darinkrauss:

config/env/reporter_test.go
darinkrauss: Yea, I'll add documentation to the reporter itself.

darinkrauss:

PR #140, opened 2017-08-03
Refactor service package into common application; add tool; update tools

darinkrauss:
@jh-bate Refactor common functionality from `service` package into generalized `application` package. This will allow us to create new "apps" that aren't necessarily services (like background workers). Refactor common functionality from existing tools into generalized `tool` package. Update all tools appropriately (except for one, deferred as tech debt to to effort involved).

Overall, this greatly simplifies the process of creating new services, tools, and other applications.

(Much of the previous work was leading up to this.)


tool/tool.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/tidepool-org/platform/log"
+)
+
+const (
+	HelpFlag    = "help"
+	VersionFlag = "version"
+	VerboseFlag = "verbose"
+)
+
+type Tool struct {
```
jh-bate: If its always a command line tool do you think the naming should reflect that?

tool/tool.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (t *Tool) Terminate() {
+	t.args = nil
+	t.cli = nil
+
+	t.Application.Terminate()
+}
+
+func (t *Tool) CLI() *cli.App {
```
jh-bate: hmmm so maybe this covers it off `theTool.CLI()` ?

darinkrauss:

tool/tool.go
darinkrauss: This is just access to the underlying third-party cli package.

tool/tool.go
darinkrauss: It doesn't have to be. I expect that certain tools will be invoked from the admin site, once that gets implemented, via exec calls.

darinkrauss:

PR #141, opened 2017-08-04
Refactor version package to align with new application package

darinkrauss:
@jh-bate Relatively minor refactor to move specific incarnation of `VersionReporter` into `application` package (since it is directly tied to the application, after all). Also add a test file that can be included to automatically initialize the default version parameters (normally these are set via the Makefile during builds, but cannot be set during testing via `ginkgo`).


darinkrauss:
Closing due to broken Travis build for this PR.

PR #152, opened 2017-08-09
Migration for data deduplicator descriptor name and version

darinkrauss:
@jh-bate I realized I forgot to create a migration script for the code just merged into master, namely tag `v1.8.0`. This is that migration. I will update subsequent PRs to evolve this migration to the current standard.


darinkrauss:
[[platform] Add version to deduplicator data in upload record](https://trello.com/c/rdtYGDlu/219-platform-add-version-to-deduplicator-data-in-upload-record)

PR #154, opened 2017-09-07
Allow valid SemVer version strings in upload type data record

darinkrauss:
@jh-bate Fix the bug discussion in the company meeting today.


darinkrauss:
[[platform] Allow upload versions strings without 'v' prefix](https://trello.com/c/sKga5oNM/250-platform-allow-upload-versions-strings-without-v-prefix)

PR #155, opened 2017-09-12
Add deduplicator support for new devices

darinkrauss:
@jh-bate @gniezen Add deduplicator support for multiple new devices.


darinkrauss:
[[platform] Add support for Abbott Libre](https://trello.com/c/jSIMreci/254-platform-add-support-for-abbott-libre)

data/deduplicator/hash_deactivate_old_test.go
```diff
@@ -100,6 +117,23 @@ var _ = Describe("HashDeactivateOld", func() {
...
 				Expect(testFactory.CanDeduplicateDataset(testDataset)).To(BeTrue())
 			})
+
+			DescribeTable("returns true when",
+				func(deviceManufacturer string, deviceModel string) {
+					testDataset.DeviceManufacturers = app.StringArrayAsPointer([]string{deviceManufacturer})
+					testDataset.DeviceModel = app.StringAsPointer(deviceModel)
+					Expect(testFactory.CanDeduplicateDataset(testDataset)).To(BeTrue())
+				},
+				Entry("is Medtronic 523", "Medtronic", "523"),
```
jh-bate: as it is is a case sensitive match maybe we should so that with a test?

darinkrauss: I try to stay away from negative testing as it is an endless process. Should we test that it fails without spaces? You can see how that would get overwhelming. The main point here was to ensure that all of the valid device manufacturer/model combinations do, in fact, work.

PR #156, opened 2017-09-12
Allow suppressed basal to have annotations

darinkrauss:
@jh-bate @gniezen What it says on the tin.


darinkrauss:
[[platform] Allow for annotations in suspended basal](https://trello.com/c/0VmTVdY4/256-platform-allow-for-annotations-in-suspended-basal)

data/types/basal/suppressed.go
```diff
@@ -56,6 +58,8 @@ func (s *Suppressed) Validate(validator data.Validator, allowedDeliveryTypes []s
 			}
 		}
 	}
+
+	// validator.ValidateInterfaceArray("annotations", s.Annotations)    // TODO: Any validations? Optional? Size?
```
jh-bate: did you mean to leave this? why not just the todo if its a note

data/types/basal/suppressed_test.go
```diff
@@ -101,73 +102,77 @@ var _ = Describe("Target", func() {
...
+				NewTestSuppressed(nil, nil, nil, "Weekday", nil, nil),
 				[]*service.Error{}),
 			Entry("parses object that has invalid schedule name",
 				&map[string]interface{}{"scheduleName": 0},
-				NewTestSuppressed(nil, nil, nil, nil, nil),
+				NewTestSuppressed(nil, nil, nil, nil, nil, nil),
 				[]*service.Error{
 					testData.ComposeError(service.ErrorTypeNotString(0), "/scheduleName", nil),
 				}),
+			Entry("parses object that has valid annotations",
```
jh-bate: as we have no validator at the mo is there any danger to just taking anything as being a valid annotation?

darinkrauss: Yes, technically there is. I haven't figured out a reasonable way to validate these "blobs". (The only way I can think of is to manually convert them into BSON or JSON and then calculate the string length, but that seems very "hacky", so I've left it off.)

data/types/basal/suppressed.go
darinkrauss: It is a TODO so that if/when we have a way to validate we remember to add it here.

PR #157, opened 2017-09-12
Add time API; return Date header with all responses

darinkrauss:
@jh-bate @gniezen What it says on the tin. This will be used by the uploader to verify that a device is set to the correct time before uploading. (We cannot reasonably rely on the computer time as it may be incorrect.)


darinkrauss:
[[platform] Add server time API](https://trello.com/c/O06Tyd6y/255-platform-add-server-time-api)

darinkrauss:
Yes, we'll add it to the official Open API documents.

PR #158, opened 2017-09-28
Support additional Medtronic devices

darinkrauss:
@jh-bate @pazaan More 6-series devices

### DIFF
--- a/data/deduplicator/hash_deactivate_old.go
+++ b/data/deduplicator/hash_deactivate_old.go
@@ -23,7 +23,7 @@ const _HashDeactivateOldDeduplicatorName = "org.tidepool.hash-deactivate-old"
 const _HashDeactivateOldDeduplicatorVersion = "1.1.0"
 
 var _HashDeactivateOldExpectedDeviceManufacturerModels = map[string][]string{
-	"Medtronic": {"523", "723", "551", "751", "554", "754"},
+	"Medtronic": {"523", "723", "551", "751", "554", "754", "1510", "1511", "1512", "1710", "1711", "1712", "1715", "1780"},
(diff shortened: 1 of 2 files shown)


darinkrauss:
[[platform] Add support for Abbott Libre](https://trello.com/c/jSIMreci/254-platform-add-support-for-abbott-libre)

data/deduplicator/hash_deactivate_old.go
```diff
@@ -23,7 +23,7 @@ const _HashDeactivateOldDeduplicatorName = "org.tidepool.hash-deactivate-old"
 const _HashDeactivateOldDeduplicatorVersion = "1.1.0"
 
 var _HashDeactivateOldExpectedDeviceManufacturerModels = map[string][]string{
-	"Medtronic": {"523", "723", "551", "751", "554", "754"},
+	"Medtronic": {"523", "723", "551", "751", "554", "754", "1510", "1511", "1512", "1710", "1711", "1712", "1715", "1780"},
```
jh-bate: So I believe based on a comment in slack we should be dropping some as we don't have devices to test on prod

> Only support devices we have actually tested with.
> So we drop 620G (1510, 1710) from that list.

darinkrauss: Yes, that was whether this should be pushed directly to `prd` or not. That's what Eric/Lennart are testing on `stg` now so we can get tested and approved for `prd`.

darinkrauss:
[[platform] Add support for Medtronic 600-series](https://trello.com/c/z7TbslXi/261-platform-add-support-for-medtronic-600-series)

PR #159, opened 2017-09-29
Update device/prime volume to allow maximum of 10 units for cannula

darinkrauss:
@jh-bate Update from 3 to 10 units per cannula fill.


darinkrauss:
[Uploading a Medtronic Directly 723 is failing](https://trello.com/c/EzUMMMi4/15-uploading-a-medtronic-directly-723-is-failing)

PR #161, opened 2017-10-30
Remove deprecated git-hooks target; update .PHONY targets

darinkrauss:
@jh-bate Remove git hooks


darinkrauss:

PR #163, opened 2017-10-30
Fix test init and mocks

darinkrauss:
@jh-bate What is says on the tin.


darinkrauss:

PR #167, opened 2017-10-30
Resolve config scopes issue with services; remove 'services' suffix

darinkrauss:
@jh-bate


darinkrauss:

PR #168, opened 2017-10-30
Fix minor issues with use of defer

darinkrauss:
@jh-bate


service/middleware/header.go
```diff
@@ -70,6 +64,7 @@ func (h *Header) MiddlewareFunc(handler rest.HandlerFunc) rest.HandlerFunc {
 				}
 
 				if len(newFields) > 0 {
+					defer service.SetRequestLogger(request, oldLogger)
```
jh-bate: so this wont be `nil` in this case?

darinkrauss: `request` and `oldLogger` can't be `nil` due to previous code that only executes this code if they are both non-`nil`.

darinkrauss:

PR #171, opened 2017-10-30
Make logger concurrency safe

darinkrauss:
@jh-bate

### DIFF
--- a/log/json/logger.go
+++ b/log/json/logger.go
@@ -6,11 +6,13 @@ import (
 	"github.com/tidepool-org/platform/log"
 )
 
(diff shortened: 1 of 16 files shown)


log/json/logger.go
```diff
@@ -6,11 +6,13 @@ import (
 	"github.com/tidepool-org/platform/log"
 )
 
-func NewLogger(writer io.Writer, levels log.Levels, level log.Level) (log.Logger, error) {
+// CONCURRENCY: SAFE IFF writer is safe
```
jh-bate: Did you mean `SAFE IFF`  as in "If and only if" ??

darinkrauss: Yes. IFF = if and only if

darinkrauss:

PR #172, opened 2017-10-30
Dexcom API support

darinkrauss:
@jh-bate This is the large one for PR option #1. Due to the deadline I was unable to keep each commit focused to just one change. This commit wrapped up a whole pile of things in order to get the Dexcom OAuth API working as quickly as possible


auth/service/service/client.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil, errors.New("config is missing")
+	}
+	if name == "" {
+		return nil, errors.New("name is missing")
+	}
+	if logger == nil {
+		return nil, errors.New("logger is missing")
+	}
+	if authStore == nil {
+		return nil, errors.New("task store is missing")
```
jh-bate: looks like this should this be "auth store is missing" ??

.travis.yml
```diff
@@ -4,7 +4,7 @@ dist: precise
 language: go
 
 go:
-  - 1.8.3
+  - 1.9.1
```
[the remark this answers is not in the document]
darinkrauss: 1.9.2 came out less than a week ago. I'd like to wait at least a couple weeks before pushing it into production, but I'll upgrade locally.

auth/service/service/client.go
darinkrauss: Yep, good catch. I'll fix it in the latest code rather than having to patch this and update all following PRs.

darinkrauss:

PR #173, opened 2017-10-30
Update dependencies

darinkrauss:
@jh-bate Just a dependency update.


darinkrauss:

PR #175, opened 2017-10-30
Add Hi and Lo annotations

darinkrauss:
@jh-bate Enforce CGM high and low limits and add annotations.


darinkrauss:

PR #176, opened 2017-10-30
Use actual system time rather than calculated

darinkrauss:
@jh-bate Fix incorrect calculation for determining UTC time of each Dexcom API datum.


jh-bate:
LGTM

darinkrauss:

PR #178, opened 2017-11-01
Set device id, model, and serial number correctly from Dexcom API

darinkrauss:
@jh-bate Support for updating data set device id, model, and serial number. Keep accurate records of all devices that went into the data set and each datum, if possible. Add `HasUpdates` function to all `Update` model structs.


darinkrauss:

PR #179, opened 2017-11-01
Allow valid missing eventSubType for exercise and health eventTypes

darinkrauss:
@jh-bate The eventSubType is optional for *all* events.


darinkrauss:

PR #182, opened 2017-11-02
Add default source for structure validator, parser, and normalizer

darinkrauss:
@jh-bate By adding a default source for structure functionality then we can capture, by default, the most information about what structures fields fail validation, parsing, or normalization.  (Not as "perfect" as specifying the source at the time of processing, but this is much more pragmatic. Almost all will use the default and only structures parsed out of the request query will use a specific non-default source.)


darinkrauss:

PR #183, opened 2017-11-02
Run Dexcom fetch task every 45 to 75 minutes (rather than always 1 hour)

darinkrauss:
@jh-bate Quick, but important, one.


darinkrauss:

PR #185, opened 2017-11-14
Fix alert settings snooze validation for Dexcom API

darinkrauss:
@jh-bate


darinkrauss:
[<XS> Ensure all appropriate data is visible, even after failure with Dexcom API](https://trello.com/c/RcAU7sVR/279-xs-ensure-all-appropriate-data-is-visible-even-after-failure-with-dexcom-api)

PR #186, opened 2017-11-21
Add fix to handle Dexcom API not correctly reporting daily G5 Mobile devices

darinkrauss:
@jh-bate


darinkrauss:
[<S> Update Dexcom API fetch algorithm to account for missing G5 Mobile from /devices endpoint](https://trello.com/c/SPGUS2xM/277-s-update-dexcom-api-fetch-algorithm-to-account-for-missing-g5-mobile-from-devices-endpoint)

PR #188, opened 2017-11-30
Delta uploads

darinkrauss:
@jh-bate


test/mongo/mongo.go
```diff
@@ -46,7 +47,7 @@ var _ = ginkgo.SynchronizedAfterSuite(func() {
 })
 
 func Address() string {
-	return "127.0.0.1"
+	return os.Getenv("TIDEPOOL_STORE_ADDRESSES")
```
jh-bate: so for the tests does this get set in the travis file?

darinkrauss: That environment variable is set in `env.test.sh` which is source'd in the Makefile which is invoked by Travis. So, we're all good here.

darinkrauss:
Accidentally closed.

darinkrauss:

PR #189, opened 2017-11-30
Tool to compute metric hash from user id

darinkrauss:
@jh-bate


metric/metric.go
```diff
@@ -0,0 +1,20 @@
+package metric
+
+import (
+	"crypto/sha1"
+	"encoding/hex"
+
+	"github.com/tidepool-org/platform/errors"
+)
+
+func HashFromUserID(userID string, salt string) (string, error) {
```
jh-bate: is there any danger in not having a test for this?

darinkrauss: Other than incorrectly calculating a hash, no. Yes, it should have a test,  but I'm going back through the code now and adding a whole bunch. I'll make sure this one gets done in a later commit.

darinkrauss:

PR #190, opened 2017-11-30
Use Go v1.9.2; minor updates

darinkrauss:
@jh-bate


darinkrauss:

PR #192, opened 2017-12-29
Add custom User-Agent header to client requests

darinkrauss:
@jh-bate Add User-Agent header to all client requests where possible. Allows for easier request tracking.


darinkrauss:

PR #193, opened 2017-12-29
Properly handle expired access tokens from Dexcom API

darinkrauss:
@jh-bate Fixes issue with excessive "Login expired - try signing out & in again" messages.


darinkrauss:

PR #194, opened 2017-12-29
Remove legacy code to fixup unexpected data from Dexcom API

darinkrauss:
@jh-bate We'll only use this after getting positive confirmation from Dexcom that this legacy code is no longer necessary.


darinkrauss:

darinkrauss:
Received confirmation from Dexcom that legacy code is not longer necessary.

PR #197, opened 2018-02-07
Avoid time period when Dexcom API refresh token failures occur

darinkrauss:
@jh-bate Ignore 2:45am to 3:45am time period where intermittent refresh token failure occur due to Dexcom backups. Known problem for Dexcom.


darinkrauss:

PR #198, opened 2018-02-13
Fix Travis issues with Docker build

darinkrauss:
@jh-bate


jh-bate:
I assume `trusty` is todo with versioning??

darinkrauss:
`trusty` is the version of Ubuntu being used for the builds. The mongo source should match the OS version.

PR #200, opened 2018-02-13
Add LifeScan Verio and Verio Flex to supported devices

darinkrauss:
@jh-bate What it says on the tin.


darinkrauss:
[<XS> Add Verio & Verio Flex to list of supported devices](https://trello.com/c/fD8JYbuZ/316-xs-add-verio-verio-flex-to-list-of-supported-devices)

PR #202, opened 2018-02-14
Return error from config.Get rather than bool for compatibility

darinkrauss:
@jh-bate Need to switch to returning error if config value not found, rather than a simple bool for more normal Golang functionality.


darinkrauss:

PR #211, opened 2018-02-16
Refactor data normalizer to use common structure normalizer; expose ReportError

darinkrauss:
@jh-bate Looks like a big change, but really it is just using the new structure.Normalizer as a "base class" for the data.Normalizer. This is part of the effort to deprecate the data.Parser, data.Validator, and data.Normalizer.


darinkrauss:

PR #212, opened 2018-02-16
Remove structure.Validating; does not work as expected

darinkrauss:
@jh-bate This was a neat idea, but doesn't work due to the concrete-type to interface{} casting issues with Go (per our discussion a while back regarding the updated errors). Basically, a nil concrete-type cast to an interface{} is not nil (because the interface{} is a pair, the type and the value; so, while the value is nil, the type is known (the concrete-type) so you can't compare the interface{} against nil and have it work as you expect. Thus, remove all of this code that makes this assumption because it will lead to unintentional and unexpected errors/panics.


darinkrauss:

PR #213, opened 2018-02-16
Remove data.ValidateInterface*; does not work as expected

darinkrauss:
@jh-bate ValidateInterface and ValidateInterfaceArray isn't generally useful (yes, every object is an interface{}, duh), so removing it.


darinkrauss:

PR #222, opened 2018-03-06
Add build-watch and ci-build-watch targets to automatic build after file change

darinkrauss:
@jh-bate This works pretty much the same as `make test-watch` where the tests are automatically run if it detects any file changes. These two targets just do `build` or `ci-build` if any files change. Works well when writing code to quickly get rid of syntax errors and the like.


darinkrauss:

PR #224, opened 2018-03-06
Add automated basal data type

darinkrauss:
@jh-bate Add new `basal/automated` type. Very much like `basal/scheduled`, but cannot be suppressed by a `basal/temp` (only by a `basal/suspend`).


darinkrauss:

PR #226, opened 2018-03-10
Refactor data type functions to constants

darinkrauss:
@jh-bate Basically just replacing `Type()`, `SubType()`, and `DeliveryType()` functions, with `Type, `SubType`, and `DeliveryType` constants for simplification. Not sure why I didn't do this in the first place.


darinkrauss:

PR #227, opened 2018-03-10
Refactor data factory into specific packages

darinkrauss:
@jh-bate The old data factory used to be entirely away from the `data/types` package hierarchy which meant that you couldn't use parts within the `data/types` package hierarchy (due to circular dependencies).  

Broke this up into a factory for each type, sub type, and delivery type *into a `factory` package under the relevant `data/types` package*. This allows you to use factories from other data types. For example, the `calculator` type can now use the `basal` type factory.

Also, the factories are simpler to understand. None of the map of func map funcs garbage, which, while perhaps clever, was not really understandable. (Understandable > clever).

(Note: Still avoiding "parsing" tests until I can get the new structure parser installed so I don't have to rewrite a whole bunch of code.)


darinkrauss:

PR #228, opened 2018-03-10
Refactor blood value and precision normalization

darinkrauss:
@jh-bate Separate out units normalization from value precision normalization and be consistent about it. (There will be another commit coming later that will remove the TODOs and enable the precision calculation for raw `mmol/L` values.)


darinkrauss:
Code no longer needed.

PR #241, opened 2018-03-27
Add support for new Trividia Health devices

darinkrauss:
@gniezen Added the new devices per our discussion. I also reordered the device list alphabetically to make it easier for me in the future.


darinkrauss:
[<XS> Add True Metrix, True Metrix Go and True Metrix Air to list of supported devices](https://trello.com/c/ww6GR5za/349-xs-add-true-metrix-true-metrix-go-and-true-metrix-air-to-list-of-supported-devices)

PR #258, opened 2018-04-18
Fix Dexcom API unexpected data; temporarily modify incoming data to expected values

darinkrauss:
@pazaan Fixes all of the (currently known) unexpected data from the Dexcom API. In some cases it is due to unclear/incomplete documentation. In other cases it directly contradicts the documentation. In any case we have to (temporarily, in some cases) fix it out our end so we end up with valid data and a working Dexcom connection. The turnaround time for getting it fixed at the source is lengthy.

I expect that we will see some additional unexpected data failures since the authentication failure (jn previous PR) was taking precedence over these types of failures.

### DIFF
--- a/dexcom/device.go
+++ b/dexcom/device.go
@@ -179,6 +179,11 @@ func (a *AlertSetting) Validate(validator structure.Validator, existingAlertName
 }
 
 func (a *AlertSetting) validateFixedLow(validator structure.Validator) {
+	// HACK: Dexcom - snooze of 28 is invalid; use snooze of 30 instead (per Dexcom)
+	if a.Snooze == 28 {
+		a.Snooze = 30
+	}
+
 	validator.String("unit", &a.Unit).OneOf(UnitMgdL) // TODO: Add UnitMmolL
 	switch a.Unit {
 	case UnitMgdL:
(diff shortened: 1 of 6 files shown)


darinkrauss:
[<XS> Tidepool integration with Dexcom API fails due unexpected data](https://trello.com/c/psPeMY5n/364-xs-tidepool-integration-with-dexcom-api-fails-due-unexpected-data)

dexcom/device.go
```diff
@@ -179,6 +179,11 @@ func (a *AlertSetting) Validate(validator structure.Validator, existingAlertName
 }
 
 func (a *AlertSetting) validateFixedLow(validator structure.Validator) {
+	// HACK: Dexcom - snooze of 28 is invalid; use snooze of 30 instead (per Dexcom)
```
pazaan: Will this always be 28? Or could it be < 30?

dexcom/event.go
```diff
@@ -95,6 +95,12 @@ func (e *Event) validateCarbs(validator structure.Validator) {
 }
 
 func (e *Event) validateExercise(validator structure.Validator) {
+	// HACK: Dexcom - value of -1 is invalid; ignore unit and value instead (per Dexcom)
+	if e.Value != nil && *e.Value == -1.0 {
```
pazaan: What does `*e` do?

dexcom/device.go
darinkrauss: From what I've seen and what Dexcom confirmed it is only 28. This only recently started happening after a recent deploy from Dexcom.

dexcom/event.go
darinkrauss: `e` is an "object" that has a field `Value` that is a pointer to a `float64`. So, the `e.Value != nil` ensures that the pointer is not nil and `*e.Value == -1.0` ensures that the `float64` the pointer points to is equal to `-1.0`.

So, in short, the `*` is dereferencing the `e.Value`.

pazaan: Ah, makes sense.

darinkrauss:
[<XS> Tidepool integration with Dexcom API fails due unexpected data](https://trello.com/c/psPeMY5n/364-xs-tidepool-integration-with-dexcom-api-fails-due-unexpected-data)

PR #260, opened 2018-04-18
Move mutator to request package; ignore missing mutators in client

darinkrauss:
@pazaan Simple move the mutator structs from client package to request package (as it really belongs there).


darinkrauss:

PR #263, opened 2018-04-25
Return on validation or normalization error after sending response

darinkrauss:
@jh-bate Immediately exit once an error response has been sent.


darinkrauss:
[Do not return invalid JSON when an error occurs upon data set creation](https://trello.com/c/rNWNdR3x/379-do-not-return-invalid-json-when-an-error-occurs-upon-data-set-creation)

PR #264, opened 2018-04-25
Return only active data sets

darinkrauss:
@jh-bate


darinkrauss:
[Use correct BSD-2 licenses in platform GitHub repository](https://trello.com/c/8REuf5Cu/382-use-correct-bsd-2-licenses-in-platform-github-repository)

PR #266, opened 2018-04-30
Fix bug in upload data type with check against incorrect type

darinkrauss:
@jh-bate Accidentally specified wrong type for "upload" records.


darinkrauss:

darinkrauss:
Yes, I'll add it to the rest of the upcoming parsing tests.

PR #267, opened 2018-04-30
Use "Units" rather than "units" for insulin

darinkrauss:
@jh-bate Apparently, the agreed upon units of insulin is "Units", not "units". So, conforming to accepted standard, per request from Howard.


darinkrauss:

PR #270, opened 2018-04-30
Improve data model time zone support

darinkrauss:
* Add tzdata to development Docker images
* Validate data time zone against known time zones
* Support time zone across all data types
* Break timezone usage into two words

@jh-bate


darinkrauss:

data/types/upload/upload.go
```diff
@@ -22,9 +22,9 @@ const (
 	DeviceTagInsulinPump                 = "insulin-pump"
 	StateClosed                          = "closed"
 	StateOpen                            = "open"
-	TimeProcessingAcrossTheBoardTimezone = "across-the-board-timezone"
+	TimeProcessingAcrossTheBoardTimeZone = "across-the-board-timezone" // TODO: Rename to across-the-board-time-zone or alternative
```
jh-bate: repeated const ??

data/types/base.go
```diff
@@ -30,8 +31,8 @@ const (
 	TagLengthMaximum        = 100
 	TagsLengthMaximum       = 100
 	TimeFormat              = time.RFC3339
-	TimezoneOffsetMaximum   = 7 * 24 * 60  // TODO: Fix! Limit to reasonable values
-	TimezoneOffsetMinimum   = -7 * 24 * 60 // TODO: Fix! Limit to reasonable values
+	TimeZoneOffsetMaximum   = 7 * 24 * 60  // TODO: Fix! Limit to reasonable values
```
jh-bate: Is it worth setting to a reasonable value now? Or are we still trying to decide what that should be ....

data/types/base.go
```diff
@@ -62,7 +63,8 @@ type Base struct {
 	Source            *string                                      `json:"source,omitempty" bson:"source,omitempty"`
 	Tags              *[]string                                    `json:"tags,omitempty" bson:"tags,omitempty"`
 	Time              *string                                      `json:"time,omitempty" bson:"time,omitempty"`
-	TimezoneOffset    *int                                         `json:"timezoneOffset,omitempty" bson:"timezoneOffset,omitempty"`
+	TimeZoneName      *string                                      `json:"timezone,omitempty" bson:"timezone,omitempty"`             // TODO: Rename to timeZoneName
```
jh-bate: cool - was going to ask that question. Why not `timeZoneName` rather than `timezone` ...

data/types/upload/upload.go
darinkrauss: Yes, this is intentional. Slowly moving Upload to DataSet. So it is defined in both places.

data/types/base.go
darinkrauss: The reasonable values are -12*60 - 14*60. We are working on getting it fixed in the Uploader (work-in-progress), the database, the backend, and the data donation program. It has to be a step-by-step process. Uploader is first, then I'll update the backend and migrate the data. So, this will be fixed some time in the relatively near future.

data/types/base.go
darinkrauss: Eventually will migrate everything to proper two word usage, but for now it is piecemeal.

PR #272, opened 2018-04-30
Add expected duration to status device event data model

darinkrauss:
@jh-bate Add `expectedDuration` to `deviceEvent/status` data model. Missed the first go around.

### DIFF
--- a/data/types/device/status/status.go
+++ b/data/types/device/status/status.go
@@ -26,9 +26,10 @@ func Names() []string {
 type Status struct {
 	device.Device `bson:",inline"`
 
-	Duration *int       `json:"duration,omitempty" bson:"duration,omitempty"`
-	Name     *string    `json:"status,omitempty" bson:"status,omitempty"`
-	Reason   *data.Blob `json:"reason,omitempty" bson:"reason,omitempty"`
+	Duration         *int       `json:"duration,omitempty" bson:"duration,omitempty"`
+	DurationExpected *int       `json:"expectedDuration,omitempty" bson:"expectedDuration,omitempty"`
+	Name             *string    `json:"status,omitempty" bson:"status,omitempty"`
+	Reason           *data.Blob `json:"reason,omitempty" bson:"reason,omitempty"`
 }
 
 func NewStatusDatum(parser data.ObjectParser) data.Datum {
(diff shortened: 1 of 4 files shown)


darinkrauss:
[<L> Add missing datatypes and validation to platform](https://trello.com/c/F9zje73u/183-l-add-missing-datatypes-and-validation-to-platform)

data/types/device/status/status.go
```diff
@@ -95,6 +97,12 @@ func (s *Status) Validate(validator structure.Validator) {
 	}
 
 	validator.Int("duration", s.Duration).GreaterThanOrEqualTo(DurationMinimum) // TODO: .Exists() - Suspend events on Animas do not have duration?
+	expectedDurationValidator := validator.Int("expectedDuration", s.DurationExpected)
```
jh-bate: naming flips ... `expectedDuration`  => `DurationExpected`

data/types/device/status/status.go
```diff
@@ -26,9 +26,10 @@ func Names() []string {
 type Status struct {
 	device.Device `bson:",inline"`
 
-	Duration *int       `json:"duration,omitempty" bson:"duration,omitempty"`
-	Name     *string    `json:"status,omitempty" bson:"status,omitempty"`
-	Reason   *data.Blob `json:"reason,omitempty" bson:"reason,omitempty"`
+	Duration         *int       `json:"duration,omitempty" bson:"duration,omitempty"`
+	DurationExpected *int       `json:"expectedDuration,omitempty" bson:"expectedDuration,omitempty"`
```
jh-bate: why not just `ExpectedDuration` as you keep on flipping the naming in other places

data/types/device/status/status.go
```diff
@@ -77,6 +78,7 @@ func (s *Status) Parse(parser data.ObjectParser) error {
 	}
 
 	s.Duration = parser.ParseInteger("duration")
+	s.DurationExpected = parser.ParseInteger("expectedDuration")
```
jh-bate: flipped ...

data/types/device/status/status.go
darinkrauss: Check out `combination.go`. This is consistent with the rest of the code. The JSON field name is `expected*` while the Go type name is the reversed `*Expected`. This is on purpose because for V2 I'll be renaming all of the JSON fields to follow the Go type name format.

For constants, types, variable names, etc. I try to use the name and then the qualifier. So:

Duration
DurationExpected
DurationMaximum
DurationMinimum

Much easier to visually parse then finding all MaximumDuration among all of the other Maximum* names.

data/types/device/status/status.go
darinkrauss: See above.

data/types/device/status/status.go
darinkrauss: See above.

PR #275, opened 2018-04-30
Add ingredients to food data model for greater flexibility

darinkrauss:
@jh-bate Add ingredients (consider it a sub-array of foods so that a user can describe a whole meal in components rather than just as a sum total whole). Add `code` field for UPC code (or a variety of other coding systems).

### DIFF
--- a/data/types/food/food_test.go
+++ b/data/types/food/food_test.go
@@ -22,12 +22,14 @@ func NewMeta() interface{} {
 	}
 }
 
-func NewFood() *food.Food {
+func NewFood(ingredientArrayDepth int) *food.Food {
 	datum := food.New()
 	datum.Base = *testDataTypes.NewBase()
 	datum.Type = "food"
 	datum.Amount = NewAmount()
 	datum.Brand = pointer.String(test.NewText(1, 100))
+	datum.Code = pointer.String(test.NewText(1, 100))
+	datum.Ingredients = NewIngredientArray(ingredientArrayDepth)
 	datum.Meal = pointer.String(test.RandomStringFromArray(food.Meals()))
 	if datum.Meal != nil && *datum.Meal == food.MealOther {
 		datum.MealOther = pointer.String(test.NewText(1, 100))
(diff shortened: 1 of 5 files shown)


darinkrauss:

data/types/food/food_test.go
```diff
@@ -22,12 +22,14 @@ func NewMeta() interface{} {
 	}
 }
 
-func NewFood() *food.Food {
+func NewFood(ingredientArrayDepth int) *food.Food {
```
jh-bate: depth or length??

data/types/food/ingredient.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	if i.Ingredients != nil {
+		i.Ingredients.Normalize(normalizer.WithReference("ingredients"))
+	}
+	if i.Nutrition != nil {
+		i.Nutrition.Normalize(normalizer.WithReference("nutrition"))
+	}
+}
+
+type IngredientArray []*Ingredient
```
jh-bate: maybe I have missed this pattern in the past but it really stick out this time. Why not just `Ingredients` ??

data/types/food/ingredient_test.go
```diff
@@ -0,0 +1,357 @@
...
+	It("IngredientCodeLengthMaximum is expected", func() {
+		Expect(food.IngredientCodeLengthMaximum).To(Equal(100))
+	})
+
+	It("IngredientNameLengthMaximum is expected", func() {
+		Expect(food.IngredientNameLengthMaximum).To(Equal(100))
+	})
+
+	Context("ParseIngredient", func() {
+		// TODO
```
jh-bate: why later??

data/types/food/ingredient.go
jh-bate: yeah OK have now seen it all over the show ...

data/types/food/food_test.go
darinkrauss: Depth. This test function creates a hierarchy of ingredients. This parameter limits the total overall depth (or otherwise it recursively creates more and more and eventually blows the stack).

data/types/food/ingredient.go
darinkrauss: I believe `IngredientArray` is a more accurate type name. I am trying to be consistent with `Array` or `Map` type suffix. (The problem is that simple adding an `s` suffix doesn't really tell *how* there are multiple. Is it via a map, an array, something else?

I originally tried it with `s` suffix, but ran into naming collisions because of not being specific about type.

data/types/food/ingredient_test.go
darinkrauss: All of the parsing tests will be updated when I revamp the parsing to use the `structure` parser (as opposed to the `data`-specific parser. Will add tests with that work.

PR #283, opened 2018-05-20
Update dependencies

darinkrauss:
@jh-bate There is only one code change required because of this dependency update (in `id/id.go`). (So much for Go packages being backwards compatible.) I reviewed all of the dependency changes and all tests pass.


darinkrauss:

PR #285, opened 2018-06-04
Add and update random test data generators

darinkrauss:
@jh-bate Only updates to tests and test support.

### DIFF
--- a/test/random.go
+++ b/test/random.go
@@ -3,51 +3,234 @@ package test
 import (
 	"math"
 	"math/rand"
+	"time"
 )
 
 func RandomBool() bool {
 	return rand.Intn(2) == 0
 }
 
+func RandomDuration() time.Duration {
+	return RandomDurationFromRange(RandomDurationMinimum(), RandomDurationMaximum())
+}
+
+func RandomDurationFromArray(array []time.Duration) time.Duration {
+	if len(array) == 0 {
+		panic("RandomDurationFromArray: array is empty")
+	}
+	return array[rand.Intn(len(array))]
+}
+
+func RandomDurationFromRange(minimum time.Duration, maximum time.Duration) time.Duration {
+	if maximum < minimum {
+		panic("RandomDurationFromRange: maximum is not greater than or equal to minimum")
+	}
+	if minimum < RandomDurationMinimum() {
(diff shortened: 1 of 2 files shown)


test/random.go
```diff
@@ -3,51 +3,234 @@ package test
...
+		minimum = RandomDurationMinimum()
+	}
+	if maximum > RandomDurationMaximum() {
+		maximum = RandomDurationMaximum()
+	}
+	return minimum + time.Duration(rand.Int63n(int64(maximum-minimum+1)))
+}
+
+func RandomDurationMaximum() time.Duration {
+	return 10 * 365 * 24 * time.Hour
```
jh-bate: they are all constants??

test/random.go
```diff
@@ -3,51 +3,234 @@ package test
...
+	}
 	return minimum + rand.Intn(maximum-minimum+1)
 }
 
+func RandomIntMaximum() int {
+	return math.MaxInt32
+}
+
+func RandomIntMinimum() int {
+	return math.MinInt32
```
jh-bate: which is a constant??

test/random.go
```diff
@@ -3,51 +3,234 @@ package test
...
+	}
+	return string(runes)
+}
+
+func RandomStringLengthMaximum() int {
+	return 64
+}
+
+func RandomStringLengthMinimum() int {
+	return 1
```
jh-bate: Naming??

test/random.go
darinkrauss: Is is the maximum random duration. So the maximum duration that the random duration generators should generate. I name it this way for consistency and scoping reasons. That is everything starts with the prefix + "Maximum", "Minimum", etc...

test/random.go
darinkrauss: See above...

test/random.go
darinkrauss: See above...

test/random.go
jh-bate: hmmm maybe in context :) I guess I keep thinking it needs a comment ... and then if it needs a comment it need renaming. anyway not blocking

darinkrauss:

PR #289, opened 2018-06-04
Refactor request Mutator into RequestMutator and ResponseMutator

darinkrauss:
@jh-bate Simple change to allow mutators to mutate both requests and responses.


darinkrauss:

PR #296, opened 2018-06-06
Update dependency management

darinkrauss:
* Use dep dependency management tool rather than Godep
* Remove unnecessary Makefile targets related to Godep

@jh-bate Forced to switch from Godep to dep to accomodate the upcoming AWS packages.


darinkrauss:

PR #297, opened 2018-06-06
Add unstructured storage

darinkrauss:
* Add unstructured storage backed by file system and AWS S3 with factory
* Add AWS API interface with actual and test implementations
* Update test config reporter to support scopes
* Add temporary file and directory test helpers
* Add test io.Reader

@jh-bate


darinkrauss:

PR #299, opened 2018-06-15
Add common service API status router

darinkrauss:
@jh-bate This is common to all services so break it out here.


darinkrauss:

PR #305, opened 2018-06-15
Refactor application runner startup

darinkrauss:
* Refactor application package to allow dependency injection
* Simplify application execution to automatically inject default dependencies
* Move all application/service/tool initialization from New to Initialize
* Move application version to application package
* Default all applications to use UTC local time

@jh-bate


darinkrauss:

PR #306, opened 2018-06-15
When deserializing errors an empty error array returns nil

darinkrauss:
@jh-bate


darinkrauss:

darinkrauss:
Yea, probably so. I'll add one when I get to more tests for the errors package as a whole.

PR #307, opened 2018-06-15
Update logger to delete an existing field if new field value is nil

darinkrauss:
@jh-bate


darinkrauss:

PR #309, opened 2018-06-15
Update user client with authorization checks

darinkrauss:
* Add EnsureAuthorizedService and EnsureAuthorizedUser to user.UserClient
* Update user client test helper to use latest mocking technique
* Remove unnecessary parameter from test RespondWith

@jh-bate


darinkrauss:

PR #313, opened 2018-06-15
Enable blob service for Docker and minor updates

darinkrauss:
* Add Dockerfile and config for running blob service
* Remove check for correct content type from legacy service responses
* Fix incorrect usage of test package in trace middleware
* Minor updates to blob service configuration

@jh-bate


darinkrauss:

PR #319, opened 2018-07-11
Use pointers for data set fields to match upload fields

darinkrauss:
@jh-bate Just use pointers for the fields of the data set structure. Not sure why I originally had these as non-pointers, but I decided (a while back) that use of pointers for almost all fields in any JSON serialized structs made the most sense.


darinkrauss:

darinkrauss:

PR #320, opened 2018-07-11
Enable new continuous Tidepool Mobile data set

darinkrauss:
@jh-bate Tidepool Mobile is switching over to use a single, continuous data set for all data (rather than hundreds to thousands over a long period of time). Also, Tidepool Mobile will be adding a number of new data types. Thus, some of the previously required fields are now optional. Also, allow data set query to filter on client name (so that Tidepool Mobile can find its one data set).


darinkrauss:

darinkrauss:

PR #324, opened 2018-08-04
Replace user.NewID with userTest.RandomID in tests

darinkrauss:
@jh-bate Standardizing on using Random* test functions to generate sample data.


darinkrauss:

PR #327, opened 2018-08-04
Cleanup blob service and tests

darinkrauss:
@jh-bate A bunch of small changes to finalize the blob service which I'll be using as a template of sorts.


darinkrauss:

PR #328, opened 2018-08-04
Rename various to more generic form in blob service

darinkrauss:
@jh-bate Simple variable name renames in order to make it more amenable as a template service.


darinkrauss:

PR #329, opened 2018-08-04
Replace data.NewID with dataTest.RandomID in tests

darinkrauss:
@jh-bate More test helper updates.

### DIFF
--- a/data/test/data_set.go
+++ b/data/test/data_set.go
@@ -1,0 +1,14 @@
+package test
+
+import (
+	"github.com/tidepool-org/platform/data"
+	"github.com/tidepool-org/platform/test"
+)
+
+func RandomSetID() string {
+	return data.NewSetID()
+}
+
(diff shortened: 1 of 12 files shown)


darinkrauss:

data/test/data_set.go
```diff
@@ -0,0 +1,14 @@
+package test
+
+import (
+	"github.com/tidepool-org/platform/data"
+	"github.com/tidepool-org/platform/test"
+)
+
+func RandomSetID() string {
```
jh-bate: is it worth `RandomDataSetID` ?? as the namespace is `test.RandomDataSetID` ... i was a little confused initially by the `Random` and `Set` ID ;)

darinkrauss: Ah, well, the namespace is really dataTest.RandomSetID, but I see your point. I'll be moving some of this stuff out anyways, so I'll leave it be for now.

PR #331, opened 2018-08-04
Refactor data source to separate package with complete tests

darinkrauss:
Pull out all data source functionality into separate package hierarchy. Will eventually be breaking out into a separate service.


darinkrauss:

PR #337, opened 2018-08-04
Add request condition revision to enforce specific update and delete operations

darinkrauss:
@jh-bate The request condition (specifically the revision) can be used to prevent two separate process from updating the same resource at the same time. See next PRs for details.


darinkrauss:

PR #338, opened 2018-08-04
Add data source revision

darinkrauss:
@jh-bate Enable the capability to only update/delete a resource if it is at an expected revision. This can prevent two services from modifying the same resource at the same time resulting in potential data loss.


darinkrauss:

PR #340, opened 2018-08-04
Add blob size maximum

darinkrauss:
@jh-bate Should have had this in earlier. Make sure blobs cannot be over 100MB in size.


darinkrauss:

PR #341, opened 2018-08-04
Fix data store tests

darinkrauss:
@jh-bate Fix previously commented out tests for the data store. (They were commented out a while back due to some general store changes. This just gets them working again.)


darinkrauss:

PR #342, opened 2018-08-04
Add continuous origin deduplicator

darinkrauss:
* Add continuous origin deduplicator to use origin id as deduplication key for continuous data sets
* Allow client to specific data set deduplicator

@jh-bate Will be used by v2 Dexcom API and updates to Tidepool Mobile. Deduplicates within the same continuous data set, but uses the new origin.id field as the deduplication key.


darinkrauss:

PR #343, opened 2018-08-15
Fix data service data source client

darinkrauss:
@jh-bate Fixed one bug found in manual testing. No easy way to add a unit test.


darinkrauss:

PR #344, opened 2018-08-15
Update origin structure

darinkrauss:
* Add origin type "service"
* Add origin payload
* Update origin name to validate as string with maximum length

@jh-bate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## HEAD
 
+* Add origin type "service"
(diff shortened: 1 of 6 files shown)


darinkrauss:
[Update Tidepool data model to include new fields for new Tidepool Mobile HealthKit integration](https://trello.com/c/vWLRN7Yf/455-update-tidepool-data-model-to-include-new-fields-for-new-tidepool-mobile-healthkit-integration)

PR #345, opened 2018-08-15
Add API to batch delete data set data

darinkrauss:
@jh-bate


darinkrauss:
[Add API to batch delete data as specified by id and origin.id fields](https://trello.com/c/qj6ankOP/456-add-api-to-batch-delete-data-as-specified-by-id-and-originid-fields)

PR #347, opened 2018-08-17
Add time zone change data

darinkrauss:
* Refactor device time change event to include time zone name
* Add time zone name validator

@jh-bate


darinkrauss:
[Add new time zone change event to Tidepool data model](https://trello.com/c/176apPhX/459-add-new-time-zone-change-event-to-tidepool-data-model)

PR #351, opened 2018-08-30
Update data source provider session id

darinkrauss:
* Add optional provider session id to data source update
* Add new provider session id to data source after reconnecting Dexcom API

@jh-bate


darinkrauss:

PR #356, opened 2018-09-05
Fix minor issues when opening data set deduplicator

darinkrauss:
* Return updated data set after creation
* Remove unnecessary condition when opening deduplicator for data set

@jh-bate


darinkrauss:

PR #359, opened 2018-09-24
Update mongo queries for data using selectors to more effectively use indexes

darinkrauss:
@jh-bate Update mgo query to use `$in` operator and minimum query for effective index use.


darinkrauss:

PR #361, opened 2018-09-29
Rename data.Delete to data.Selector for more general usage

darinkrauss:
@jh-bate Just a rename and no other changes.


darinkrauss:

PR #372, opened 2018-09-29
Deprecate current cgm settings alert structures

darinkrauss:
@jh-bate In preparation for new CGM setting additions.


darinkrauss:
[Integrate Dexcom v2 API](https://trello.com/c/VuGi9jWN/392-integrate-dexcom-v2-api)

PR #377, opened 2018-10-17
Add user id to ArchiveDeviceDataUsingHashesFromDataSet selector to properly use indexes

darinkrauss:
@pazaan The change you requested to the code to use the new indexes. The tests did not require changing because they already tested presence of the user id (since user id is *always* associated with the upload id in all records).


darinkrauss:
[Update data deduplication mongo query to properly utilize existing indexes](https://trello.com/c/Kclq2fvd/486-update-data-deduplication-mongo-query-to-properly-utilize-existing-indexes)

PR #382, opened 2018-10-20
Update tool dependencies

darinkrauss:
* Update tool dependencies
* Add build tools as additional dependencies
* Update Makefile to install build tools from vendor directory
* Update Makefile to remove unused tools
* Fix a sort order difference between Mac and Linux with the lint process

@pazaan Fixes the root cause of the build failures. Vendor the tool dependencies and update `Makefile` to install and use those vendored tool dependencies.


darinkrauss:
[Fix build break due to non-vendored lint tools](https://trello.com/c/4VTY1mj2/493-fix-build-break-due-to-non-vendored-lint-tools)

PR #384, opened 2018-11-06
Fix data set id validation to allow Jellyfish id format

darinkrauss:
@pazaan Copy-paste error from quite a while ago. However, since this is just a Jellyfish/Platform difference, this only presents when we internally try to delete a data set via Platform that was created in Jellyfish.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## HEAD
 
+* Fix data set id validation to allow Jellyfish id format
 * Update tool dependencies
 * Add build tools as additional dependencies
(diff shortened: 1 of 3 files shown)


darinkrauss:
[Fix issue with deleting data sets sourced from Jellyfish with unexpected ids](https://trello.com/c/05Y0JvSb/497-fix-issue-with-deleting-data-sets-sourced-from-jellyfish-with-unexpected-ids)

darinkrauss:
Replaced by PR https://github.com/tidepool-org/platform/pull/387.

PR #387, opened 2018-12-05
Fix data set id validation to allow Jellyfish id format

darinkrauss:
@pazaan Same as already approved PR https://github.com/tidepool-org/platform/pull/384. Moving to a `develop` branch so created a new PR to keep it clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## HEAD
 
+* Fix data set id validation to allow Jellyfish id format
+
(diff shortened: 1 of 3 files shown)


darinkrauss:
[Fix issue with deleting data sets sourced from Jellyfish with unexpected ids](https://trello.com/c/05Y0JvSb/497-fix-issue-with-deleting-data-sets-sourced-from-jellyfish-with-unexpected-ids)

PR #391, opened 2018-12-06
Relax Dexcom API validations to match Tidepool data model validations

darinkrauss:
* Relax Dexcom API validations to match Tidepool data model validations
* Rename insulin dose constants to properly include units

@pazaan

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## HEAD
 
+## v1.31.1
+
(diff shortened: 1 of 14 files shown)


darinkrauss:
[Relax Dexcom API validations to match Tidepool data model validations](https://trello.com/c/3F05AbND/507-relax-dexcom-api-validations-to-match-tidepool-data-model-validations)

PR #392, opened 2018-12-06
Fix blob size verification and handling

darinkrauss:
* Return HTTP status code 413 Request Entity Too Large when blob size exceeds maximum
* Replace blob.Create with blob.Content
* Remove blob Content-Length header usage and blob.Content.Size as it does not work with content encoding
* Update test config reporter for better debugging


darinkrauss:
[Fix blob size verification and handling](https://trello.com/c/sJFkHbub/506-fix-blob-size-verification-and-handling)

PR #394, opened 2018-12-12
Update dump users tool to be more generic and usable

darinkrauss:
@pazaan One last branch that was hanging out before I can start with the image service PRs. 

This tool is used as a base for building other specific tools used to dump users based upon very specific criteria. In particular, this base is used to build the tools we use to pull users using very specific study selection criteria.

### DIFF
--- a/tools/dump_users/dump_users.go
+++ b/tools/dump_users/dump_users.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


tools/dump_users/dump_users.go
```diff
@@ -590,6 +576,52 @@ func (t *Tool) getUserDataDevice(query interface{}, logger log.Logger) (*Device,
 	}, nil
 }
 
+func (t *Tool) getUserDataActiveTypesStats(userID string, user *User, logger log.Logger) error {
+	logger.Debug("Getting user data active types stats")
+
+	pipeline := []bson.M{
+		{
+			"$match": bson.M{
+				"_active": true,
```
pazaan: TL;DR: put `_userId` before `_active` in this query.
We don't have an index on both `_active` and `_userId`. The result set will be much smaller if we have a query that makes use of one of the indexes that has `_userId` first, rather than `_active`.

darinkrauss: The `bson.M` type is part of the `mgo.v2/bson` package and defined as:

```
type M map[string]interface{}
```

A Golang `map` is internally implemented as a hash map and so are ordered by the hash of the key, not by insertion order. So, it doesn't really make a difference what order the keys are listed in here.

Try running this Go playground I created to see: https://play.golang.org/p/lBJSHzcXBoO

This is why the `mgo` structure for creating and index is an array of keys. That is:

```
type Index struct {
    Key        []string   // Index key fields; prefix name with dash (-) for descending order
    Unique     bool     // Prevent two documents from having the same index key
    ...
```

pazaan: Makes sense. Can we use https://godoc.org/labix.org/v2/mgo#Query.Hint?

darinkrauss: Sure, but how do we know this query won't already use this index?

darinkrauss: Actually, now that I think about it, I know the query already uses this index because I've already used two derivatives of this tool to pull users for two different studies and the queries were responsive.

pazaan: That seems accidental rather than intentional. I'll pass this for now, but seems like something we should discuss.

darinkrauss: Ah, I see the confusion. I thought you meant that it wouldn't use an index at all. I believe it uses this index:

```
	{
		"v" : 1,
		"key" : {
			"_userId" : 1,
			"_schemaVersion" : 1,
			"time" : -1
		},
		"name" : "_userId_1__schemaVersion_1_time_-1",
		"ns" : "data.deviceData",
		"partialFilterExpression" : {
			"_active" : true
		}
	},
```

which is an index on `_active` first given the partial filter expression and then `_userId` second.

This seems like exactly what we want to be using, don't you?

pazaan: Ah, that's true. Don't mind me... 🤦‍♂️

PR #399, opened 2019-01-17
Refactor before image service

darinkrauss:
Best to do this one commit-by-commit as each commit is pretty focused and straightforward. Only  https://github.com/tidepool-org/platform/commit/25bfa5aaf9702b1d6a0eebe08798285c33868183 and https://github.com/tidepool-org/platform/commit/cd68479ba18a0b67c3be9a4feb0b0a3c469b21e1 have significant changes.


darinkrauss:
[Various refactoring in preparation for new image service](https://trello.com/c/FTiwOHYn/534-various-refactoring-in-preparation-for-new-image-service)

darinkrauss:
Regarding commit https://github.com/tidepool-org/platform/pull/399/commits/a4022ff007e4b71da0e761d5437a8e376d0fc05f, there aren't any standard conventions for an alternative to `bytes` AFAIK. The problem with `data` is that we have a top-level package `data` that would cause conflict, as well. I agree that `bites` is not the best, but it shouldn't conflict with any package names! :)

PR #402, opened 2019-01-22
Corrected README for building go tools

derrickburns:
The README referred to a make target that does not exist.

@darinkrauss I am not certain that this is correct, but I think it is.  Please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,17 @@ The Tidepool Platform API.
 
 # Setup
 
-1. Install Go version 1.10 or later.
+1. Install Go version 1.11.4 or later.
 1. Install `dep` dependency management tool. See https://golang.github.io/dep/docs/installation.html.
+1. Install mongodb (if it is not already installed).
+1. Start mongodb (if it is not already running).
 1. Create a brand new Go directory.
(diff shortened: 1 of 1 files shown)


pazaan:
@derrickburns - you're right. Looks like it was removed in https://github.com/tidepool-org/platform/blame/53f758772091408c45d50d4019fcb8a4fa7d1305/Makefile#L91

README.md
```diff
@@ -23,6 +23,8 @@ For example:
 ```
 brew install go
 brew install dep
+brew install mongo
```
pazaan: I'm not sure if we want/need to prescribe how to run mongo. For example, I run mongo using a Docker container, with port 27018 poked through to localhost.

darinkrauss: I agree. I think the instructions above about installing and starting Mongo are good enough. We might want to specify the version, though. Production is currently 3.2 while I'm running 3.6.x locally.

darinkrauss: The same probably goes for the version of Go. How it is installed is irrelevant, I suppose.

OTOH, this is just an example, so maybe we should just leave it.

PR #403, opened 2019-01-23
Convert from go dep to using go modules.

derrickburns:
Go 1.11 supports a new standard, [go modules](https://github.com/golang/go/wiki/Modules#how-to-install-and-activate-module-support), for managing go dependencies that replaces the `dep` dependency manager.  Now that we have upgraded to Go 1.11, this commit simply switches to using go modules. 

While this commit seems large (resulting in the deletion of over a thousand files from the source tree), it is rather easy to reproduce, since most of it was automatic.
 
Specifically, I ran 
```
go mod init
go mod tidy
```
which resulted in the creation of
```
go.mod
go.sum
```
Then, I removed 
```
Gopkg.lock
Gopkg.toml
```
and the vendor files that `go dep` created.  See [this post ](https://blog.callr.tech/migrating-from-dep-to-go-1.11-modules/) for more info on the process.

The remaining vendor files are the build tools.

I confirmed that the `Makefile` still runs properly by running `make build` and `make test`.  All tests pass.


derrickburns:
@darinkrauss Is there any reason to vendor all the dependencies into our repo?  For example, did you want all the code in the vendor directory to avoid a failure scenario in the build where the build machine cannot get the files?

tjotala:
The vendor directory is motivated by desire to control those dependencies. See discussion under the golang section in https://docs.google.com/document/d/1-_q33-YSyvqXIwrSFNBbpRwJqb7C2pX3DfDFj_ue9oU/edit

derrickburns:
Yes, I understand the motivation.

My question is, is it necessary to copy the files into our GitHub repo OR is it sufficient to specify which specific immutable version of each dependency we require?  For example, the Google Bazel build tool
ensure reproducibility using the latter approach.

Derrick

darinkrauss:
My one question about this is - what problem or issue are you trying to solve? Are you hitting specific problems with how it is done now or are there any pain points I'm not aware of? Go modules is still listed as experimental in version 1.11, so I don't think we should update it until it is the officially supported Golang "standard" (which it isn't *yet*).

derrickburns:
I am closing the PR.  The consensus of the team is that we will should not embrace Go language/tool chain features that are marked experimental. Go modules is marked experimental in 1.11.  If and when Go modules is part of the standard tool chain, we may reconsider such changes. But even then, there are reasons to more slowly adopt changes to the Go tool chain.

PR #405, opened 2019-01-31
Fix delete user

darinkrauss:
Probably easiest to look at it commit-by-commit. 

Also, the final commit refactors the user service to use the non-deprecated service objects. The only service remaining using the deprecate service objects is the data service.


pazaan:
@derrickburns Just wanted you to be aware of the above comment so you know about it 😄

derrickburns:
@pazaan I appreciate the cc.

PR #407, opened 2019-03-21
Do not log status pings.

derrickburns:


service/middleware/access_log.go
```diff
@@ -38,11 +39,24 @@ func NewAccessLog() (*AccessLog, error) {
 	return &AccessLog{}, nil
 }
 
+func (a *AccessLog) ignore(req *rest.Request) bool {
+	// ignore liveness and readiness probes
+	if strings.Compare(req.URL.RequestURI(), "/status") == 0 {
```
pazaan:
```suggestion
	if req.URL.RequestURI() == "/status") {
```

From the [go documentation](https://golang.org/pkg/strings/#Compare):
> Compare is included only for symmetry with package bytes. It is usually clearer and always faster to use the built-in string comparison operators ==, <, >, and so on

pazaan:
@derrickburns Also backmerge from `develop` :)

service/middleware/access_log.go
```diff
@@ -3,6 +3,7 @@ package middleware
 import (
 	"fmt"
 	"net"
+	"strings"
```
pazaan: `strings` is an unneeded import now.

derrickburns:
[Quiet noisy log messages from platform services.](https://trello.com/c/BH1npNwU/639-quiet-noisy-log-messages-from-platform-services)

derrickburns:
[Suppress noisy platform log messages.](https://trello.com/c/YbawaBoZ/643-suppress-noisy-platform-log-messages)

PR #417, opened 2019-06-17
Combination boluses with 0 normal okay if extended > 0

pazaan:
A regression was introduced where a 0 normal component was allowed
for combination boluses only if expectedNormal existed.
However, it's also valid for there to be a 0 normal component if
extended is non-zero
- Fixes https://trello.com/c/s67Yc2qP


pazaan:
[Backend Regression (?) - Dual Wave Boluses do not store correctly on the backend.](https://trello.com/c/s67Yc2qP/719-backend-regression-dual-wave-boluses-do-not-store-correctly-on-the-backend)

pazaan:
[Backend Regression - Dual Wave Boluses do not store correctly on the backend](https://tidepool.atlassian.net/browse/BACK-26)

PR #424, opened 2019-08-27
[BACK-810] Add absorption duration field to nutrition/carbohydrate object

jamesraby:
Initial PR (for me) to add a simple field to the data platform.

Adds absorption duration field to nutrition object


data/types/food/carbohydrate.go
```diff
@@ -6,6 +6,8 @@ import (
 )
 
 const (
+	AbsorptionDurationTimeMaximum        = 1000.0
```
pazaan: [nit] prefer `AbsorptionDurationMaximum` (since duration is time)

data/types/food/carbohydrate.go
```diff
@@ -6,6 +6,8 @@ import (
...
+	AbsorptionDurationTimeMinimum        = 0.0
```
pazaan: [nit] prefer `AbsorptionDurationMinimum`

data/types/food/carbohydrate.go
```diff
@@ -6,6 +6,8 @@ import (
 )
 
 const (
+	AbsorptionDurationMaximum            = 1000.0
```
pazaan: API Spec puts max at 86400. Sorry for not noticing earlier.

data/types/food/carbohydrate.go
```diff
@@ -24,11 +26,12 @@ func CarbohydrateUnits() []string {
 }
 
 type Carbohydrate struct {
-	DietaryFiber *float64 `json:"dietaryFiber,omitempty" bson:"dietaryFiber,omitempty"`
-	Net          *float64 `json:"net,omitempty" bson:"net,omitempty"`
-	Sugars       *float64 `json:"sugars,omitempty" bson:"sugars,omitempty"`
-	Total        *float64 `json:"total,omitempty" bson:"total,omitempty"`
-	Units        *string  `json:"units,omitempty" bson:"units,omitempty"`
+	AbsorptionDuration *float64 `json:"absorptionDuration,omitempty" bson:"absorptionDuration,omitempty"`
```
pazaan: `absorptionDuration` should be in `Nutrition`, not as far as `Carbohydrate`. And preferably should be an int32, instead of a float64.

pazaan:
@jamesraby The Travis build is failing the `lint` check. I recommend running `make pre-commit` before committing :)

jamesraby:
Unfortunately - I can not run make due to golang setup.  Lennart has a PR to make this a bit easier that is coming in shortly - I thought I was going to be able to get this through before his PR - but I do not see any error messages coming out of the build - so I am kind of guessing what the problem is

pazaan:
* Fixes [BACK-810](https://tidepool.atlassian.net/browse/BACK-810)

darinkrauss:
FYI - No other data model property name specifies the units within the name itself (eg. `AbsorptionDurationSeconds`). So, adding it here creates an inconsistency with rest of the data model. If you want to be explicit about the units then I suggest you consider what many of the newer data model additions use which is to have a "struct" with a units and a "value" property.  For example:

```
absorptionDuration: {
    value: 10000,
    units: "seconds"
}
```

(The "value" field name can be whatever makes sense for this struct.)

Check out the `food` type and its contained types for a variety of examples of this pattern.

Ultimately it is up to the owner of the data model (currently @pazaan ?) to make this decision, but I did want to point out the implications of this change.

My :two_cents:

jamesraby:
@darinkrauss - thanks - I agree.  I did see names with the units in them - but those were all constants - such as `CarbohydrateNetGramsMaximum`.  Unless any objections - will change it back.

darinkrauss:
Right, the constants need the units in the name because the constant value is particular to a unit for that value property. So, you'll see constants like:

```
	DistanceFeetPerMile            = 5280.0
	DistanceKilometersPerMile      = 1.609344
	DistanceMetersPerMile          = 1609.344
	DistanceUnitsFeet              = "feet"
	DistanceUnitsKilometers        = "kilometers"
	DistanceUnitsMeters            = "meters"
	DistanceUnitsMiles             = "miles"
	DistanceUnitsYards             = "yards"
	DistanceValueFeetMaximum       = DistanceValueMilesMaximum * DistanceFeetPerMile
	DistanceValueFeetMinimum       = DistanceValueMilesMinimum * DistanceFeetPerMile
	DistanceValueKilometersMaximum = DistanceValueMilesMaximum * DistanceKilometersPerMile
	DistanceValueKilometersMinimum = DistanceValueMilesMinimum * DistanceKilometersPerMile
	DistanceValueMetersMaximum     = DistanceValueMilesMaximum * DistanceMetersPerMile
	DistanceValueMetersMinimum     = DistanceValueMilesMinimum * DistanceMetersPerMile
	DistanceValueMilesMaximum      = 100.0
	DistanceValueMilesMinimum      = 0.0
	DistanceValueYardsMaximum      = DistanceValueMilesMaximum * DistanceYardsPerMile
	DistanceValueYardsMinimum      = DistanceValueMilesMinimum * DistanceYardsPerMile
	DistanceYardsPerMile           = 1760.0
```

because distance can have units of feet, yards, miles, meters, and kilometers. The distance struct itself is just:

```
type Distance struct {
	Units *string  `json:"units,omitempty" bson:"units,omitempty"`
	Value *float64 `json:"value,omitempty" bson:"value,omitempty"`
}
```

PR #440, opened 2019-11-14
Suppress excess logging, add error output

derrickburns:
@tjotala This is an example of a change we would want to push out quickly.  We need a process that enables that.  Right now, I am under the impression that I have to create a Jira ticket just to get this approved.

### DIFF
--- a/service/api/api.go
+++ b/service/api/api.go
@@ -55,10 +55,10 @@ func (a *API) InitializeMiddleware() error {
 	if err != nil {
 		return err
 	}
-	accessLogMiddleware, err := middleware.NewAccessLog()
-	if err != nil {
-		return err
-	}
+	//accessLogMiddleware, err := middleware.NewAccessLog()
+	//if err != nil {
+		//return err
+	//}
 	recoverMiddleware, err := middleware.NewRecover()
 	if err != nil {
 		return err
@@ -77,7 +77,7 @@ func (a *API) InitializeMiddleware() error {
 		loggerMiddleware,
 		errorMiddleware,
 		traceMiddleware,
-		accessLogMiddleware,
+		//accessLogMiddleware,
 		statusMiddleware,
 		timerMiddleware,
 		recorderMiddleware,
--- a/service/api/api.go
+++ b/service/api/api.go
@@ -57,7 +57,7 @@ func (a *API) InitializeMiddleware() error {
 	}
 	//accessLogMiddleware, err := middleware.NewAccessLog()
 	//if err != nil {
-		//return err
+	//return err
 	//}
 	recoverMiddleware, err := middleware.NewRecover()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


derrickburns:
@HaroldBernard Perhaps we can create a "fasttrack" jira ticket using your tidebot? 

`/jira` -> would create a ticket with the same title as the PR, assign it to the person making the comment, move it to 'In progress' state with a back reference to the GH PR, and send a Jira message to @tapani requesting approval?

tjotala:
> @tjotala This is an example of a change we would want to push out quickly. We need a process that enables that. Right now, I am under the impression that I have to create a Jira ticket just to get this approved.

Our SOP for development (https://docs.google.com/document/d/19vWtyDNesVyxE5gKXwppgRGWvn-EXrttCLJH5nRys1U/edit#) has allowances for _trivial_ changes with respect to reviews. It also notes that GitHub is part of our "Design History File" and for a trivial change such as this I think PR approval in GitHub will suffice.

tjotala:
> Any reason that it is not building?
> 
> Changes look ok - although not completely sure what it means to take out the accessLogMiddleware

@derrickburns Looks to me like you're removing the access log from middleware stack & I'm assuming that is redundant with the API gateway in place doing the same. Can you confirm? Assuming we would not deploy any services in the legacy environment (sans the API gateway), that seems OK.

derrickburns:
Correct

service/api/api.go
```diff
@@ -55,10 +55,10 @@ func (a *API) InitializeMiddleware() error {
 	if err != nil {
 		return err
 	}
-	accessLogMiddleware, err := middleware.NewAccessLog()
-	if err != nil {
-		return err
-	}
+	//accessLogMiddleware, err := middleware.NewAccessLog()
```
darinkrauss: I would completely remove (not just comment out). The code is in the git history so can be recovered if ever necessary.

task/service/service/service.go
```diff
@@ -200,9 +200,9 @@ func (s *Service) initializeDexcomClient() error {
 
 	dxcmPrvdr, err := dexcomProvider.New(s.ConfigReporter().WithScopes("provider"), s.dataSourceClient, s.TaskClient())
 	if err != nil {
-		s.Logger().Warn("Unable to create dexcom provider")
+		s.Logger().WithError(err).Warn("Unable to create dexcom provider")
 	} else {
-		s.Logger().Debug("Loading dexcom client config")
+		s.Logger().WithError(err).Debug("Loading dexcom client config")
```
darinkrauss: No need for `.WithError(err)` here since, by definition, `err` is `nil` according to the logic here.

darinkrauss: BTW, no harm to include this here, but it just won't ever do anything.

PR #443, opened 2019-11-20
[BACK-1015] Add new DeviceTransmitterEvent

jamesraby:
Add the new DeviceTransmitterEvent to data model.

Change name of DeviceSensor to Sensor


data/types/device/sensor/devicesensor.go
```diff
@@ -1,4 +1,4 @@
-package devicesensor
+package sensor
```
pazaan: Suggest that this file be renamed to `sensor.go` for consistency.

data/types/device/sensor/devicesensor_suite_test.go
```diff
@@ -1,4 +1,4 @@
-package devicesensor_test
+package sensor_test
```
pazaan: Suggest that this file be renamed to `sensor_suite_test.go` for consistency.

data/types/device/sensor/devicesensor_test.go
```diff
@@ -1,4 +1,4 @@
-package devicesensor_test
+package sensor_test
```
pazaan: Suggest that this file be renamed to `sensor_test.go` for consistency.

data/types/device/sensor/devicesensor.go
jamesraby: agreed

data/types/device/sensor/devicesensor_test.go
jamesraby: done

data/types/device/sensor/devicesensor_suite_test.go
jamesraby: done

PR #448, opened 2020-01-15
Add `update-password` command to `tapi`

pazaan:
@tjotala Adding `update-password` functionality to the `tapi` tool (replacing the `tidepool-pwreset.js` tool).
The corresponding `develop` PR is [available here](https://github.com/tidepool-org/platform/pull/447)

### DIFF
--- a/tools/tapi/cmd/user.go
+++ b/tools/tapi/cmd/user.go
@@ -106,10 +106,6 @@ func UserCommands() cli.Commands {
 							Name:  UserIDFlag,
 							Usage: "`USERID` of the user to update",
 						},
-						cli.StringFlag{
-							Name:  PasswordFlag,
-							Usage: "`PASSWORD` of the user to update (required if authenticated as the user being deleted)",
-						},
 					),
 					Before: ensureNoArgs,
 					Action: userUpdatePassword,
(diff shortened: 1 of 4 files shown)


tools/tapi/cmd/user.go
```diff
@@ -98,6 +98,22 @@ func UserCommands() cli.Commands {
...
+					Name:  "update-password",
+					Usage: "update a user's password",
+					Flags: CommandFlags(
+						cli.StringFlag{
+							Name:  UserIDFlag,
+							Usage: "`USERID` of the user to update",
+						},
+						cli.StringFlag{
+							Name:  PasswordFlag,
+							Usage: "`PASSWORD` of the user to update (required if authenticated as the user being deleted)",
```
tjotala: Misleading description? This user is not being deleted, right?

tools/tapi/cmd/user.go
```diff
@@ -195,3 +211,27 @@ func userDelete(c *cli.Context) error {
 
 	return reportMessage(c, "User deleted.")
 }
+
+func userUpdatePassword(c *cli.Context) error {
+	userID := c.String(UserIDFlag)
+	password := c.String(PasswordFlag)
```
tjotala: I'm confused. This prompts for the user's current password, but does not appear to use it for anything?

pazaan: Good catch! 🤦‍♂

pazaan:
@tjotala Good questions! Updated based on your comments.

PR #453, opened 2020-01-28
Pump Status and Dosing Decision objects

jamesraby:
Add pump status and dosing decision objects for new data model additions


jamesraby:
Going to new branching strategy.  Subsumed by #462

PR #456, opened 2020-02-03
#BACK-1162 add Carb Unit field to calculator object

jamesraby:
Add carb unit field so UI can calculate carbs correctly

### DIFF
--- a/data/types/calculator/calculator_test.go
+++ b/data/types/calculator/calculator_test.go
@@ -44,6 +44,7 @@ func NewCalculator(units *string) *calculator.Calculator {
 	datum.InsulinSensitivity = pointer.FromFloat64(test.RandomFloat64FromRange(dataBloodGlucose.ValueRangeForUnits(units)))
 	datum.Recommended = NewRecommended()
 	datum.Units = units
+	datum.CarbUnits = pointer.FromString(test.RandomStringFromArray(calculator.CarbUnits()))
 	return datum
 }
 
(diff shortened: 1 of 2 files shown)


data/types/calculator/calculator_test.go
```diff
@@ -1622,6 +1625,13 @@ var _ = Describe("Calculator", func() {
 					pointer.FromString("mg/dl"),
 					func(datum *calculator.Calculator, units *string) { datum.BolusID = nil },
 				),
+				Entry("Invalid Carb Units",
+					pointer.FromString("mg/dl"),
```
pazaan:
```suggestion
```
Line not required

data/types/calculator/calculator_test.go
```diff
@@ -1622,6 +1625,13 @@ var _ = Describe("Calculator", func() {
...
+					func(datum *calculator.Calculator, units *string) {
+						datum.CarbUnits = pointer.FromString("Invalid")
+					},
+					errorsTest.WithPointerSourceAndMeta(structureValidator.ErrorValueStringNotOneOf("Invalid", calculator.CarbUnits()), "/carbUnits", NewMeta()),
+				),
```
pazaan: Also add tests for the 2 valid values.

data/types/calculator/calculator_test.go
jamesraby: This is confusing - but I believe what I have is correct.  This is just creating a unit parameter.  This is actually not used in this test case - but necessary.  We are setting the `CarbUnit` parameter to invalid in the func - so the test should work fine.

data/types/calculator/calculator_test.go
jamesraby: In terms of valid values - it does create valid values (albeit random) in the NewCalculator routine - and that is tested in the first unit test.  We could create one for each valid value - but I do not think that is generally done through the framework.

PR #457, opened 2020-02-10
BACK-1161 Mitigate Dexcom API bug

pazaan:
Adds a "Forgiving Time" parser to mitigate the recently introduced
bug in the Dexcom API
* Fixes [BACK-1161]

### DIFF
--- a/structure/parser.go
+++ b/structure/parser.go
@@ -26,6 +26,7 @@ type ObjectParser interface {
 	String(reference string) *string
 	StringArray(reference string) *[]string
 	Time(reference string, layout string) *time.Time
+	ForgivingTime(reference string, layout string) *time.Time
 
 	Object(reference string) *map[string]interface{}
 	Array(reference string) *[]interface{}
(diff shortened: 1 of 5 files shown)


structure/parser.go
```diff
@@ -81,3 +82,20 @@ type ArrayParser interface {
...
 	WithReferenceErrorReporter(reference int) ErrorReporter
 }
+
+// ForgivingTimeString is a helper function added specifically to handle https://tidepool.atlassian.net/browse/BACK-1161
+// It should be deprecated once Dexcom fixes their API.
+func ForgivingTimeString(stringValue string) (forgivingTime string) {
+	if len(stringValue) < 19 {
+		forgivingBytes := []byte("0000-01-01T00:00:00")
+		for i := range forgivingBytes {
+			if i < len(stringValue) {
```
toddkazakov: This condition will always be true

pazaan: Not always - For example, if `stringValue` is `2019-02-01T13:00`, then `len(stringValue) < 19` but `len(forgivingBytes) > len(stringValue)`.

toddkazakov: Right. I thought you were iterating over `stringValue` which would have made that condition unnecessary.

PR #467, opened 2020-03-03
[WIP] BACK-953 Add prescription data model

toddkazakov:


prescription/prescription.go
```diff
@@ -0,0 +1,127 @@
...
+	PrescriptionTherapySettingCPT                  = "cpt" // TODO: Certified Personal Trainer?
+
+	PrescriptionLoopModeClosedLoop  = "closedLoop"
+	PrescriptionLoopModeSuspendOnly = "suspendOnly"
+)
+
+
+type Prescription struct {
+	ID                string           `json:"id" bson:"id"`
+	AccessCode        string           `json:"accessCode,omitempty" bson:"accessCode,omitempty"`
```
tjotala: plaintext or hashed?

jamesraby:
Just took a quick glance.

I would base off of master vs develop with new flow.

I will study the prescription docs today and give more feedback

prescription/prescription.go
```diff
@@ -0,0 +1,127 @@
...
+	CGMType                    *CGMType                           `json:"cgmType" bson:"cgmType"`
+	// TODO: Add Suspend threshold - Dependent on latest data model changes
+	// TODO: Add Insulin model - Dependent on latest data model changes
+	// TODO: Add Bolus schedule? Does not exist in current pump settings model.
+}
+
+type PumpType struct {
+	Manufacturers *[]string `json:"manufacturers,omitempty" bson:"manufacturers,omitempty"`
+	Model         *string   `json:"model,omitempty" bson:"model,omitempty"`
+	SerialNumber  *string   `json:"serialNumber,omitempty" bson:"serialNumber,omitempty"`
```
tjotala: We may not know the serial number ahead of time.

prescription/prescription.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+type CGMType struct {
+	Manufacturers *[]string `json:"manufacturers,omitempty" bson:"manufacturers,omitempty"`
+	Model         *string   `json:"model,omitempty" bson:"model,omitempty"`
+	SerialNumber  *string   `json:"serialNumber,omitempty" bson:"serialNumber,omitempty"`
```
tjotala: We may not know the serial number ahead of time.

prescription/prescription.go
```diff
@@ -0,0 +1,127 @@
...
+	Sex               *string          `json:"sex,omitempty" bson:"sex,omitempty"`
+	Weight            *Weight          `json:"weight,omitempty" bson:"weight,omitempty"`
+	YearOfDiagnosis   *string          `json:"yearOfDiagnosis,omitempty" bson:"yearOfDiagnosis,omitempty"`
+	PhoneNumber       *string          `json:"phoneNumber,omitempty" bson:"phoneNumber,omitempty"`
+	Address           *Address         `json:"address,omitempty" bson:"address,omitempty"`
+	InitialSettings   *InitialSettings `json:"initialSettings,omitempty" bson:"initialSettings,omitempty"`
+	Training          *string          `json:"training,omitempty" bson:"training,omitempty"`
+	TherapySettings   *string          `json:"therapySettings,omitempty" bson:"therapySettings,omitempty"`
+	LoopMode          *string          `json:"loopMode,omitempty" bson:"loopMode,omitempty"`
+	AcknowledgedTerms *bool            `json:"acknowledgedTerms,omitempty" bson:"acknowledgedTerms,omitempty"`
```
tjotala: Redundant with user account / profile? `termsAccepted`

prescription/prescription.go
```diff
@@ -0,0 +1,127 @@
...
+	PrescriptionStateExpired   = "expired"
+	PrescriptionStateActive    = "active"
+	PrescriptionStateInactive  = "inactive"
+
+	PrescriptionTrainingInPerson = "inPerson"
+	PrescriptionTrainingInModule = "inModule"
+
+	PrescriptionTherapySettingInitial              = "initial"
+	PrescriptionTherapySettingTransferPumpSettings = "transferPumpSettings"
+	PrescriptionTherapySettingCPT                  = "cpt" // TODO: Certified Personal Trainer?
```
tjotala: https://en.wikipedia.org/wiki/Current_Procedural_Terminology

prescription/prescription.go
```diff
@@ -0,0 +1,127 @@
...
+	Value *float64 `json:"value,omitempty" bson:"value,omitempty"`
+	Units *string  `json:"units,omitempty" bson:"units,omitempty"`
+}
+
+type Address struct {
+	Line1      *string `json:"line1,omitempty" bson:"line1,omitempty"`
+	Line2      *string `json:"line2,omitempty" bson:"line2,omitempty"`
+	City       *string `json:"city,omitempty" bson:"city,omitempty"`
+	State      *string `json:"state,omitempty" bson:"state,omitempty"`
+	PostalCode *string `json:"postalCode,omitempty" bson:"postalCode,omitempty"`
```
tjotala: We may want to include `Country`, even though for now it'll always be US.

prescription/prescription.go
```diff
@@ -0,0 +1,127 @@
...
+	PumpType                   *PumpType                          `json:"pumpType" bson:"pumpType"`
+	CGMType                    *CGMType                           `json:"cgmType" bson:"cgmType"`
+	// TODO: Add Suspend threshold - Dependent on latest data model changes
+	// TODO: Add Insulin model - Dependent on latest data model changes
+	// TODO: Add Bolus schedule? Does not exist in current pump settings model.
+}
+
+type PumpType struct {
+	Manufacturers *[]string `json:"manufacturers,omitempty" bson:"manufacturers,omitempty"`
+	Model         *string   `json:"model,omitempty" bson:"model,omitempty"`
```
tjotala: We may need a way to share the "master" list of valid/supported manufacturers & models with the client apps, to avoid replicating the list everywhere. I expect that the Web UX for creating prescriptions will have to provide some way to select the device(s) from a list, rather than a free-form entry. Ultimately, the Loop app can only accept the device(s) that it supports.

prescription/prescription.go
```diff
@@ -0,0 +1,127 @@
...
+	PrescriptionTherapySettingInitial              = "initial"
+	PrescriptionTherapySettingTransferPumpSettings = "transferPumpSettings"
+	PrescriptionTherapySettingCPT                  = "cpt" // TODO: Certified Personal Trainer?
+
+	PrescriptionLoopModeClosedLoop  = "closedLoop"
+	PrescriptionLoopModeSuspendOnly = "suspendOnly"
+)
+
+
+type Prescription struct {
```
tjotala: I think we need to support digitally signing the prescriptions - at a minimum with the prescriber's key, but ideally with the patient's key as well. Neither have keys at present time.

prescription/prescription.go
toddkazakov: This is intended to be used by the prescriber and it was listed as a requirement. I'll rename it to `prescriberTermsAccepted`.

prescription/prescription.go
jamesraby: Does the client go to the backend to get information?  My understanding is that the client (Loop) just sends info to the backend.  I always assumed that the client has the option of not using a backend

jamesraby: Can the prescriber change the prescription after the initial one?  Do we keep both "initialSettings" objects around?  

I do think that the settings object is fine for recording the initial prescription.

toddkazakov: The user will have the ability to retrieve a prescription (with the initial settings) from the Loop app, thus fetching info from the backend. They can be changed but we will keep the full revision history.

prescription/prescription.go
toddkazakov: Turned out to be Certified *Pump* Trainer

toddkazakov:
Will reopen this PR against `master`

PR #469, opened 2020-03-10
Refinement of new Tidepool data model types and changes

darinkrauss:
@jamesraby For your review. I think these changes straighten out what we need here for Loop and provide better consistency/conformance with the rest of the Tidepool data model. Since there were some things moved around and several things renamed, it looks like more changes than actually occurred.

Also, I didn't think it would be fair of me to just do the bare minimum of changing the JSON API without also updating the related Go structs and all of the tests. So, the API, Go code, and all unit tests have been updated accordingly and everything should build and pass all unit tests.

Feel free to merge this change into your branch after your review and then do an official PR afterwards. Of course, if you have any questions, comments, see problems or issues, or want to discuss anything please let me know.

Some details about what I updated and why:

• General
  • Replace all \*Date named fields with \*Time (for consistency; Loop uses terms like startDate, but the Tidepool data model uses terms like startTime)
  • Use field names common to Tidepool data model such as Amount and Rate (for consistency; rather than ones like Quantity and UnitsPerHour)
  • Limit new required fields to absolute minimum (for flexibility; even though Loop may provide more than minimum)
  • Use fully-qualified package import names to avoid confusion (on my part when reading/updating code)
  • Remove unnecessary Normalize functions on non-Base derived structs (optional if secondary level struct)
  • Pull out Random\* and Clone\* test functions into separate files in package test directory
  • Removed unnecessary "if *** != nil" checks before validating simple types (the validator will handle nil appropriately for simple types)
  • Use common Go practice of using first letter of class as "self" variable (eg. "func (s *SuspendThreshold) ..."; note "s" follows first letter of "SuspendThreshold")
  • Renamed backend internal _version field to VersionInternal to allow Version field to be publicly (in ApplicationSettings)
• ApplicationSettings
  • Rename appSettings to applicationSettings (to avoid abbreviations)
  • Rename loopAppVersion to just version to be more generally useful (not specific to Loop)
• DosingDecision
  • Rename carbsOnBoard to carbohydratesOnBoard
  • Move Forecast into DosingDecision package
  • Remove Forecast.Type (having it on every single element in a forecast is redundant and would bloat the JSON; using a forecast array per type so all values within forecast are all of the same type)
  • Move RecommendedBolus into its own struct for consistency
  • Add new Units struct that includes units for insulin (and avoids abbreviations)
  • Remove unused DeviceTypes (not sure where this came from)
  • Move Alerts and Forecast (as BloodGlucoseForecast) from PumpStatus (required here and per Google doc)
  • Add InsulinOnBoard and RecommendedBolus structs
• PumpSettings
  • Make InsulinModel optional (remove ".Exists()"; backward compatibility)
  • Rename bgPreMealTarget to bgTargetPreprandial for accuracy
  • Re-sort validations to keep in matching order to declaration (easy for me to grok and make sure I haven't made a mistake)
• PumpStatus
  • Rename BasalDeliveryState to just BasalDelivery (to avoid redundancy; it already contains a State field within it)
  • Rename BolusState to just BolusDelivery (to avoid redundancy; it already contains a State field within it))
  • Define Bolus and Basal specific Dose structs (the common struct had a lot of information that wasn't relevant to each one; I tracked down the Loop code to see exactly what was used for bolus and basal doses and it shrunk each struct considerably)
  • Add Battery.Time and Battery.Units (Loop currently reports percentage, but some devices report volts; allows expansion as necessary)
  • Rename Device.DeviceID to just Device.ID (to avoid redundancy)
  • Add additional version related fields to Device
  • Create separate Reservoir struct to mirror Battery struct
  • Move Alerts and Forecast to DosingDecision (required there and per Google doc)
  • Rename pumpstatus.go to pump_status.go (for consistency)


jamesraby:
Wow!

PR #480, opened 2020-04-02
[SIR-51] Temporarily remap invalid 'g6 pro' transmitter generation to 'g6'

toddkazakov:
`g6 pro` is currently not recognized transmitter generation, which causes data fetches to be failing for multiple users. Dexcom advised us to temporarily remap `g6 pro` to `g6` on our side until they fix the issue.


toddkazakov:
Closing this as it was fixed by Dexcom.

PR #481, opened 2020-04-04
[BACK-1297] [BACK-1311] Make all `EnsureIndex` calls in `platform`

pazaan:
There were several services that had `EnsureIndex()` functions that were not being called.

* Fixes [BACK-1297]

[BACK-1297]: https://tidepool.atlassian.net/browse/BACK-1297


pazaan:
@toddkazakov - yes, `hydrophone` does confirmations, but we should still fix this here. Good catch!

PR #490, opened 2020-04-24
[BACK-1279] Remove `_schemaVersion` from data model

pazaan:


pazaan:
@toddkazakov For re-review. As mentioned in the commit, the reference to `_schemaVersion` remains in `selectFields`, because the database will still have `_schemaVersion` data in it that needs to be projected until BACK-1276 is complete

PR #492, opened 2020-05-22
Append mongo connection option params correctly

toddkazakov:
The first two connection params `ssl` and `replicaSet` were not set correctly due to the missing ampersand.


toddkazakov:
TLS is enforced at the server level. @pazaan is currently working on replacing `mgo` with the official mongo driver, where the issue is already fixed.

PR #501, opened 2020-06-11
[BACK-1419] Use birthday in addition to access codes for claiming prescriptions 

toddkazakov:
Had to reopen this PR to have a saner diff


toddkazakov:
@jamesraby this needs re-review - reopened the PR to get rid of the outdated commit history

PR #505, opened 2020-07-09
[BACK-1424] Validate device specific increments in prescription settings

toddkazakov:
Different devices can have device specific increments and limits that are not taken into account in our current data model. Those are hard coded in Tidepool Loop, but we must make sure all submitted prescriptions are valid for the prescribed devices.

This PR also updates the travis configuration so we use go 1.14 for executing the unit tests in travis, as well as the go version which is used for building the prescriptions service. The prescription service has an indirect dependency on a module which requires 1.14. All other service still use go1.11.4 - this is defined per service in each [Dockerfile](https://github.com/tidepool-org/platform/blob/master/Dockerfile.data#L2).


devices/client.go
```diff
@@ -0,0 +1,57 @@
...
+	"go.uber.org/fx"
+	"google.golang.org/grpc"
+
+	"github.com/tidepool-org/platform/errors"
+)
+
+var ClientModule = fx.Provide(NewClient)
+
+type Config struct {
+	DevicesClientAddress string `envconfig:"TIDEPOOL_DEVICES_CLIENT_ADDRESS" required:"true"`
```
pazaan, line 18: This is nice!

prescription/api/router.go
```diff
@@ -4,29 +4,34 @@ import (
...
 
-	PrescriptionService prescription.Service
-	UserClient          user.Client
+	DeviceSettingsValidator service.DeviceSettingsValidator
+	PrescriptionService     prescription.Service
+	UserClient              user.Client
 }
 
-func NewRouter(p Params) service.Router {
+func NewRouter(p Params) router.Router {
```
pazaan, line 30: Nice clarifying change 👌

devices/client.go
```diff
@@ -0,0 +1,57 @@
...
+	return client
+}
+
+func (c *Client) Initialize(ctx context.Context) error {
+	cfg := &Config{}
+	if err := envconfig.Process("", cfg); err != nil {
+		return err
+	}
+
+	conn, err := grpc.DialContext(ctx, cfg.DevicesClientAddress, grpc.WithInsecure(), grpc.WithBlock())
```
pazaan, line 46: Is `WithInsecure` okay? I assume that encryption is handled by mTLS in linkerd/gloo, but just wanted to be sure.

devices/client.go
toddkazakov: We can finally search for environment variables!

devices/client.go
toddkazakov: It's ok - mTLS is indeed handled by linkerd.

PR #509, opened 2020-08-11
[BACK-1572] Migrate or archive duplicate `jellyfish` Omnipod data

pazaan:
Although [BACK-1379] fixed the `deviceId` field with additional
apostrophes in Omnipod data, it did not migrate the `id` and `_id`
fields, which are required for correct deduplication and/or archival.

Since the `deviceId` field has already had the apostrophes removed,
this new migration script finds duplicates by generating the expected
hash and comparing it against Mongo, and either fixes data or archives
duplicates accordingly.

* Fixes [BACK-1572]

[BACK-1379]: https://tidepool.atlassian.net/browse/BACK-1379
[BACK-1572]: https://tidepool.atlassian.net/browse/BACK-1572

### DIFF
--- a/migrations/back-1572/back-1572.go
+++ b/migrations/back-1572/back-1572.go
@@ -1,0 +1,212 @@
+package main
+
+import (
+	"time"
+
+	"github.com/globalsign/mgo"
+	"github.com/globalsign/mgo/bson"
+	"github.com/urfave/cli"
+
+	"github.com/tidepool-org/platform/application"
+	"github.com/tidepool-org/platform/errors"
+	migrationMongo "github.com/tidepool-org/platform/migration/mongo"
+	storeStructuredMongo "github.com/tidepool-org/platform/store/structured/mongo"
+)
+
+func main() {
+	application.RunAndExit(NewMigration())
+}
+
+type Migration struct {
+	*migrationMongo.Migration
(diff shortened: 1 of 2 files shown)


migrations/back-1572/back-1572.go
```diff
@@ -0,0 +1,212 @@
...
+
+func (m *Migration) migrateOmnipodDocuments(dataSession *storeStructuredMongo.Session) (int, int) {
+	logger := m.Logger()
+
+	logger.Debug("Finding distinct users")
+
+	var userIDs []string
+	var hashUpdatedCount, archivedCount int
+
+	err := dataSession.C().Find(bson.M{}).Distinct("_userId", &userIDs)
```
toddkazakov, line 94: Does it make sense to narrow down the query by listing only datums created after a given date and are active? If it's fast enough maybe it's ok to leave it as is.

pazaan: The deduplication only works on ID - it doesn't care whether data is inactive or not, so we still need to take them into account (a bug, to be sure, but still how it currently works).

toddkazakov: If that's the case, should the `"_active": true` condition be removed from the queries below?

PR #512, opened 2020-08-17
[BACK-1532] Add prescription temporary correction ranges

toddkazakov:
- Adds pre-meal temporary correction range 
- Adds workout temporary correction range
- Adds validation for temporary corrections according to [Therapy Settings Document](https://docs.google.com/document/d/1uKTktpTe8HiKQjyUvljhcE_yHlKfYAaEAMoQDmpCHdo/edit)

### DIFF
--- a/data/blood/glucose/target.go
+++ b/data/blood/glucose/target.go
@@ -71,6 +71,39 @@ func (t *Target) Normalize(normalizer data.Normalizer, units *string) {
 	}
 }
 
+type Bounds struct {
+	Lower float64
+	Upper float64
+}
+
+// GetBounds returns the upper and lower bounds expressed by the range attributes.
+// Returns nil if the combination of attributes is invalid.
+func (t *Target) GetBounds() *Bounds {
+	if t.Target != nil && t.Range != nil {
+		return &Bounds{
+			Lower: *t.Target - *t.Range,
+			Upper: *t.Target + *t.Range,
+		}
+	} else if t.Target != nil && t.High != nil {
+		return &Bounds{
+			Lower: *t.Target,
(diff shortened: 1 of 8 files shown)


data/blood/glucose/target.go
```diff
@@ -71,6 +71,39 @@ func (t *Target) Normalize(normalizer data.Normalizer, units *string) {
...
+
+// GetBounds returns the upper and lower bounds expressed by the range attributes.
+// Returns nil if the combination of attributes is invalid.
+func (t *Target) GetBounds() *Bounds {
+	if t.Target != nil && t.Range != nil {
+		return &Bounds{
+			Lower: *t.Target - *t.Range,
+			Upper: *t.Target + *t.Range,
+		}
+	} else if t.Target != nil && t.High != nil {
```
jamesraby, line 87: Do not fully understand the logic here - almost seems like the last clause should be moved up to this spot

jamesraby (dismissed):
Just check the one comment on creating the bounds object

data/blood/glucose/target.go
toddkazakov: This replicates the exact logic that's currently in the validator. Unfortunately, this struct allows the same data to be represented in multiple ways. I found examples in the db of the different combinations of fields that are allowed by the validation rules, so we can't easily refactor the code. 

The order of the clauses here doesn't matter, because we can't possibly have an invalid combination of fields in the struct - `Validate(...)` takes care of this.

pazaan (dismissed):
LGTM

pazaan (approved):
LGTM :shipit:

PR #513, opened 2020-08-21
Added migration for back-393

derrickburns:
This adds the `sharerId` field to existing gatekeeper.perms documents.


pazaan (changes requested):
Remove the binary (doesn't need to be committed).
Otherwise looks good!

derrickburns:
@pazaan Binary removed.

PR #516, opened 2020-08-28
Fix mongo aliases

pazaan:
The `echo ?` in the alias would cause shell completion issues if there
is a file in the PWD named a single letter. This would break the mongo
shell alias.
Quoted the `?` so that it always returns `?`.


derrickburns:
Ship it

PR #517, opened 2020-08-28
Fix bug introduced in last Dockerfile.tools change

pazaan:
Production image wasn't building correctly due to a line continuation
that wasn't removed.


toddkazakov (approved):
LGTM

PR #521, opened 2020-09-17
Marketo kafka message sender

HaroldBernard:
making pr to deploy for test and to eventually merge to master


kafka/client/client.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+//Kafka struct containing the kafka topic and broker
+type Kafka struct {
+	topic  string
+	broker string
+}
+
+//NewKafka creates a kafka struct containing the kafka topic and broker
+func (k *Kafka) NewKafka() *Kafka {
```
derrickburns, line 26: I recommend splitting this into two functions:  1) one that reads variables from the environment (using envconfig) and produces a struct that just contains the configuration 2) one that creates a Kafka object from the configuration struct.

kafka/client/client.go
```diff
@@ -0,0 +1,89 @@
...
+		topicWithPrefix,
+		broker,
+	}
+}
+
+// Initialize COME BACK AND REVIEW THIS
+var Initialize CloudEventsClient = &Kafka{}
+
+// KafkaSender sends message to correct topic and broker
+func (k *Kafka) KafkaSender() *kafka_sarama.Sender {
```
derrickburns, line 42: Since the code can fail, it should probably return an error code also.

kafka/client/client.go
```diff
@@ -0,0 +1,89 @@
...
+
+	sender, err := kafka_sarama.NewSender([]string{k.NewKafka().broker}, saramaConfig, k.NewKafka().topic)
+	if err != nil {
+		log.Printf("failed to create protocol: %s", err.Error())
+	}
+	return sender
+}
+
+// KafkaClient builds kafka client
+func (k *Kafka) KafkaClient(Sender *kafka_sarama.Sender) cloudevents.Client {
```
derrickburns, line 55: Since this can fail, it should also return an error code.

derrickburns (changes requested):

kafka/client/client.go
HaroldBernard: done

kafka/client/client.go
HaroldBernard: done

kafka/client/client.go
HaroldBernard: done

kafka/client/client.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/Shopify/sarama"
+	"github.com/cloudevents/sdk-go/protocol/kafka_sarama/v2"
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/google/uuid"
+	"github.com/kelseyhightower/envconfig"
+)
+
+//CloudEventsClient is the method signature for Kafka message
+type CloudEventsClient interface {
+	KafkaMessage(event string, userID string, email string, role []string)
```
toddkazakov, line 16: The name and/or signature seem a little off to me - the function name is generic, but the signature is too specific (implies user events only).

user/service/client/client.go
```diff
@@ -202,6 +203,14 @@ func (c *Client) Delete(ctx context.Context, id string, deleet *user.Delete, con
...
 		}
 	}
+	//Sending delete user event to kafka
+	var role []string
+	if result != nil && result.Roles != nil {
+		role = *result.Roles
+		logger.Infof("PResult: %v", *result.Roles)
+	}
+	logger.Infof("PResult: %v", *result)
+	c.CloudEventsClient().KafkaMessage("delete-user", id, *result.Username, role)
```
toddkazakov, line 213: Consider extracting "delete-user" event name to a new constant.

kafka/client/client.go
HaroldBernard: Yeah you are right. Thanks I need to work on naming. Would it make more sense to be more marketo specific for the function? Like MarketoKafkaMessage?

HaroldBernard: Never mind. Addressed by making this function more general and extracted all specific user even details

user/service/client/client.go
HaroldBernard: addressed in changes recent commit

user/service/client/client.go
```diff
@@ -202,6 +204,23 @@ func (c *Client) Delete(ctx context.Context, id string, deleet *user.Delete, con
...
 	}
+	//Sending delete user event to kafka
+	userName := *result.Username
+	logger.Infof("preparing to send delete user kafka message for %v", userName)
+	if !strings.HasSuffix(userName, "@tidepool.io") && !strings.HasSuffix(userName, "@tidepool.org") {
+		var role []string
+		if result != nil && result.Roles != nil {
+			role = *result.Roles
+		}
+		eventType := "delete-user"
```
toddkazakov, line 215: I suggest you define this as `const DELETE_USER_EVENT_TYPE="delete-user"` at the top of this file

kafka/client/client.go
```diff
@@ -0,0 +1,81 @@
...
+
+// KafkaClient builds kafka client
+func (k *Kafka) KafkaClient(Sender *kafka_sarama.Sender) (cloudevents.Client, error) {
+	c, err := cloudevents.NewClient(Sender, cloudevents.WithTimeNow(), cloudevents.WithUUIDs())
+	return c, err
+}
+
+// KafkaMessage produces kafka message
+func (k *Kafka) KafkaMessage(data map[string]interface{}) {
+	event := data["event"].(string)
```
toddkazakov, line 53: This can panic if the payload doesn't have `event` defined. `KafkaMessage(event string, data map[string]interface{})` might a better function signature.

kafka/client/client.go
```diff
@@ -0,0 +1,81 @@
...
+	e := cloudevents.NewEvent()
+	e.SetID(uuid.New().String())
+	e.SetType(event)
+	e.SetSource("github.com/tidepool-org/platform/kafka/client")
+	_ = e.SetData(cloudevents.ApplicationJSON, data)
+
+	kafkaSender, err := k.KafkaSender()
+	if err != nil {
+		log.Printf("failed to create client, %v", err)
```
toddkazakov, line 62: It's better to use platform's `log/logger` for consistency with other services.

user/service/client/client.go
```diff
@@ -202,6 +204,23 @@ func (c *Client) Delete(ctx context.Context, id string, deleet *user.Delete, con
 			logger.WithError(err).Error("Unable to destroy profile")
 		}
 	}
+	//Sending delete user event to kafka
+	userName := *result.Username
+	logger.Infof("preparing to send delete user kafka message for %v", userName)
+	if !strings.HasSuffix(userName, "@tidepool.io") && !strings.HasSuffix(userName, "@tidepool.org") {
```
toddkazakov, line 210: I think this logic belongs to the consumer (i.e. the new marketo service). You might have different consumers which might be interested in all events, not only for users which don't have a tidepool account.

user/service/client/client.go
```diff
@@ -202,6 +204,23 @@ func (c *Client) Delete(ctx context.Context, id string, deleet *user.Delete, con
...
+		var role []string
+		if result != nil && result.Roles != nil {
+			role = *result.Roles
+		}
+		eventType := "delete-user"
+		userEventMessage := map[string]interface{}{
+			"event": eventType,
```
toddkazakov, line 217: Do you need the event type as part. of the payload? It's already sent as cloud event type.

user/service/client/client.go
```diff
@@ -202,6 +204,23 @@ func (c *Client) Delete(ctx context.Context, id string, deleet *user.Delete, con
 			logger.WithError(err).Error("Unable to destroy profile")
 		}
 	}
+	//Sending delete user event to kafka
+	userName := *result.Username
+	logger.Infof("preparing to send delete user kafka message for %v", userName)
+	if !strings.HasSuffix(userName, "@tidepool.io") && !strings.HasSuffix(userName, "@tidepool.org") {
+		var role []string
+		if result != nil && result.Roles != nil {
+			role = *result.Roles
```
toddkazakov, line 213: `roles` might be a better name, because soon we'll have more than one role.

user/service/client/client.go
```diff
@@ -202,6 +204,23 @@ func (c *Client) Delete(ctx context.Context, id string, deleet *user.Delete, con
 			logger.WithError(err).Error("Unable to destroy profile")
 		}
 	}
+	//Sending delete user event to kafka
```
toddkazakov, line 207: You should probably extract this block as a separate method

user/service/client/client.go
```diff
@@ -202,6 +204,23 @@ func (c *Client) Delete(ctx context.Context, id string, deleet *user.Delete, con
...
+	userName := *result.Username
+	logger.Infof("preparing to send delete user kafka message for %v", userName)
+	if !strings.HasSuffix(userName, "@tidepool.io") && !strings.HasSuffix(userName, "@tidepool.org") {
+		var role []string
+		if result != nil && result.Roles != nil {
+			role = *result.Roles
+		}
+		eventType := "delete-user"
+		userEventMessage := map[string]interface{}{
```
toddkazakov, line 216: I'd publish the entire user object.

user/service/client/client.go
HaroldBernard: will do

kafka/client/client.go
HaroldBernard: ok makes sense. that is what I originally had. will do this

kafka/client/client.go
HaroldBernard: ok will do

user/service/client/client.go
HaroldBernard: copying from shoreline: "The only reason I added the tidepool email address check here (it is already in marketo) is because ginny is creating and deleting test accounts around every 3 minutes or so and for the same reason as above thought it was unnecessary logging, messages, and noise."

I can remove if does not matter

user/service/client/client.go
HaroldBernard: ah ok thank you will do

user/service/client/client.go
HaroldBernard: will do

user/service/client/client.go
HaroldBernard: Background: this is what I originally was doing and from memory @derrickburns correct me if I am wrong, was advised to just just send the user ID and check the database for the state of the user. Then ran into issue with finding the user (now obviously) in the database because the user was deleted. Since marketo needed these fields I added them to the payload.

Let me know if I should just pass the entire object (dont see why not) or a further discussion is needed

user/service/client/client.go
HaroldBernard: not if I do what I originally had and you suggested here: `This can panic if the payload doesn't have event defined. KafkaMessage(event string, data map[string]interface{}) might a better function signature.`

user/service/client/client.go
toddkazakov: I think it's important to capture events for all accounts even ones used only for testing. This way we can delete them in keycloak.

user/service/client/client.go
toddkazakov: Actually, the linter suggests `DeleteUserEventType`

user/service/client/client.go
toddkazakov: You should not read the user from the DB - you could use the API, but imo it's better to just publish the user object on the topic.

HaroldBernard: sorry that is what I meant

kafka/client/client.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	return &userEventsNotifier{
+		EventProducer: producer,
+	}, nil
+}
+
+func (u *userEventsNotifier) NotifyUserDeleted(ctx context.Context, user user.User, userProfile *profile.Profile) error {
+	var fullName = "Tidepool User"
+	if userProfile != nil || userProfile.FullName != nil {
```
toddkazakov, line 36: Should be `&&`

PR #523, opened 2020-09-28
[BACK-1451] Replace `mgo` driver with `go-mongo-driver`

pazaan:
@toddkazakov This draft PR is just the update to the Mongo driver. I'll add separated commits that address the crux of BACK-1451.


auth/store/mongo/provider_session_repository.go
```diff
@@ -127,28 +135,30 @@ func (p *ProviderSessionSession) DeleteAllProviderSessions(ctx context.Context,
...
-		return nil, errors.New("session closed")
-	}
-
 	now := time.Now()
 	logger := log.LoggerFromContext(ctx).WithField("id", id)
 
 	providerSessions := auth.ProviderSessions{}
-	err := p.C().Find(bson.M{"id": id}).Limit(2).All(&providerSessions)
+	opts := options.Find().
+		SetLimit(2)
```
toddkazakov, line 151: There's a unique index on the id field - this query cannot return more than 1 record.

data/deduplicator/deduplicator.go
```diff
@@ -14,9 +14,9 @@ type Factory interface {
 }
 
 type Deduplicator interface {
-	Open(ctx context.Context, session dataStoreDEPRECATED.DataSession, dataSet *dataTypesUpload.Upload) (*dataTypesUpload.Upload, error)
-	AddData(ctx context.Context, session dataStoreDEPRECATED.DataSession, dataSet *dataTypesUpload.Upload, dataSetData data.Data) error
-	DeleteData(ctx context.Context, session dataStoreDEPRECATED.DataSession, dataSet *dataTypesUpload.Upload, selectors *data.Selectors) error
-	Close(ctx context.Context, session dataStoreDEPRECATED.DataSession, dataSet *dataTypesUpload.Upload) error
-	Delete(ctx context.Context, session dataStoreDEPRECATED.DataSession, dataSet *dataTypesUpload.Upload) error
+	Open(ctx context.Context, session dataStoreDEPRECATED.DataRepository, dataSet *dataTypesUpload.Upload) (*dataTypesUpload.Upload, error)
```
toddkazakov, line 17: Can we undeprecate the deprecated data store?

data/service/api/v1/users_data_delete_test.go
```diff
@@ -14,24 +14,24 @@ var _ = Describe("UsersDataDelete", func() {
 	// 			context = NewTestContext()
```
toddkazakov, line 14: We should just delete those commented tests

data/service/context/standard.go
```diff
@@ -25,9 +25,9 @@ type Standard struct {
 	permissionClient        permission.Client
 	dataDeduplicatorFactory deduplicator.Factory
 	dataStoreDEPRECATED     dataStoreDEPRECATED.Store
-	dataSession             dataStoreDEPRECATED.DataSession
+	dataCollection          dataStoreDEPRECATED.DataRepository
```
toddkazakov, line 28: I think naming should be improved to be more consistent - we should stick with either one of `repository`, `collection` or `session`. What do you think?

data/source/store/structured/mongo/mongo.go
```diff
@@ -358,30 +350,36 @@ func (s *Session) Destroy(ctx context.Context, id string, condition *request.Con
...
+func (c *DataCollection) get(ctx context.Context, logger log.Logger, id string, condition *request.Condition) (*dataSource.Source, error) {
 	results := dataSource.SourceArray{}
 	query := bson.M{
 		"id": id,
 	}
 	if condition != nil && condition.Revision != nil {
 		query["revision"] = *condition.Revision
 	}
-	err := s.C().Find(query).Limit(2).All(&results)
+	opts := options.Find().SetLimit(2)
```
toddkazakov, line 371: Similar to one of my previous comments - wherever we have unique indexes we can simplify the logic and use `FindOne`, because we can't have more than 1 record.

auth/store/mongo/provider_session_repository.go
pazaan: Good catch! That's pre-existing, but we should definitely fix it while we're here. 👍

data/deduplicator/deduplicator.go
pazaan: I was thinking about this too, I just forgot to do it 😛

data/service/context/standard.go
pazaan: Yes, we should. I had tried to standardise on `repository`, but you found one that slipped through. Thanks!

store/structured/mongo/config.go
```diff
@@ -1,108 +1,88 @@
...
-	Timeout          time.Duration `json:"timeout"`
-	OptParams        *string       `json:"optParams"`
-}
+	"go.uber.org/fx"
+)
 
-//NewConfig creates and returns an incomplete Config object
-func NewConfig() *Config {
-	return &Config{
+func NewConfig(lifecycle fx.Lifecycle) *Config {
```
toddkazakov, line 18: I saw that you're passing a nil Lifecycle in some cases. To avoid this, I suggest splitting this to `NewConfig() *Config` which creates the struct and `AppendLifecycleHooksToConfig(config *Config, lifecycle fx.Lifecycle)` which appends the lifecycle hooks. You can then just call `NewConfig()` from the code that doesn't use `fx`. `NewStore(p Params) (*Store, error)` should also be split into two separate functions.

Here's an example:
```go

var StoreModule = fx.Options(
	fx.Provide(NewConfig),
	fx.Invoke(AppendLifecycleHooksToConfig),
	fx.Provide(NewStore),
	fx.Invoke(AppendLifecycleHooksToStore),
)


func NewStore(config *Config) *Store {
	return &Store{
		config: config,
		ctx:    context.Background(),
	}
}

func AppendLifecycleHooksToStore(store *Store, lifecycle fx.Lifecycle) {
	lifecycle.Append(fx.Hook{
		OnStart: func(ctx context.Context) error {
			return store.Initialize(ctx)
		},
		OnStop: func(ctx context.Context) error {
			return store.Terminate(ctx)
		},
	})
}


func NewConfig() *Config {
	return &Config{}
}

func AppendLifecycleHooksToConfig(cfg *Config, lifecycle fx.Lifecycle) {
	lifecycle.Append(fx.Hook{
		OnStart: func(ctx context.Context) error {
			if err := cfg.Load(); err != nil {
				return err
			}

			return cfg.Validate()
		},
	})
}
```

store/structured/mongo/store.go
```diff
@@ -45,21 +48,31 @@ func NewStore(p Params) (*Store, error) {
...
-		OnStart: func(ctx context.Context) error {
-			return store.Initialize(ctx)
-		},
-		OnStop: func(ctx context.Context) error {
-			return store.Terminate(ctx)
-		},
-	})
+	var err error
+
+	if p.Lifecycle != nil {
```
toddkazakov, line 53: I think this constructor should be split into two separate functions. One for creating the struct and a second one for attaching the lifecycle hooks. I gave you an example of how to do this in a comment in `store/structured/mongo/config.go`.

store/structured/mongo/store.go
```diff
@@ -45,21 +48,31 @@ func NewStore(p Params) (*Store, error) {
...
+		p.Lifecycle.Append(fx.Hook{
+			OnStart: func(ctx context.Context) error {
+				return store.Initialize(ctx)
+			},
+			OnStop: func(ctx context.Context) error {
+				return store.Terminate(ctx)
+			},
+		})
+	} else {
+		// If we're not using `fx`, then do the Initialization as part of `NewStore`
```
toddkazakov, line 63: I have always preferred to keep initialization logic outside of constructors (especially when networking is involved). I think the calling code should call `store.Initialize(ctx)` with its own context.

store/structured/mongo/store.go
```diff
@@ -20,7 +20,10 @@ var StoreModule = fx.Provide(
 type Store struct {
 	client *mongoDriver.Client
 	config *Config
-	ctx    context.Context
+	// TODO: Discuss with TK. Do we event need context? It's not used in the Store, since
+	// it's passed into every function. Maybe on if we have the passed in context as optional,
+	// then use o.ctx?
+	ctx context.Context
```
toddkazakov, line 26: I don't think it's used - must a leftover of me refactoring the some old code

store/structured/mongo/store.go
```diff
@@ -70,16 +83,30 @@ func (o *Store) Initialize(ctx context.Context) error {
 }
 
 func (o *Store) GetRepository(collection string) *Repository {
+	if o.client == nil {
```
toddkazakov, line 86: Perhaps `panic` is better, because this is not meant to be called before the store is initialized.

store/structured/mongo/store.go
```diff
@@ -70,16 +83,30 @@ func (o *Store) Initialize(ctx context.Context) error {
...
+		// TODO: TK - should this return an error instead?
+		return nil
+	}
+
 	return NewRepository(o.GetCollection(collection))
 }
 
 func (o *Store) GetCollection(collection string) *mongoDriver.Collection {
+	if o.client == nil {
```
toddkazakov, line 95: Perhaps `panic` is better, because this is not meant to be called before the store is initialized.

store/structured/mongo/store.go
```diff
@@ -70,16 +83,30 @@ func (o *Store) Initialize(ctx context.Context) error {
...
+		return nil
+	}
+
 	db := o.client.Database(o.config.Database)
 	prefixed := fmt.Sprintf("%s%s", o.config.CollectionPrefix, collection)
 	return db.Collection(prefixed)
 }
 
 func (o *Store) Ping(ctx context.Context) error {
+	if o.client == nil {
```
toddkazakov, line 106: Perhaps `panic` is better, because this is not meant to be called before the store is initialized.

task/queue/queue.go
```diff
@@ -232,8 +236,7 @@ func (q *Queue) dispatchTasks(ctx context.Context) time.Duration {
 func (q *Queue) dispatchTask(ctx context.Context, tsk *task.Task) {
 	logger := q.logger.WithField("taskId", tsk.ID)
 
-	ssn := q.store.NewTaskSession()
-	defer ssn.Close()
+	ssn := q.store.NewTaskRepository()
```
toddkazakov, line 239:
```suggestion
	repository := q.store.NewTaskRepository()
```

task/queue/queue.go
```diff
@@ -254,8 +257,7 @@ func (q *Queue) completeTask(ctx context.Context, tsk *task.Task) {
 
 	q.workersAvailable++
 
-	ssn := q.store.NewTaskSession()
-	defer ssn.Close()
+	ssn := q.store.NewTaskRepository()
```
toddkazakov, line 260:
```suggestion
	repository := q.store.NewTaskRepository()
```

task/service/service/client.go
```diff
@@ -24,36 +24,26 @@ func NewClient(str taskStore.Store) (*Client, error) {
 }
 
 func (c *Client) ListTasks(ctx context.Context, filter *task.TaskFilter, pagination *page.Pagination) (task.Tasks, error) {
-	ssn := c.taskStore.NewTaskSession()
-	defer ssn.Close()
-
+	ssn := c.taskStore.NewTaskRepository()
```
toddkazakov, line 27:
```suggestion
	repository := c.taskStore.NewTaskRepository()
```

task/service/service/client.go
```diff
@@ -24,36 +24,26 @@ func NewClient(str taskStore.Store) (*Client, error) {
...
 	return ssn.ListTasks(ctx, filter, pagination)
 }
 
 func (c *Client) CreateTask(ctx context.Context, create *task.TaskCreate) (*task.Task, error) {
-	ssn := c.taskStore.NewTaskSession()
-	defer ssn.Close()
-
+	ssn := c.taskStore.NewTaskRepository()
```
toddkazakov, line 32:
```suggestion
	repository := c.taskStore.NewTaskRepository()
```

task/service/service/client.go
```diff
@@ -24,36 +24,26 @@ func NewClient(str taskStore.Store) (*Client, error) {
...
 	return ssn.CreateTask(ctx, create)
 }
 
 func (c *Client) GetTask(ctx context.Context, id string) (*task.Task, error) {
-	ssn := c.taskStore.NewTaskSession()
-	defer ssn.Close()
-
+	ssn := c.taskStore.NewTaskRepository()
```
toddkazakov, line 37:
```suggestion
	repository := c.taskStore.NewTaskRepository()
```

task/service/service/client.go
```diff
@@ -24,36 +24,26 @@ func NewClient(str taskStore.Store) (*Client, error) {
...
 	return ssn.GetTask(ctx, id)
 }
 
 func (c *Client) UpdateTask(ctx context.Context, id string, update *task.TaskUpdate) (*task.Task, error) {
-	ssn := c.taskStore.NewTaskSession()
-	defer ssn.Close()
-
+	ssn := c.taskStore.NewTaskRepository()
```
toddkazakov, line 42:
```suggestion
	repository := c.taskStore.NewTaskRepository()
```

task/service/service/client.go
```diff
@@ -24,36 +24,26 @@ func NewClient(str taskStore.Store) (*Client, error) {
...
 	return ssn.UpdateTask(ctx, id, update)
 }
 
 func (c *Client) DeleteTask(ctx context.Context, id string) error {
-	ssn := c.taskStore.NewTaskSession()
-	defer ssn.Close()
-
+	ssn := c.taskStore.NewTaskRepository()
```
toddkazakov, line 47:
```suggestion
	repository := c.taskStore.NewTaskRepository()
```

task/service/service/service.go
```diff
@@ -122,7 +125,7 @@ func (s *Service) initializeTaskStore() error {
 func (s *Service) terminateTaskStore() {
 	if s.taskStore != nil {
 		s.Logger().Debug("Closing task store")
-		s.taskStore.Close()
+		s.taskStore.Terminate(nil)
```
toddkazakov, line 128:
```suggestion
		s.taskStore.Terminate(context.Background())
```

task/store/mongo/mongo.go
```diff
@@ -32,38 +33,64 @@ func NewStore(cfg *storeStructuredMongo.Config, lgr log.Logger) (*Store, error)
...
+func (s *Store) TaskRepository() *TaskRepository {
+	return &TaskRepository{
+		s.Store.GetRepository("tasks"),
 	}
 }
 
 func (s *Store) EnsureIndexes() error {
-	ssn := s.taskSession()
-	defer ssn.Close()
+	ssn := s.TaskRepository()
```
toddkazakov, line 47:
```suggestion
	repository := s.TaskRepository()
```

task/store/mongo/mongo.go
```diff
@@ -137,40 +162,41 @@ func (t *TaskSession) CreateTask(ctx context.Context, create *task.TaskCreate) (
...
-		return nil, errors.New("session closed")
-	}
-
 	now := time.Now()
 	logger := log.LoggerFromContext(ctx).WithField("id", id)
 
-	tsks := task.Tasks{}
-	err := t.C().Find(bson.M{"id": id}).Limit(2).All(&tsks)
+	tasks := task.Tasks{}
+	opts := options.Find().SetLimit(2)
```
toddkazakov, line 177: I think this can be replaced with `FindOne`

user/store/structured/mongo/mongo.go
```diff
@@ -182,12 +170,17 @@ func (s *Session) get(logger log.Logger, id string, condition *request.Condition
 		query["revision"] = *condition.Revision
 	}
 	query = storeStructuredMongo.ModifyQuery(query, queryModifiers...)
-	err := s.C().Find(query).Limit(2).All(&results)
+	opts := options.Find().SetLimit(2)
```
toddkazakov, line 173: I think this can be replaced with `FindOne`

toddkazakov (changes requested):
Good job!

- There are few cases where sorting order has changed, which I believe should be fixed
- I think the store and config initialization methods can be cleaned up a bit so they can be used from code that doesn't use `fx` without passing `nil` lifecycles
- Many naming suggestions which I don't see as criticial

data/service/api/v1/users_data_delete_test.go
```diff
@@ -5,117 +5,4 @@ import (
 )
 
 var _ = Describe("UsersDataDelete", func() {
```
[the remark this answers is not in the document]
pazaan: Or should we implement unit test? :p

data/store/mongo/mongo.go
```diff
@@ -101,32 +138,33 @@ func (d *DataSession) GetDataSetsForUserByID(ctx context.Context, userID string,
...
-
 	now := time.Now()
 
 	dataSets := []*upload.Upload{}
 	selector := bson.M{
 		"uploadId": dataSetID,
 		"type":     "upload",
 	}
-	err := d.C().Find(selector).Limit(2).All(&dataSets)
+	opts := options.Find().SetLimit(2)
```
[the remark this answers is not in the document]
pazaan: Yes, good catch!

data/store/mongo/mongo.go
```diff
@@ -634,10 +620,10 @@ func (d *DataSession) UnarchiveDeviceDataUsingHashesFromDataSet(ctx context.Cont
 			},
 		},
 	}
-	pipe := d.C().Pipe(pipeline)
-	iter := pipe.Iter()
+	// TODO: Do we _really_ want an Aggregate here?
```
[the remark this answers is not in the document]
pazaan: I'll remove that TODO. It will show up in the Slow Query reports if it needs to be addressed.

user/service/client/client.go
```diff
@@ -95,9 +95,9 @@ func (c *Client) Delete(ctx context.Context, id string, deleet *user.Delete, con
 		return false, err
 	}
 
-	session := c.UserStructuredStore().NewUserRepository()
+	repository := c.UserStructuredStore().NewUserRepository()
 
-	result, err := session.Get(ctx, id, condition)
+	result, err := repository.Get(ctx, id, condition)
```
toddkazakov, line 100: We should use `context.Background()`, otherwise the context will be terminated for requests which time out. This will be deprecated soon, but I think it's still important.

toddkazakov (changes requested):
Looks great. In previous review I missed that we should pass a background context to all deletion queries so they are not terminated when the request times out.

user/service/client/client.go
pazaan: Ah yes, good point!

toddkazakov (dismissed):
LGTM 🚀

store/structured/mongo/config.go
```diff
@@ -27,15 +28,15 @@ func LoadConfig() (*Config, error) {
...
-	Scheme           string        `json:"scheme" envconfig:"TIDEPOOL_STORE_SCHEME"`
-	Addresses        []string      `json:"addresses" envconfig:"TIDEPOOL_STORE_ADDRESSES" required:"true"`
-	TLS              bool          `json:"tls" envconfig:"TIDEPOOL_STORE_TLS" default:"true"`
-	Database         string        `json:"database" envconfig:"TIDEPOOL_STORE_DATABASE" required:"true"`
-	CollectionPrefix string        `json:"collectionPrefix" envconfig:"TIDEPOOL_STORE_COLLECTION_PREFIX"`
-	Username         *string       `json:"-" envconfig:"TIDEPOOL_STORE_USERNAME"`
-	Password         *string       `json:"-" envconfig:"TIDEPOOL_STORE_PASSWORD"`
-	Timeout          time.Duration `json:"timeout" envconfig:"TIDEPOOL_STORE_TIMEOUT" default:"60s"`
-	OptParams        *string       `json:"optParams" envconfig:"TIDEPOOL_STORE_OPT_PARAMS"`
+	Scheme           string        `json:"scheme" envconfig:"SCHEME"`
```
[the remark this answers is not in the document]
pazaan: I can just revert it 😄

toddkazakov (dismissed):

toddkazakov (approved):
👍

PR #524, opened 2020-09-29
Allow un-setting prescription settings

toddkazakov:
When prescription are updated we're using the mongo `$set` operator that doesn't allow un-setting fields when it's used along with `omitempty` bson struct tag. This caused `latestRevision` attribute to be populated by the merged object of latestRevision + non-empty updated fields.

This bug causes intermittent unit test failures in the `master` branch.

### DIFF
--- a/prescription/revision.go
+++ b/prescription/revision.go
@@ -204,25 +204,25 @@ func (r *Revision) GetPrescriberUserID() string {
 }
 
 type Attributes struct {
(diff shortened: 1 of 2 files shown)


pazaan (approved):
LGTM! :shipit:

tjotala (approved):
LGTM :shipit:

PR #525, opened 2020-09-29
Do not build image and notification services

toddkazakov:
This should reduce our build times.

The image service is not used at the moment.
The notification service is just a shell without any functionality.


pazaan (approved):
LGTM! 🚀

derrickburns (approved):

tjotala (approved):
LGTM :shipit: good call on culling dormant services.

PR #526, opened 2020-10-02
[BACK-1737] Delete blobs, data, data sources, restricted tokens and provider sessions when users are deleted

toddkazakov:


auth/events/events.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	logger.Infof("Deleting provider sessions for user")
+	if err := u.client.DeleteAllProviderSessions(u.ctx, payload.UserID); err != nil {
+		errs = append(errs, err)
+		logger.WithError(err).Error("unable to delete provider sessions for user")
+	}
+
+	if len(errs) != 0 {
+		return errors.New("Unable to auth data for user")
```
derrickburns, line 44: Is this error message correct?

auth/service/service/service.go
```diff
@@ -39,6 +45,18 @@ func New() *Service {
 	}
 }
 
+func (s *Service) Run() error {
+	errs := make(chan error, 0)
```
derrickburns, line 49: Does this leak channels, or is it called only once?

blob/service/service.go
```diff
@@ -33,6 +39,18 @@ func New() *Service {
...
+func (s *Service) Run() error {
+	errs := make(chan error, 0)
+	go func() {
+		errs <- s.userEventsHandler.Run(context.Background())
+	}()
+	go func() {
+		errs <- s.Service.Run()
+	}()
+
+	return <-errs
```
derrickburns, line 51: Same issue about leaks.

auth/service/service/service.go
toddkazakov: Called once on startup

blob/service/service.go
toddkazakov: Called once on startup

auth/events/events.go
toddkazakov: Fixed

derrickburns (approved):
Lgtm

PR #533, opened 2020-11-30
BACK-1743 Remove user service

toddkazakov:
This PR removes the user service, so all requests are handled by shoreline. Dependent packages which were used solely by the user service were also removed.


jamesraby (approved):
Nice!!  Love reducing our footprint

adinhodovic (approved):

PR #536, opened 2021-01-07
Fix validation of workout and pre-meal correction ranges

toddkazakov:
- Remove legacy validation logic
- Use correct device absolute bounds


jamesraby (approved):
lgtm

PR #537, opened 2021-01-08
Allow uuids as user ids

toddkazakov:


adinhodovic (approved):

toddkazakov:
@tjotala can you please review?

auth/user.go
```diff
@@ -34,4 +34,4 @@ func ErrorValueStringAsUserIDNotValid(value string) error {
 	return errors.Preparedf(structureValidator.ErrorCodeValueNotValid, "value is not valid", "value %q is not valid as user id", value)
 }
 
-var idExpression = regexp.MustCompile("^[0-9a-f]{10}$")
+var idExpression = regexp.MustCompile("(^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$)|(^[0-9a-f]{10}$)")
```
tjotala, line 37: Question: wouldn't the regexp be `^(uuid|oldstyle)$` - that is the anchors and parens surrounding the two alternatives? I didn't think the `|` works outside of parens, unless there's an implied set of parens outside of the whole pattern...?

toddkazakov: That was my initial attempt, but a unit test failed. I checked it with regexr and it matched a trailing substring - if you test the string `01234567890` it will match `1234567890`. Now that I think of it I might have missed a pair of brackets.

tjotala: The new pattern has an extra `^` anchor (I think redundant/harmless) plus some extra parens.

Seems like the pattern should be: `\A(?:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}|[0-9a-f]{10})\z`. Tested with https://regoio.herokuapp.com/ using `4fe2bfc3-f149-46d2-9ca2-9af9fc305a03` and `1234567890`, plus butchering them in various ways.

toddkazakov: Thanks for chiming in here I managed to screw this one up multiple times. Your regexp seems to be better.. TIL that `\A` and `\z` should be used for denoting start and end of string!

tjotala (approved):
LGTM :shipit:

PR #548, opened 2021-04-06
Gracefully shutdown the services

adinhodovic:


auth/service/service/service.go
```diff
@@ -46,14 +49,23 @@ func New() *Service {
...
+	errs := make(chan error)
 	go func() {
-		errs <- s.userEventsHandler.Run(context.Background())
+		errs <- s.userEventsHandler.Run()
 	}()
 	go func() {
 		errs <- s.Service.Run()
 	}()
 
+	go func() {
```
toddkazakov, line 60: I think it's best if the definition of this function is defined in `service/service/service.go` or in the runner so this code can be shared. You could define a function that takes a channel as parameter and then just invoke it in each service with `go s.listenForSignals(doneChan)`. I think it's better if you don't use the errors channel - a `doneChan := make(chan struct{})` will be more appropriate.

blob/service/service.go
```diff
@@ -72,13 +84,12 @@ func (s *Service) Initialize(provider application.Provider) error {
 }
 
 func (s *Service) Terminate() {
-	s.terminateRouter()
+	s.Authenticated.Terminate()
```
toddkazakov, line 87: Great catch!

toddkazakov (commented):
As discussed in slack, I think it's better to extract the common logic for handling OS signals.

toddkazakov (dismissed):

toddkazakov (approved):

PR #549, opened 2021-04-14
[BACK-1914] Fix combo boluses for 600 Series Medtronic Pumps

toddkazakov:
Now the `bolus/combination` validator allows both `extended` and `normal` to be zero if either `expectedNormal` or `expectedExtended` are larger than 0.

### DIFF
--- a/data/types/bolus/combination/combination.go
+++ b/data/types/bolus/combination/combination.go
@@ -67,10 +67,13 @@ func (c *Combination) Validate(validator structure.Validator) {
 		extendedExpectedValidator := validator.Float64("expectedExtended", c.ExtendedExpected)
 		extendedExpectedValidator.InRange(ExtendedMinimum, ExtendedMaximum)
 		if c.Normal != nil {
-			if *c.Normal != NormalMinimum {
-				extendedExpectedValidator.Exists()
-			} else {
+			if *c.Normal == NormalMinimum {
+				if c.ExtendedExpected == nil {
+					validator.Float64("expectedNormal", c.NormalExpected).GreaterThan(NormalMinimum)
+				}
 				extendedExpectedValidator.GreaterThan(ExtendedMinimum)
+			} else {
+				extendedExpectedValidator.Exists()
 			}
 		}
 	} else {
@@ -90,6 +93,9 @@ func (c *Combination) Validate(validator structure.Validator) {
 		expectedExtendedValidator := validator.Float64("expectedExtended", c.ExtendedExpected)
 		if c.Extended != nil && *c.Extended >= ExtendedMinimum && *c.Extended <= ExtendedMaximum {
 			if *c.Extended == ExtendedMinimum {
+				if c.Normal != nil && *c.Normal == NormalMinimum {
+					expectedExtendedValidator.GreaterThan(ExtendedMinimum)
+				}
 				expectedExtendedValidator.Exists()
 			}
 			expectedExtendedValidator.InRange(*c.Extended, ExtendedMaximum)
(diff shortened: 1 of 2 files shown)


data/types/bolus/combination/combination.go
```diff
@@ -96,7 +101,7 @@ func (c *Combination) Validate(validator structure.Validator) {
 			// If Normal is zero, then _either_:
 			if c.Extended != nil {
 				if c.NormalExpected == nil {
-					validator.Float64("extended", c.Extended).GreaterThan(ExtendedMinimum)
+					validator.Float64("extended", c.Extended).GreaterThanOrEqualTo(ExtendedMinimum)
```
gniezen, line 104: In the case where
- `normal` is zero,
- `NormalExpected` doesn't exist, and
- `extended` is zero
`ExtendedExpected should be greater than `ExtendedMinimum`

Same for when
- `normal` is zero,
- `extended` is zero
- `extendedExpected` doesn't exist
`NormalExpected should be greater than `ExtendedMinimum`

Or are we checking this somewhere else and I'm missing it?

gniezen (commented):
Some questions inline.

data/types/bolus/combination/combination.go
toddkazakov: Neither of those cases are covered by the current validation. If first three conditions are zero we check that `ExtendedExpected` is in range [0,100] for the first example and `NormalExpected` is in range [0,100] for the second example.

data/types/bolus/combination/combination_test.go
```diff
@@ -1119,6 +1110,31 @@ var _ = Describe("Combination", func() {
...
+					},
+				),
+				Entry("allow normal and extended to be 0 when both normal expected and extended expected are in range",
+					func(datum *combination.Combination) {
+						datum.Duration = pointer.FromInt(0)
+						datum.Extended = pointer.FromFloat64(0.0)
+						datum.Normal = pointer.FromFloat64(0.0)
+						datum.NormalExpected = pointer.FromFloat64(100.0)
+					},
+				),
```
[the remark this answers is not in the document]
toddkazakov: I added a (failing) test and fixed the issue

data/types/bolus/combination/combination.go
toddkazakov: I added failing tests for those 2 cases and will work on a fix later today.

toddkazakov:
It feels like this logic can be refactored and simplified, however it's quite complex and we mostly have negative tests so we risk introducing regressions.

data/types/bolus/combination/combination_test.go
```diff
@@ -1119,6 +1139,31 @@ var _ = Describe("Combination", func() {
...
+						datum.Normal = pointer.FromFloat64(0.0)
+						datum.NormalExpected = pointer.FromFloat64(100.0)
+						datum.ExtendedExpected = nil
+					},
+				),
+				Entry("allow normal and extended to be 0 when normal expected is missing",
+					func(datum *combination.Combination) {
+						datum.Duration = pointer.FromInt(0)
+						datum.Extended = pointer.FromFloat64(0.0)
+						datum.Normal = pointer.FromFloat64(0.0)
```
gniezen, line 1155: Should we add a `datum.ExtendedExpected = pointer.FromFloat64(100.0)` here, so that it's the inverse of the previous test?

data/types/bolus/combination/combination_test.go
```diff
@@ -1119,6 +1139,31 @@ var _ = Describe("Combination", func() {
...
+						datum.NormalExpected = nil
+					},
+				),
+				Entry("allow normal and extended to be 0 when both normal expected and extended expected are in range",
```
gniezen, line 1159: This test mentions `extended expected`, but it's not in the test itself?

gniezen:
Yeah, this logic is hurting my brain, but trying to refactor it is probably not worth it.

gniezen (dismissed):
Couple of clarifying questions, but LGTM! :shipit:

data/types/bolus/combination/combination_test.go
toddkazakov: The default value of `ExtendedExpected` is randomly generated before each test and is always within range. The default value of `NormalExpected` is `nil` and is set explicitly in the test.

data/types/bolus/combination/combination_test.go
toddkazakov: That's a good idea. Pushed a new commit, but this PR will require re-approval.

gniezen (approved):
Still LGTM! :ship:

PR #550, opened 2021-04-23
[BACK-1927] Disallow "1" in access code

tjotala:
Darin noticed that we allow "1" in the prescription access code, while disallowing "I", "0", and "O". For symmetry with "0" and "O" and to avoid input mistakes, we should disallow "1" as well.


toddkazakov (approved):
LGTM

PR #551, opened 2021-04-23
Reset prescription access code correctly to prevent duplicates in sparse index

toddkazakov:


prescription/store/mongo/prescription_repository.go
```diff
@@ -433,7 +435,9 @@ func newMongoUpdateFromPrescriptionUpdate(prescrUpdate *prescription.Update) bso
 		if code != "" {
 			set["accessCode"] = code
 		} else {
-			set["accessCode"] = nil
+			update["$unset"] = bson.M{
```
darinkrauss, line 438: [nit] In older code, I usually created the `unset := bson.M{}` structure, added it to the `update`, and used that for consistency and clarity (like this code does for `set`, but this works as-is.

darinkrauss (approved):
Only one really minor nit (which you can change or not). LGTM!

prescription/store/mongo/prescription_repository.go
toddkazakov: Mongo spits out an error when `unset` is empty:
```
Error: {
                  WriteConcernError: nil,
                  WriteErrors: [
                      {
                          Index: 0,
                          Code: 9,
                          Message: "'$unset' is empty. You must specify a field like so: {$unset: {<field>: ...}}",
                      },
                  ],
                  Labels: nil,
              }
```

I'll leave it as it is for the time being.

PR #552, opened 2021-04-23
[BACK-1928] Fix validation inconsistency of create prescription request validation

toddkazakov:
After I took a fresh look at the code, I don't think we should have different validation logic for validating request fields of the `RevisionCreate` request and for validating the `Revision` attributes. The reason this happened in first place is due to code duplication. This PR simplifies struct validation by removing the duplicate code by making use of embedded structs that are now shared between the revision and request models.


darinkrauss (approved):
A couple of really minor nits (which you can change or not). LGTM!

PR #561, opened 2021-05-12
Do not return a validation errors if there are unparsed fields

toddkazakov:
The object parser returns errors if there are unparsed field in an object. This causes unexpected errors every time a new field is added from a third-party provider even if their change is backward compatible. For this reason I believe it's better if we don't default to this.

Dexcom added a new field to their responses which is undocumented, thus not parsed by us. This resulted in sync failures for all tasks in the last ~20h.


tjotala (approved):
LGTM. :shipit:

PR #569, opened 2021-06-15
[BACK-1294] [BACK-1295] Integrate with clinics service

toddkazakov:
- Refactored existing services and models to support clinics and clinicians
- Refactored api endpoints which are now split in two groups - for clinic level access and for patient access
- Increased test coverage for clinic/patient authorization 
- Only prescribers can change prescription status to "submitted"

Api docs are updated in https://github.com/tidepool-org/TidepoolApi/pull/21


toddkazakov:
@adinhodovic snyk fails, because there's a vulnerable dependency of the mongo driver. I tried upgrading it, but tests fail so I think we should track it as a separate ticket so it can be properly tested.

adinhodovic (approved):
LGTM :+1:

PR #574, opened 2021-07-06
[BACK-1931] Share patient account with clinic when claiming prescription

toddkazakov:
The snyk failure is addressed in #573


Roukoswarf (approved):
looks good.

PR #575, opened 2021-07-26
[BACK-1352] [BACK-1354] [BACK-2025] Add summary stats

Roukoswarf:
Summary service from the platform perspective is functional.

Automatic updating of summaries be tested by sending in a task to the task service with:
```
{
    "name": "summary",
    "type": "org.tidepool.summary.update",
    "priority": 5,
    "availableTime": "2021-06-30T15:04:05.999999999Z",
    "expirationTime": "2022-06-30T15:04:05.999999999Z"
}
```

Directly related to https://github.com/tidepool-org/development/pull/189


data/summary.go
```diff
@@ -0,0 +1,19 @@
+package data
+
+import (
+    "time"
+    "go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+type Summary struct {
+    ID                 primitive.ObjectID  `json:"-" bson:"_id,omitempty"`
+	UserID             string              `json:"UserID" bson:"_userId"`
```
tjotala, line 10: Nit: consider renaming this field `userId` for consistency with existing APIs.
To be fair, we're already inconsistent with `userid`, `userId`, and `_userId` - but I think `userID` would be new.

data/summary.go
```diff
@@ -0,0 +1,19 @@
...
+
+	LastUpdated        time.Time   `json:"lastUpdated" bson:"lastUpdated"`
+	LastData           time.Time   `json:"lastData" bson:"lastData"`
```
tjotala, line 13: Based on reading the code, it looks like `lastData` will be end time. Since we spend a fair bit of effort figuring what is the start time, consider exposing it here since AFAICT it is not (necessarily) same as `lastUpload`.

data/summary.go
```diff
@@ -0,0 +1,19 @@
...
+	LastUpload         time.Time   `json:"lastUpload" bson:"lastUpload"`
+	AverageGlucose     float64     `json:"avgGlucose" bson:"avgGlucose"`
```
tjotala, line 15: Consider leveraging the existing structure from `./data/types/blood/blood.go` to include units (mg/dL vs. mmol/L) along with the raw `avgGlucose` value.

data/service/api/v1/summary.go
```diff
@@ -0,0 +1,113 @@
...
+		dataService.MakeRoute("GET", "/v1/agedsummaries/:minutes", Authenticate(GetAgedSummaries)),
+	}
+}
+
+func GetSummary(dataServiceContext dataService.Context) {
+	res := dataServiceContext.Response()
+	req := dataServiceContext.Request()
+	dataClient := dataServiceContext.DataClient()
+
+	//details := request.DetailsFromContext(req.Context())
```
tjotala, line 27: Consider removing these commented out blocks for clarity, since the access permissions are checked by the router wrapping the function in `Authenticate`. Otherwise, a reader might wonder if we've explicitly disabled permission checks (which is not true).

data/service/api/v1/summary.go
```diff
@@ -0,0 +1,113 @@
...
+	/*if details == nil {
+		request.MustNewResponder(res, req).Error(http.StatusUnauthorized, request.ErrorUnauthenticated())
+		return
+	}*/
+
+	responder := request.MustNewResponder(res, req)
+
+	id := req.PathParam("userId")
+	if id == "" {
```
tjotala, line 36: Curiosity: I've seen this pattern all over the existing service code, but never asked if it is actually needed.
The route declaration already "knows" that this path should include `:userId`, and therefore missing it _ought_ to be grounds for a 404.

data/service/api/v1/summary.go
tjotala: Actually, it looks like `Authenticate` only checks that some auth credential was provided (i.e. the commented out code). It does not appear to be check that the session token in the request is valid for this particular `userId`.

So unless I missed something, this needs more permission validation, similar to `./data/service/api/v1/users_datasets_get.go`.

data/service/api/v1/summary.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+
+func UpdateSummary(dataServiceContext dataService.Context) {
+	res := dataServiceContext.Response()
+	req := dataServiceContext.Request()
+	dataClient := dataServiceContext.DataClient()
+
+	//details := request.DetailsFromContext(req.Context())
+	/*if details == nil {
```
tjotala, line 57: Since this API is (presumably) only called by other services, the permissions check here would be to check that the token is a server token (i.e. `details.isService()`)

data/service/api/v1/summary.go
```diff
@@ -0,0 +1,113 @@
...
+	responder := request.MustNewResponder(res, req)
+
+	minutesStr := req.PathParam("minutes")
+    if minutesStr == "" {
+		responder.Error(http.StatusBadRequest, request.ErrorParameterMissing("minutes"))
+		return
+	}
+
+    minutes64, err := strconv.ParseUint(minutesStr, 10, 32)
+    if err != nil {
```
tjotala, line 100: Consider checking for a reasonable range for minutes. At least `> 0`?

data/service/api/v1/summary.go
```diff
@@ -0,0 +1,113 @@
...
+	responder.Data(http.StatusOK, summary)
+}
+
+
+func GetAgedSummaries(dataServiceContext dataService.Context) {
+	res := dataServiceContext.Response()
+	req := dataServiceContext.Request()
+	dataClient := dataServiceContext.DataClient()
+
+	//details := request.DetailsFromContext(req.Context())
```
tjotala, line 85: I'm not sure I understand what is the use-case for this API?

It seems to return potentially unbounded number of user summaries - without regard to access permissions or user IDs (arbitrary set of user IDs? clinic patient user IDs? users who have shared with the session token owner?).

It should probably take pagination parameters (see `./data/service/api/v1/data_set.go` for an example).

data/service/service/client.go
```diff
@@ -37,6 +39,89 @@ func (c *Client) GetDataSet(ctx context.Context, id string) (*data.DataSet, erro
...
+
+        // create empty summary to ensure calc in the future
+        summary.UserID = id
+        summary.LastUpdated = time.Time{}
+        repository.UpdateSummary(ctx, summary)
+    }
+    return summary, nil
+}
+
+func (c *Client) UpdateSummary(ctx context.Context, id string) (*data.Summary, error) {
```
tjotala, line 64: Does this need to consider race conditions, given that it needs the original summary for rolling calc and the fetch/compute/update operation is not atomic?

data/store/mongo/mongo.go
```diff
@@ -913,3 +925,273 @@ func validateAndTranslateSelectors(selectors *data.Selectors) (bson.M, error) {
...
+
+    totalMinutes := float64(math.Round(endTime.Sub(startTime).Minutes()))
+
+    var inRangeMinutes int64 = 0
+    var belowRangeMinutes int64 = 0
+    var aboveRangeMinutes int64 = 0
+    var totalGlucose float64 = 0
+
+    for _, r := range dataSets {
+        if *r.Value < 3.9 {
```
tjotala, line 1019: Consider making these named constants even if they are universally accepted. Naked numbers always leave room for someone not familiar with diabetes to go "huh, whazzat?"

data/summary.go
```diff
@@ -0,0 +1,19 @@
...
+	TimeInRange        float64     `json:"timeInRange" bson:"timeInRange"`
```
tjotala, line 16: I think it would be good to include the thresholds for TIR in this, even if they are constants right now. That would allow clients to render the TIR and ranges:
image

More to the point though, we may want to allow people to customize their TIR thresholds. We already allow that in Tidepool Web:

image

data/service/api/v1/summary.go
```diff
@@ -0,0 +1,113 @@
...
+
+	responder := request.MustNewResponder(res, req)
+
+	id := req.PathParam("userId")
+	if id == "" {
+		responder.Error(http.StatusBadRequest, request.ErrorParameterMissing("id"))
+		return
+	}
+
+	summary, err := dataClient.GetSummary(req.Context(), id)
```
tjotala, line 41: I think this will return a mostly empty `Summary` object (the NoDoc branch in `dataClient.GetSummary`) **if** there's never been a calculation done on that user's account. Granted, it's a transient corner case for new accounts and accounts that we haven't "refreshed" yet.

Should the API instead return 204, 404 or something equally explicit to say "don't have it yet"?

data/store/mongo/mongo.go
```diff
@@ -913,3 +925,273 @@ func validateAndTranslateSelectors(selectors *data.Selectors) (bson.M, error) {
...
+
+    if err != nil {
+		return nil, errors.Wrap(err, "unable to get last upload date")
+	}
+
+	if err = cursor.All(ctx, &dataSet); err != nil {
+		return nil, errors.Wrap(err, "unable to decode last upload date")
+	}
+
+    lastUpload, err := time.Parse(time.RFC3339Nano, *dataSet[0].CreatedTime)
```
tjotala, line 973: This looks very similar to the `GetLastUpdated` method below, with the exception of using different field for the `lastUpload`. Is that intentional? If the code should be same, then consider simply calling `GetLastUpdated`?

data/store/mongo/mongo.go
```diff
@@ -913,3 +925,273 @@ func validateAndTranslateSelectors(selectors *data.Selectors) (bson.M, error) {
...
+	return summary, err
+}
+
+func (d *SummaryRepository) GetAgedSummaries(ctx context.Context, minutes uint) ([]*data.Summary, error) {
+	if ctx == nil {
+		return nil, errors.New("context is missing")
+	}
+
+	//logger := log.LoggerFromContext(ctx).WithField("id", id)
+	startTime := time.Now().Add(time.Minute * -time.Duration(minutes))
```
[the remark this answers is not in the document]
Roukoswarf: Sadly, no.
That results in a type error, an integer cannot be multiplied by time.Minutes. If it was a constant, it would allow it as constants are dynamically typed to Duration, but integers are not.

data/service/api/v1/summary.go
Roukoswarf: Added check for Service token for the permissions, as this API is only to be used by the task service to find what needs doing.
Will test pagination, need to make the task service handle pagination as well for this change.

data/service/service/client.go
Roukoswarf: It should already be race safe, as it uses the original summary's lastData as it's lower bound, and then dynamically calculates the weighting after finding the latest data, then queries by that specific range.

data/service/api/v1/summary.go
Roukoswarf: Would it be okay practice to return 201 or 202 here?
The intent was to create the object for a future pass, to start calculating summaries on first query, to prevent any unnecessary calculation. (clinic-less users)
Clients can detect a 0 datestamped summary as "no data", but if we rather the clients check for return codes, we could return 201 or 202.

data/store/mongo/mongo.go
Roukoswarf: Intentional, would need GetLastUpdated to just return the last record, and handle returning the Time value in the api code above it.
If it was modified to do that, it could then be generic enough for re-use, as we need both Time and CreatedTime fields for the summary data.

This seems like a worthy change for code reuse sake

data/summary.go
Roukoswarf: This is a larger discussion, though, I don't expect these ranges to be dynamic in the future due to the problem with them being variant, but storing them for the sake of clarity is a worthy cause.

Added them into the summary model as configurable constants for now.

data/service/api/v1/summary.go
Roukoswarf: Made it return 204 if summary returned has 0000-00-00 datestamp, this should also cover for the case of repeated calls which do return a value from the db, but are not ready yet.

data/service/api/v1/summary.go
Roukoswarf: This is technically true in this case, as %20 is never unencoded into an empty arg while still matching the url.
block removed.

data/store/mongo/mongo.go
Roukoswarf: added both values to GetLastUpdated return, removed this repetition.

data/service/api/v1/summary.go
Roukoswarf: Decided against adding pagination to this for now, as writing custom pagination for this may be more confusing than its worth.

data/data_set.go
```diff
@@ -313,7 +313,7 @@ type DataSet struct {
 	Payload             *metadata.Metadata      `json:"payload,omitempty" bson:"payload,omitempty"`
 	Source              *string                 `json:"source,omitempty" bson:"source,omitempty"`
 	State               *string                 `json:"-" bson:"_state,omitempty"`
-	Time                *string                 `json:"time,omitempty" bson:"time,omitempty"`
+	Time                *string              `json:"time,omitempty" bson:"time,omitempty"`
```
tjotala, line 316: SuperTinyNit: is your editor changing spaces to tabs? :)

Roukoswarf: we use tabs in the code, so i set my editor to tabs recently... usually have tabs set to 4 spaces.

data/types/blood/glucose/summary/summary.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+    "time"
+    "go.mongodb.org/mongo-driver/bson/primitive"
+	"github.com/tidepool-org/platform/data/types/blood/glucose"
+)
+
+type Summary struct {
+    ID                 primitive.ObjectID  `json:"-" bson:"_id,omitempty"`
+    UserID             string              `json:"userId" bson:"_userId"`
```
toddkazakov, line 11: We should probably stick to `userId` as an attribute in mongo documents.

data/types/blood/glucose/summary/summary.go
```diff
@@ -0,0 +1,28 @@
...
+
+    // date tracking
+    LastUpdated        time.Time   `json:"lastUpdated" bson:"lastUpdated"`
+    FirstData          time.Time   `json:"firstData" bson:"firstData"`
+    LastData           time.Time   `json:"lastData" bson:"lastData"`
+    LastUpload         time.Time   `json:"lastUpload" bson:"lastUpload"`
+
+    // actual values
+    AverageGlucose     glucose.Glucose     `json:"avgGlucose" bson:"avgGlucose"`
+    TimeInRange        float64             `json:"timeInRange" bson:"timeInRange"`
```
toddkazakov, line 21: Should this be a pointer? When a summary is not yet calculated this will be set to `0` which is different from all datums being out of range.

data/store/mongo/mongo.go
```diff
@@ -913,3 +929,271 @@ func validateAndTranslateSelectors(selectors *data.Selectors) (bson.M, error) {
...
+// assumes all except freestyle is 5 minutes
+func getDuration(row *continuous.Continuous) int64 {
+    if strings.Contains(*row.DeviceID, "AbbottFreeStyleLibre") {
+        return 15
+    }
+    return 5
+}
+
+func (d *DataRepository) CalculateSummary(ctx context.Context, summary *summary.Summary) (*summary.Summary, error) {
+    const lowBloodGlucose = 3.9
```
toddkazakov, line 942: magic number

data/store/mongo/mongo.go
```diff
@@ -913,3 +929,271 @@ func validateAndTranslateSelectors(selectors *data.Selectors) (bson.M, error) {
...
+    const highBloodGlucose = 10
```
toddkazakov, line 943: magic number

toddkazakov (changes requested):
I think the biggest issue is that summaries are not calculated before the summary is fetched. This assumes clients make a call per user to `GET /v1/summary/:userId` endpoint which might not be a viable option, given that summaries will be displayed in patients list view.

Commented out code and debug print statements should be removed and the rest should be properly formatted. CI pipelines ought to pass all stages.

data/store/mongo/mongo.go
Roukoswarf: this should be/is industry standard, and currently there are no plans for it to be configurable per user, due to inconsistencies between users being unclear for clinician use.

data/store/mongo/mongo.go
Roukoswarf: this should be/is industry standard, and currently there are no plans for it to be configurable per user, due to inconsistencies between users being unclear for clinician use.

data/service/service/client.go
```diff
@@ -37,6 +40,95 @@ func (c *Client) GetDataSet(ctx context.Context, id string) (*data.DataSet, erro
...
+        dataRepository := c.dataStore.NewDataRepository()
+
+        // check to ensure the user has data
+        _, _, err := dataRepository.GetLastUpdated(ctx, id)
+
+		if err != nil {
+            return nil, err
+        }
+
+        // create empty summary to ensure calc in the future
```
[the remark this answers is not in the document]
Roukoswarf: Side-effecting remove, POST will create if needed, GET will not create, it removes the self-migration aspect, but is clearer when things are happening.

data/store/mongo/mongo.go
```diff
@@ -913,3 +929,271 @@ func validateAndTranslateSelectors(selectors *data.Selectors) (bson.M, error) {
 
 	return selector, nil
 }
+
+// assumes all except freestyle is 5 minutes
+func getDuration(row *continuous.Continuous) int64 {
```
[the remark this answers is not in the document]
Roukoswarf: renamed

data/store/mongo/mongo.go
```diff
@@ -913,3 +929,271 @@ func validateAndTranslateSelectors(selectors *data.Selectors) (bson.M, error) {
...
+
+    totalMinutes := float64(math.Round(endTime.Sub(startTime).Minutes()))
+
+    var inRangeMinutes int64 = 0
+    var belowRangeMinutes int64 = 0
+    var aboveRangeMinutes int64 = 0
+    var totalGlucose float64 = 0
+    var normalizedValue float64
+
+    for _, r := range dataSets {
```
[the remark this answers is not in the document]
Roukoswarf: extracted and split.

data/store/mongo/mongo_data.go
```diff
@@ -0,0 +1,1042 @@
...
+		return userIDs, errors.New("context is missing")
+	}
+
+	selector := bson.M{
+		"time":    bson.M{"$gte": lastUpdated.Format(time.RFC3339Nano)},
+		"type":    "cbg",
+		"_active": true,
+	}
+
+	result, err := d.Distinct(ctx, "_userId", selector)
```
toddkazakov, line 1011: Is this an indexed query?

data/store/mongo/mongo_data.go
```diff
@@ -0,0 +1,1042 @@
...
+
+	if err != nil {
+		return status, errors.Wrap(err, "unable to get last cbg date")
+	}
+
+	if err = cursor.All(ctx, &dataSet); err != nil {
+		return status, errors.Wrap(err, "unable to decode last cbg date")
+	}
+
+	if dataSet != nil {
```
toddkazakov, line 955: You should check the if `len(dataSet) > 0` instead of checking if the data set is nil. I think it can never be `nil`.

toddkazakov (changes requested):

data/service/service/client.go
```diff
@@ -37,6 +152,209 @@ func (c *Client) GetDataSet(ctx context.Context, id string) (*data.DataSet, erro
...
+		if len(userIDsReqUpdate) >= backfillBatch {
+			break
+		}
+	}
+
+	var summaries []*summary.Summary
+
+	for _, userID := range userIDsReqUpdate {
+		// check to ensure the user has data
+		userExists, err := dataRepository.UserHasData(ctx, userID)
```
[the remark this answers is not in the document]
Roukoswarf: This is old code that wasn't deleted from the round-trip-removal change, it used to need to check when the source of this list was external.

migrations/migrations.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 func main() {
-	quitChannel := make(chan os.Signal)
+	quitChannel := make(chan os.Signal, 1)
```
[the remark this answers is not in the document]
Roukoswarf: yeah, golang wasnt happy with this when running tests using latest golang/tools, and did not seem to upset current prod used versions negatively.

data/store/mongo/mongo_summary.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	// find results on a brand new set
+	if lastUpdated.IsZero() {
+		lastUpdated = time.Now()
+	}
+
+	var summaries []*summary.Summary
+	selector := bson.M{
+		"lastUpdated": bson.M{"$lte": lastUpdated.Add(-20 * time.Minute)},
```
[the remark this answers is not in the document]
Roukoswarf: probably a bit too short.
the task will update 100 users per 5-10 minute cycle, so best case you would get 400 users updated (and many more un-changed users with their lastUpdate set ahead) every 20 minute cycle.
Overhead could be reduced without much change to the actual effect by setting this to 1 hour, leaving room for a maximum of 1200 user updates before previously updated users are considered.

I have changed this to be 60 minutes.

data/service/service/client.go
```diff
@@ -37,6 +152,209 @@ func (c *Client) GetDataSet(ctx context.Context, id string) (*data.DataSet, erro
 	return repository.GetDataSet(ctx, id)
 }
 
+func (c *Client) GetSummary(ctx context.Context, id string) (*summary.Summary, error) {
+	summaryRepository := c.dataStore.NewSummaryRepository()
+	dataRepository := c.dataStore.NewDataRepository()
+
+	userSummary, err := summaryRepository.GetSummary(ctx, id)
+
+	if err == mongo.ErrNoDocuments {
```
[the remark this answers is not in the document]
Roukoswarf: due to a previous review comment fix, this was already dead code, and has been removed.

data/service/api/v1/summary.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/tidepool-org/platform/permission"
+	"github.com/tidepool-org/platform/request"
+	"github.com/tidepool-org/platform/service"
+)
+
+func SummaryRoutes() []dataService.Route {
+	return []dataService.Route{
+		dataService.MakeRoute("GET", "/v1/summaries/:userId", Authenticate(GetSummary)),
+		dataService.MakeRoute("POST", "/v1/summaries/:userId", Authenticate(UpdateSummary)),
+		dataService.MakeRoute("GET", "/v1/summaries", Authenticate(GetAgedSummaries)),
```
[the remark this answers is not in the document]
Roukoswarf: added clarifying comment:
```
// NOTE: the following route does advance lastUpdated for not-aged summaries as it creates the list
// This generally does not change the result of subsequent calls, only the amount of processing required
```

data/store/mongo/mongo_data.go
```diff
@@ -0,0 +1,1042 @@
...
+func (d *DataRepository) GetFreshUsers(ctx context.Context, lastUpdated time.Time) ([]string, error) {
+	var userIDs []string
+
+	if ctx == nil {
+		return userIDs, errors.New("context is missing")
+	}
+
+	selector := bson.M{
+		"time":    bson.M{"$gte": lastUpdated.Format(time.RFC3339Nano)},
+		"type":    "cbg",
```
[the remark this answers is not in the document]
Roukoswarf: yes, for initial release only cbg will be used to keep the numbers simpler.

task/store/mongo/mongo.go
```diff
@@ -100,6 +112,74 @@ func (t *TaskRepository) EnsureIndexes() error {
...
+		&opts,
+	)
+
+	if summaryTask.Err() != nil {
+		if summaryTask.Err() != mongo.ErrNoDocuments {
+			return errors.Wrap(summaryTask.Err(), "unable to create summary backfill task")
+		}
+	}
+
+	TasksStateTotal.WithLabelValues(task.TaskStatePending, create.Type).Inc()
```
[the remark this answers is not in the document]
Roukoswarf: I'm not sure this can happen.
The function will return early for any error that is not ErrNoDocuments, which should cover all cases.
ErrNoDocuments is returned when a task is created as none previously existed, due to the setOnInsert+upsert combination.

data/store/mongo/mongo_data.go
Roukoswarf: no... its not :(
guess i forgot that one when last looking at distinct indexing issues.

The best query i can come up with currently that tricks the engine to use index without a fetch stage is:
```
db.deviceData.explain().distinct("_userId", {time: {$gte:"2001"}, type:"cbg", _active:true, _userId:{$ne: -111}})
```
which is slightly ugly, but probably safe? putting a more specific invalid userid match in, like null, doesnt trick it.

Roukoswarf: perhaps this could also be applied to the previously-abandoned more specific query for DistinctUserIDs as well, which I removed all conditions from to get it to index.

Roukoswarf: No... no its not.
I missed this one in my last pass at trying to get these indexed.
Adding a `"_userId": bson.M{"$ne": -1111}` condition to these queries tricks the distinct into being indexed while being a filtered query, and seems like the cleanest way to trick it.
I'll also go back and add this trick to the `DistinctUserIDs` query to make it only grab relevant users, as well.

data/store/mongo/mongo_summary.go
```diff
@@ -0,0 +1,226 @@
...
+	} else if err != nil {
+		return nil, errors.Wrap(err, "unable to get aged summaries")
+	}
+
+	if err = cursor.All(ctx, &summaries); err != nil {
+		return nil, errors.Wrap(err, "unable to decode aged summaries")
+	}
+
+	for _, v := range summaries {
+		userIDs = append(userIDs, v.UserID)
```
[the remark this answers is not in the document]
Roukoswarf: Renamed to `GetUsersWithSummariesBefore` to closely match it's deviceData equivalent.

toddkazakov (approved):
Looks much better now - lets get this tested to make sure it doesn't cause any issues 👍

tjotala (dismissed):
LGTM

PR #576, opened 2021-08-24
[BACK-1746] Convert string dates to ISODATE in DB

Roukoswarf:
This has been tested using uploader and postman APITests along with manual testing, and covers all found cases using string dates.
Unit tests will most likely fail for now, as they manually specify date strings, and I believe it will be best to gather feedback on the approach, and discuss unit testing strategies before adjusting them.

NOTE: this is current based on the summary commits (PR#575) as there are many cases where summary calculation handles dates.


request/values_parser.go
```diff
@@ -192,6 +192,22 @@ func (v *Values) Time(reference string, layout string) *time.Time {
 	return &timeValue
 }
 
+// the api side is always a string, handle this the same as Time
+func (v *Values) MultiTime(reference string, layout string) *time.Time {
```
toddkazakov, line 196: What does MultiTime mean?

toddkazakov: If I am not mistaken, the implementation is identical to the function above

structure/parser/object_parser.go
```diff
@@ -216,6 +216,39 @@ func (o *Object) Time(reference string, layout string) *time.Time {
...
+
+	rawValue, ok := o.raw(reference)
+	if !ok {
+		return nil
+	}
+
+	switch rawValue.(type) {
+	case time.Time:
+		timeValue = rawValue.(time.Time)
+	case string:
```
toddkazakov, line 232: Code duplication

structure/parser/object_parser.go
```diff
@@ -216,6 +216,39 @@ func (o *Object) Time(reference string, layout string) *time.Time {
...
+			o.base.WithReference(reference).ReportError(ErrorTypeNotTime(rawValue))
+			return nil
+		}
+
+		timeValue, err = time.Parse(layout, stringValue)
+		if err != nil {
+			o.base.WithReference(reference).ReportError(ErrorValueTimeNotParsable(stringValue, layout))
+			return nil
+		}
+	default:
```
toddkazakov, line 244: Code duplication

request/values_parser.go
toddkazakov, line 196: You can simply call `v.Time(reference, layout)` instead of copying the implementation.

data/store/mongo/mongo.go
```diff
@@ -1081,8 +1087,8 @@ func (d *DataRepository) GetLastUpdated(ctx context.Context, id string) (time.Ti
 	}
 
 	if dataSet != nil {
-		lastUpload, err = time.Parse(time.RFC3339Nano, *dataSet[0].CreatedTime)
-		lastData, err = time.Parse(time.RFC3339Nano, *dataSet[0].Time)
+		lastUpload = *dataSet[0].CreatedTime
+		lastData = *dataSet[0].Time
 	} else {
 		return lastUpload, lastData, ErrorUserCbgNotFound
```
toddkazakov, line 1093: This is a code smell - consider returning a struct instead.

toddkazakov (commented):
Looks good overall - I didn't see any functional issues. Formatting, unit tests and CI pipeline must be fixed. Considering making this branch the base for summaries, instead the other way around.

tjotala:
Given that the current API accepts and returns TZ-less `deviceTime` values, changing that to a full date/time with TZ represents a change in the documented API contract. Has anyone checked if any of the client apps will choke if they suddenly see a `deviceTime` returned that does include TZ? The more difficult part would be to ensure that there are no 3rd party apps that choke on it.

toddkazakov:
> Given that the current API accepts and returns TZ-less `deviceTime` values, changing that to a full date/time with TZ represents a change in the documented API contract. Has anyone checked if any of the client apps will choke if they suddenly see a `deviceTime` returned that does include TZ? The more difficult part would be to ensure that there are no 3rd party apps that choke on it.

@tjotala My assumption was that data is fetched only from tide-whisperer, where `deviceTime` is formatted correctly (see https://github.com/tidepool-org/tide-whisperer/pull/97/files#diff-64ba38f0c6d37aef130538ce766000c79f9593bb7942d070469f65cb9a43ee59R341). If I am not mistaken `platform/data` doesn't have a endpoint that returns device date.

toddkazakov:
@tjotala looks like the `deviceTime` attribute is present in the data set, so it might cause problems. Good catch!

Roukoswarf:
I have made MarshalJSON for DeviceTime now drop the TZ as it should, good catch @tjotala

data/store/mongo/mongo.go
Roukoswarf: this is part of the summary code, and will be addressed there.

data/store/mongo/mongo.go
```diff
@@ -635,9 +635,9 @@ func (d *DataRepository) UnarchiveDeviceDataUsingHashesFromDataSet(ctx context.C
 
 	result := struct {
 		ID struct {
-			Active            bool   `bson:"_active"`
-			ArchivedDataSetID string `bson:"archivedDatasetId"`
-			ArchivedTime      string `bson:"archivedTime"`
+			Active            bool      `bson:"_active"`
+			ArchivedDataSetID string    `bson:"archivedDatasetId"`
+			ArchivedTime      time.Time `bson:"archivedTime"`
```
toddkazakov, line 640: Should `ArchivedTime` be of type `*time.Time` to better represent the case when `ArchivedTime` is not present in the database?

data/types/basal/basal_test.go
```diff
@@ -119,7 +120,7 @@ var _ = Describe("Basal", func() {
 			It("returns the expected identity fields", func() {
 				identityFields, err := datum.IdentityFields()
 				Expect(err).ToNot(HaveOccurred())
-				Expect(identityFields).To(Equal([]string{*datum.UserID, *datum.DeviceID, *datum.Time, datum.Type, datum.DeliveryType}))
+				Expect(identityFields).To(Equal([]string{*datum.UserID, *datum.DeviceID, (*datum.Time).Format(types.TimeFormat), datum.Type, datum.DeliveryType}))
```
toddkazakov, line 123: This test won't fail if `types.TimeFormat` is changed. I think it's best to duplicate the time format in the tests.

data/types/base.go
```diff
@@ -68,6 +69,26 @@ type Base struct {
 	VersionInternal   int                           `json:"-" bson:"_version,omitempty"`
 }
 
+// custom DeviceTime formatting
+type DeviceTime time.Time
+
+func (t DeviceTime) MarshalJSON() ([]byte, error) {
```
toddkazakov, line 75: You can implement the [TextMarshaller](https://pkg.go.dev/encoding#TextMarshaler) interface here:
```
type DeviceTime time.Time

func (d DeviceTime) MarshalText() (text []byte, err error) {
	return []byte(time.Time(d).Format("2006-January-02")), nil
}
```

data/types/base_test.go
```diff
@@ -972,13 +940,6 @@ var _ = Describe("Base", func() {
 				Expect(identityFields).To(BeEmpty())
 			})
 
-			It("returns error if time is empty", func() {
```
toddkazakov, line 975: Should you test the `(*b.Time).IsZero()` check here?

data/types/base_test.go
```diff
@@ -989,7 +950,7 @@ var _ = Describe("Base", func() {
 			It("returns the expected identity fields", func() {
 				identityFields, err := datum.IdentityFields()
 				Expect(err).ToNot(HaveOccurred())
-				Expect(identityFields).To(Equal([]string{*datum.UserID, *datum.DeviceID, *datum.Time, datum.Type}))
+				Expect(identityFields).To(Equal([]string{*datum.UserID, *datum.DeviceID, (*datum.Time).Format(types.TimeFormat), datum.Type}))
```
toddkazakov, line 953: I think you should duplicate the time format in the tests to make sure we have a failing test when the format is changed.

data/types/blood/blood_test.go
```diff
@@ -123,7 +123,7 @@ var _ = Describe("Blood", func() {
 			It("returns the expected identity fields", func() {
 				identityFields, err := datum.IdentityFields()
 				Expect(err).ToNot(HaveOccurred())
-				Expect(identityFields).To(Equal([]string{*datum.UserID, *datum.DeviceID, *datum.Time, datum.Type, *datum.Units, strconv.FormatFloat(*datum.Value, 'f', -1, 64)}))
+				Expect(identityFields).To(Equal([]string{*datum.UserID, *datum.DeviceID, (*datum.Time).Format(types.TimeFormat), datum.Type, *datum.Units, strconv.FormatFloat(*datum.Value, 'f', -1, 64)}))
```
toddkazakov, line 126: https://github.com/tidepool-org/platform/pull/576/files#r712068293

data/types/bolus/bolus_test.go
```diff
@@ -170,7 +171,7 @@ var _ = Describe("Bolus", func() {
 			It("returns the expected identity fields", func() {
 				identityFields, err := datum.IdentityFields()
 				Expect(err).ToNot(HaveOccurred())
-				Expect(identityFields).To(Equal([]string{*datum.UserID, *datum.DeviceID, *datum.Time, datum.Type, datum.SubType}))
+				Expect(identityFields).To(Equal([]string{*datum.UserID, *datum.DeviceID, (*datum.Time).Format(types.TimeFormat), datum.Type, datum.SubType}))
```
toddkazakov, line 174: https://github.com/tidepool-org/platform/pull/576/files#r712068293

data/types/device/device_test.go
```diff
@@ -119,7 +120,7 @@ var _ = Describe("Device", func() {
 			It("returns the expected identity fields", func() {
 				identityFields, err := datum.IdentityFields()
 				Expect(err).ToNot(HaveOccurred())
-				Expect(identityFields).To(Equal([]string{*datum.UserID, *datum.DeviceID, *datum.Time, datum.Type, datum.SubType}))
+				Expect(identityFields).To(Equal([]string{*datum.UserID, *datum.DeviceID, (*datum.Time).Format(types.TimeFormat), datum.Type, datum.SubType}))
```
toddkazakov, line 123: https://github.com/tidepool-org/platform/pull/576/files#r712068293

data/types/upload/upload_test.go
```diff
@@ -425,7 +420,7 @@ var _ = Describe("Upload", func() {
 					errorsTest.WithPointerSourceAndMeta(structureValidator.ErrorValueNotEqualTo("invalidType", "upload"), "/type", &dataTypes.Meta{Type: "invalidType"}),
 					errorsTest.WithPointerSourceAndMeta(structureValidator.ErrorValueNotExists(), "/client/name", &dataTypes.Meta{Type: "invalidType"}),
 					errorsTest.WithPointerSourceAndMeta(structureValidator.ErrorValueNotExists(), "/client/version", &dataTypes.Meta{Type: "invalidType"}),
-					errorsTest.WithPointerSourceAndMeta(structureValidator.ErrorValueStringAsTimeNotValid("invalid", "2006-01-02T15:04:05"), "/computerTime", &dataTypes.Meta{Type: "invalidType"}),
+					//errorsTest.WithPointerSourceAndMeta(structureValidator.ErrorValueStringAsTimeNotValid("invalid", "2006-01-02T15:04:05"), "/computerTime", &dataTypes.Meta{Type: "invalidType"}),
```
toddkazakov, line 423: Remove this test case if it's not valid anymore

toddkazakov: I believe the parser is now responsible for returning errors if the format is not valid. You should probably move this test to the parser instead of removing it.

data/types/base.go
```diff
@@ -68,6 +69,26 @@ type Base struct {
...
+	b := make([]byte, 0, len(DeviceTimeFormat)+2)
+	b = append(b, '"')
+	b = time.Time(t).AppendFormat(b, DeviceTimeFormat)
+	b = append(b, '"')
+	return b, nil
+}
+func (b *Base) MarshalJSON() ([]byte, error) {
```
toddkazakov, line 82: Can you please add a test?

data/types/base_test.go
Roukoswarf: good catch, re-added this test.

data/types/upload/upload_test.go
Roukoswarf: parser already has tests for string formats, and invalid formats, added test for Time format now to cover this.

data/types/base.go
```diff
@@ -197,11 +199,10 @@ func (b *Base) Validate(validator structure.Validator) {
 		stringValidator.Exists().NotEmpty().LengthLessThanOrEqualTo(TagLengthMaximum)
 	}).EachUnique()
 
-	timeValidator := validator.String("time", b.Time)
+	timeValidator := validator.Time("time", b.Time)
 	if b.Type != "upload" { // HACK: Need to replace upload.Upload with data.DataSet
-		timeValidator.Exists()
+		timeValidator.Exists().NotZero()
```
toddkazakov, line 204: Is there a reason there is a `NotZero()` condition here? The time field is pointer to time, thus it's empty value is `nil`.

data/store/mongo/mongo_test.go
```diff
@@ -279,12 +301,48 @@ var _ = Describe("Mongo", func() {
...
+					var result *upload.Upload
+					var createdTime time.Time
+					var modifiedTime time.Time
+					var deletedTime time.Time
+					var recordTime time.Time
+
+					BeforeEach(func() {
+						legacyUpload = NewLegacyDataSet(userID, deviceID)
+						createdTime, _ = time.Parse(time.RFC3339, "2016-09-01T11:00:00Z")
+						modifiedTime, _ = time.Parse(time.RFC3339, "2016-10-01T11:00:00Z")
```
toddkazakov, line 315: The source string doesn't have nanos, which is what we would store in the database.

Instead of parsing a static string time, you can just generate a random time with the test utils, e.g.: `test.PastFarTime()`.

data/types/base.go
```diff
@@ -248,14 +249,14 @@ func (b *Base) IdentityFields() ([]string, error) {
 	if b.Time == nil {
 		return nil, errors.New("time is missing")
 	}
-	if *b.Time == "" {
+	if (*b.Time).IsZero() {
```
toddkazakov, line 252: I'm not entirely sure checking for `IsZero()` is necessary (or even semantically correct). The empty value of `b.Time` is `nil`, not a zero time.

toddkazakov (changes requested):
Almost there!

data/store/mongo/mongo.go
```diff
@@ -635,9 +635,9 @@ func (d *DataRepository) UnarchiveDeviceDataUsingHashesFromDataSet(ctx context.C
 
 	result := struct {
 		ID struct {
-			Active            bool   `bson:"_active"`
-			ArchivedDataSetID string `bson:"archivedDatasetId"`
-			ArchivedTime      string `bson:"archivedTime"`
+			Active            bool       `bson:"_active"`
+			ArchivedDataSetID *string    `bson:"archivedDatasetId"`
+			ArchivedTime      *time.Time `bson:"archivedTime"`
```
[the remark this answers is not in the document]
toddkazakov: Now that I read the code again it might not even be possible to have the record in this state. This is the code before the changes:
```golang
ArchivedTime string `bson:"archivedTime"`

...

if result.ID.Active != (result.ID.ArchivedDataSetID == "") || result.ID.Active != (result.ID.ArchivedTime == "") {
   continue
}

if result.ID.Active {
...
} else {
	set["archivedDatasetId"] = result.ID.ArchivedDataSetID
	set["archivedTime"] = result.ID.ArchivedTime
}
```

data/types/base_test.go
```diff
@@ -69,22 +71,6 @@ var _ = Describe("Base", func() {
 		})
 	})
 
-	Context("with new datum", func() {
```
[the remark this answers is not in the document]
Roukoswarf: accidental, re-used its section for another test, then deleted its replacement, deleting the whole thing instead of a part of it, reverting.

data/types/base_test.go
```diff
@@ -380,10 +351,6 @@ var _ = Describe("Base", func() {
 					func(datum *types.Base) { datum.DeviceID = pointer.FromString(dataTest.NewDeviceID()) },
 					structure.Origins(),
 				),
-				Entry("device time missing",
```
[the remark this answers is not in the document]
Roukoswarf: it was made invalid during the change to time.Time DeviceTime, when going back to string I re-added any invalidated tests, but missed this one.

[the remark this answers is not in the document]
Roukoswarf: nevermind, it was never invalid, the time invalid below was, not sure how this happened.

data/store/mongo/mongo.go
```diff
@@ -671,8 +671,8 @@ func (d *DataRepository) UnarchiveDeviceDataUsingHashesFromDataSet(ctx context.C
 			unset["archivedDatasetId"] = true
 			unset["archivedTime"] = true
 		} else {
-			set["archivedDatasetId"] = result.ID.ArchivedDataSetID
-			set["archivedTime"] = result.ID.ArchivedTime
+			set["archivedDatasetId"] = *result.ID.ArchivedDataSetID
```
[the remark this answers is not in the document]
toddkazakov: Probably not true - see https://github.com/tidepool-org/platform/pull/576#discussion_r720296972

data/types/base.go
toddkazakov: It's ok to leave as it is, because the original code had the check for an empty string

data/types/base.go
Roukoswarf: i started with not-pointers, corrected.

toddkazakov: I just saw that you had to remove `timeValidator.AsTime(TimeFormat)` which is roughly equivalent to the `NotZero()` check here.

data/store/mongo/mongo.go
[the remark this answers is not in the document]
Roukoswarf: I want to read the code history on this a bit more, but at this point it seems like it is an impossible state, and hopefully this never becomes a problem.

Though, as part of the migration, old records should be checked for empty dates, in the case that there was a previously acceptable state where empty dates were possible.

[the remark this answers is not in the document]
Roukoswarf: History shows no changes in logic that should affect this, it should™ be fine, but still worth checking during migration.

toddkazakov (approved):
LGTM :shipit:

PR #583, opened 2021-12-16
[BACK-1997] Send prescription access code emails when a prescription is submitted

toddkazakov:
Sends a templated email using the kafka mailer when a prescription is put in a submitted state. Stacked PR to make reviewing/merging easier.


Roukoswarf (approved):
looks good

PR #584, opened 2021-12-21
simplify mongo command aliases, add comment on username format

Roukoswarf:
these new aliases should be slightly more readable, removing much of the sub-shelling and logic.


toddkazakov (approved):
LGTM :shipit:

PR #588, opened 2022-01-17
[BACK-1903] Support Tandem CIQ

darinkrauss:
- https://tidepool.atlassian.net/browse/BACK-1903
- Add bolus/automated
- Allow basal/automated to be suppressed by basal/temp
- Add suppressed to basal/automated
- Limit length of schedule name
- Update parsing and validation errors for suppressed Basals
- Fix validation for pumpSettingsOverride expected duration
- Update tests

### DIFF
--- a/data/types/basal/temporary/temporary.go
+++ b/data/types/basal/temporary/temporary.go
@@ -3,6 +3,7 @@ package temporary
 import (
 	"github.com/tidepool-org/platform/data"
 	"github.com/tidepool-org/platform/data/types/basal"
+	dataTypesBasalAutomated "github.com/tidepool-org/platform/data/types/basal/automated"
 	dataTypesBasalScheduled "github.com/tidepool-org/platform/data/types/basal/scheduled"
 	"github.com/tidepool-org/platform/data/types/insulin"
 	"github.com/tidepool-org/platform/metadata"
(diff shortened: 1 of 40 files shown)


data/types/basal/temporary/temporary.go
```diff
@@ -182,10 +186,12 @@ func parseSuppressed(parser structure.ObjectParser) Suppressed {
 func validateSuppressed(validator structure.Validator, suppressed Suppressed) {
 	if suppressed != nil {
 		switch suppressed := suppressed.(type) {
+		case *dataTypesBasalAutomated.SuppressedAutomated:
```
toddkazakov, line 189: Both case statements can be merged since they use the same validator, e.g.:
```golang
switch suppressed := suppressed.(type) {
case *dataTypesBasalAutomated.SuppressedAutomated,
     *dataTypesBasalScheduled.SuppressedScheduled:
         suppressed.Validate(validator)
}
```

toddkazakov (approved):
Looks good to me

data/types/basal/temporary/temporary.go
darinkrauss: True. I think I'd like to keep it as-is so the `validateSuppressed` function follows the same form as the `parseSuppressed` function just above it (which cannot combine case statements).

darinkrauss:
Closing in favor of PR against the `master` branch.

PR #597, opened 2022-03-30
[BACK-2029] Add Dexcom G6 Pro to the list of devices we can import through Dexcom Clarity

tjotala:
This adds "g6 pro" to the list of supported Dexcom CGM devices, so that data from those devices can be imported via Dexcom Clarity.


Roukoswarf (approved):
LGTM

toddkazakov (approved):

PR #600, opened 2022-06-08
Summary service tuning for live data

Roukoswarf:
Current sticking points fixed by this PR:
1. varying precision of date strings from different data sources can result in "past" data being pulled into a query due to db string comparison
2. backfill checks summary table first, resulting in unique key errors if a user uploads new data in the lengthy gap between the distinct summary query and insert, as getting distinct deviceData cgm users takes significant time.
3. summary will calculate far-future data, resulting in data which can be made to not match tidepool web

### DIFF
--- a/data/store/mongo/mongo_data.go
+++ b/data/store/mongo/mongo_data.go
@@ -942,8 +942,9 @@ func (d *DataRepository) GetCGMDataRange(ctx context.Context, id string, startTi
 }
 
 func (d *DataRepository) GetLastUpdatedForUser(ctx context.Context, id string) (*summary.UserLastUpdated, error) {
+	var err error
+	var cursor *mongo.Cursor
 	var status summary.UserLastUpdated
-	var dataSetOld []*continuous.Continuous
 	var dataSet []*continuous.Continuous
 
 	if ctx == nil {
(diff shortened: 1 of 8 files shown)


data/store/mongo/mongo_data.go
```diff
@@ -949,43 +954,67 @@ func (d *DataRepository) GetLastUpdatedForUser(ctx context.Context, id string) (
...
+		"time": bson.M{"$lte": futureCutoff.Format(time.RFC3339Nano),
+			"$gte": pastCutoff.Format(time.RFC3339Nano)},
+	}
+
 	selector := bson.M{
 		"_active": true,
 		"_userId": id,
 		"type":    "cbg",
+		"time": bson.M{"$lte": futureCutoff,
+			"$gte": pastCutoff.Format(time.RFC3339Nano)},
```
tjotala, line 973: Shouldn't both of these be dates (i.e. not strings)?

tjotala (approved):
Couple of comments, but not necessarily blocking.

data/store/mongo/mongo_data.go
Roukoswarf: good catch, this would be a weird bug to track down after date migration.

tjotala (approved):
LGTM

PR #602, opened 2022-06-16
Fix quick deletion of new data causing summary update error

Roukoswarf:
fix an edge case where a user uploads data, but quickly deletes it before summary can be calculated


tjotala (approved):
LGTM

PR #603, opened 2022-06-16
Fix GetSummary returning 500 instead of 404

Roukoswarf:
Stop GetSummary from returning a 500 error when it should be returning 404 for summary not found


tjotala (approved):
LGTM

PR #605, opened 2022-06-22
Fix task claim race condition

Roukoswarf:
Multiple workers can currently claim a single task and run it at the same time, this should fix that.


tjotala (approved):
LGTM :shipit:

tjotala (approved):
It'd be great if there was a test that specifically tries to update two tasks just to see that the 2nd update fails with modifieCount == 0 as expected.

Other than that LGTM.

PR #606, opened 2022-06-30
Summary calculation of overlapping and deleted+added data

Roukoswarf:
There are some workflows that add+remove data at the same time, along with some cases where duplicate data may not be marked inactive.

Many of these cases also lack a DeviceID (in reality, most do).

This PR removes DeviceID handling, replacing it with skipping of nearby records, and prevention of data deletion causing summary calculation to go back in time slightly and double calculate values.


Roukoswarf:
Superseded by #609

PR #615, opened 2022-12-13
[BACK-2290] Dexcom patient connection

clintonium-119:
[BACK-2290]

- Updates oauth redirect logic and html to include custodial signup key and email when available
- Adds hydrophone client to auth service

[BACK-2290]: https://tidepool.atlassian.net/browse/BACK-2290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


auth/service/api/v1/oauth.go
```diff
@@ -104,20 +107,48 @@ func (r *Router) OAuthProviderRedirectGet(res rest.ResponseWriter, req *rest.Req
...
+	// Include custodial account signup credentials in redirect URL query, if applicable
+	confirmations, err := r.ConfirmationClient().GetAccountSignupConfirmationWithResponse(ctx, confirmationClient.UserId(restrictedToken.UserID))
+	if err != nil {
+		r.htmlOnError(res, req, err)
+		return
+	}
+
+	signupParams := url.Values{}
+	if confirmations.JSON200 != nil && len(confirmationClient.ConfirmationList(*confirmations.JSON200)) > 0 {
+		confirmation := confirmationClient.ConfirmationList(*confirmations.JSON200)[0]
```
toddkazakov, line 129: Do you need to check the confirmation status here? Can `hydrophone` return sign up confirmations that are already completed?

toddkazakov (dismissed):
Looks great! Left an inline question. Approving in case you think there's no need to address it.

auth/service/api/v1/oauth.go
clintonium-119: I think this is OK, since hydrophone filters the query for `pending` status:
https://github.com/tidepool-org/hydrophone/blob/e398604f78b07f2c154b4899b379f762af7d81da/api/signup.go#L426

toddkazakov (approved):

clintonium-119:
I should call out that I had to comment out one of the tests in order to pass.
https://github.com/tidepool-org/platform/pull/615/commits/772cae45d1883ca6f23b11902dc8c26b41a6f0a4

There are also a few tests around db connections that fail locally because my local mongo instance allows connection to IPs that the unit test fails on.  These tests seem a bit strange to me, and I'm thinking they could well be done away with.

toddkazakov (approved):
LGTM

clintonium-119:
@toddkazakov I've updated the packages throwing snyk warnings, and have disabled the problematic `golint` step in the CI build.

auth/service/api/v1/oauth.go
```diff
@@ -132,8 +165,14 @@ func (r *Router) OAuthProviderRedirectGet(res rest.ResponseWriter, req *rest.Req
...
 		return
 	} else if len(providerSessions) > 0 {
-		r.htmlOnError(res, req, errors.Newf("provider session already exists for user, type, and name"), alreadyConnectedError)
-		return
+		// Delete existing provider sessions and tasks if matching name and type found for user.
+		// This operation will also reset the data source to a `disconnected` state, and remove any associated tasks
+		// A new provider session and task will be created below which will update the existing data source state to `connected`.
+		for _, session := range providerSessions {
+			if deleteSessionErr := r.AuthClient().DeleteProviderSession(ctx, session.ID); deleteSessionErr != nil {
+				r.htmlOnError(res, req, errors.Newf("could not remove existing provider session"), alreadyConnectedError)
```
lostlevels, line 173: Correct me if I'm wrong, but this will send multiple html documents as one invalid doc if it fails to delete multiple providers right?
e.g.,
```
<html> ... could not remove existing provider session </html>
<html> ... could not remove existing provider session </html>
```

Perhaps you could buffer the error messages then send only one doc or just send one doc with that error unless there will always only be one provider.

Also does this not return after the error?

clintonium-119: Good catch.  I've added a return after the delete session error which should only allow it to fire once.

PR #617, opened 2023-01-02
[BACK-2163] API for Generating Challenge Values for App Attestation or App Assertion

lostlevels:
This commit adds app validation routes that will generate a challenge for the user to use in verification. App Validation is the process of determining whether an app is a valid instance of a mobile app. This involves two parts. Attestation to confirm an app is a valid instance and Assertion for the app to assert legitimacy in exchange for sensitive or private information.


toddkazakov (dismissed):
Looks great!

toddkazakov (dismissed):

auth/service/api/v1/appvalidate.go
```diff
@@ -120,12 +124,20 @@ func (r *Router) VerifyAssertion(res rest.ResponseWriter, req *rest.Request) {
 		return
 	}
 
+	logFields := log.Fields{
+		"userID": details.UserID(),
+		"keyId":  assertVerify.KeyID,
+	}
+
+	// log debug fields (only in qa environments)
```
toddkazakov, line 132: Should this be removed?

toddkazakov (approved):

appvalidate/coastal_secrets.go
```diff
@@ -0,0 +1,210 @@
...
+	b, err := json.Marshal(internalPayload)
+	if err != nil {
+		return nil, fmt.Errorf("Coastal marshal error: %w", err)
+	}
+
+	u, err := url.Parse(c.Config.BaseURL)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse Coastal API baseURL: %w", err)
+	}
+	u.Path = path.Join(u.Path, "devices/api/v1/certificates")
```
ewollesen, line 162: This could be done once in `newCoastalSecretsConfig` instead, and just stored as a string there.

appvalidate/assertion.go
```diff
@@ -0,0 +1,73 @@
...
+	AssertionCounter uint32    `bson:"assertionCounter,omitempty"`
+}
+
+type AssertionResponse struct {
+	Data any `json:"data"`
+}
+
+type AssertionClientData struct {
+	Challenge   string          `json:"challenge"`
+	Partner     string          `json:"partner"`     // Which partner are we requesting a secret from - currently only one supported
```
ewollesen, line 38: Is this comment accurate? I believe there are two?

appvalidate/partner_secrets.go
```diff
@@ -0,0 +1,59 @@
...
+	ErrEmptySecretsConfig        = errors.New("empty secrets config")
+	ErrInvalidPartnerPayload     = errors.New("invalid partner payload")
+)
+
+type PartnerSecrets struct {
+	cs *CoastalSecrets
+	ps *PalmTreeSecrets
+}
+
+// NewPartnerSecrets creates a new Secrets wrapper with all supported
```
ewollesen, line 22: The callers should be checking for nil, and if found, they should just continue or fail. Using a half-baked secrets isn't going to work anyway. 

As far as I can tell, the callers do just that:

https://github.com/tidepool-org/platform/blob/master/auth/service/api/v1/appvalidate.go#L153

I don't think we should be returning a half-initialized PartnerSecrets. 

If the callers really need to have something non-nil (but as far as I can tell, they don't), then a dummy implementation should be returned that just returns dummy responses and maybe logs to remind people that it's a dummy that's in use.

appvalidate/palmtree_secrets.go
```diff
@@ -0,0 +1,181 @@
...
+	KeyData []byte `envconfig:"PALMTREE_TLS_KEY_DATA"`
+}
+
+type PalmTreeSecrets struct {
+	Config PalmTreeSecretsConfig
+	client *http.Client
+}
+
+func NewPalmTreeSecrets() (*PalmTreeSecrets, error) {
+	cfg, err := newPalmTreeSecretsConfig()
```
ewollesen, line 49: if you were to pass the cfg as an argument, this function wouldn't even be necessary, but if you wanted to keep it anyway (sometimes its nice to have a constuctor), it'd be simpler,as it wouldn't return an error. Plus then tests would be easier to write, as the config provided could be easily faked/mocked/whatever.

Why aren't there any tests?

appvalidate/palmtree_secrets.go
```diff
@@ -0,0 +1,181 @@
...
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(payload); err != nil {
+		return nil, err
+	}
+
+	u, err := url.Parse(pt.Config.BaseURL)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse PalmTree API baseURL: %w", err)
+	}
+	u.Path = path.Join(u.Path, fmt.Sprintf("v1/certificate-authorities/%s/enrollments", url.PathEscape(pt.Config.CalID)))
```
ewollesen, line 140: As above, a minor thing, but this could be done once at initialization.

auth/service/service/service.go
```diff
@@ -56,6 +55,10 @@ type Service struct {
 	authClient         *Client
 	userEventsHandler  events.Runner
 	deviceCheck        apple.DeviceCheck
+	appValidator       *appvalidate.Validator
```
ewollesen, line 58: I feel like these should be interfaces so they can be easily swapped out for tests.

ewollesen (changes requested):
The Makefile change causes a breakage, so that should be fixed.

Otherwise, nothing I have is a deal breaker, but I feel like areas of this could really use more love (or at least more tests).

Also, I don't think we should take a dependency on https://github.com/bas-d/appattest

We barely even use it. Could we strip out the parts we want/need (with attribution of course, it's licensed BSD 3-clause) and not have to worry about having it around, or its dependency (that we don't use at all). Or just write those few bits ourselves? We're really just barely using it as far as I can see.

auth/service/service/service.go
lostlevels: Rational behind these is the Validator itself contains only interfaces.

ewollesen: Ah, fair enough then.

lostlevels:
> The Makefile change causes a breakage, so that should be fixed.
> 
> Otherwise, nothing I have is a deal breaker, but I feel like areas of this could really use more love (or at least more tests).
> 
> Also, I don't think we should take a dependency on https://github.com/bas-d/appattest
> 
> We barely even use it. Could we strip out the parts we want/need (with attribution of course, it's licensed BSD 3-clause) and not have to worry about having it around, or its dependency (that we don't use at all). Or just write those few bits ourselves? We're really just barely using it as far as I can see.

The functionality we use from `bas-d/appattest` is worth it IMO. The validation is *not* trivial - see:
https://developer.apple.com/documentation/devicecheck/validating-apps-that-connect-to-your-server

appvalidate/palmtree_secrets.go
lostlevels: Not sure the value of tests here as it's only calling out to external tests. The only other thing would be validating the certificates but those are also just standard library calls. Any tests in particular you have in mind?

ewollesen:
> > The Makefile change causes a breakage, so that should be fixed.
> > Otherwise, nothing I have is a deal breaker, but I feel like areas of this could really use more love (or at least more tests).
> > Also, I don't think we should take a dependency on https://github.com/bas-d/appattest
> > We barely even use it. Could we strip out the parts we want/need (with attribution of course, it's licensed BSD 3-clause) and not have to worry about having it around, or its dependency (that we don't use at all). Or just write those few bits ourselves? We're really just barely using it as far as I can see.
> 
> We definitely use the functionality from the appattest library - see the `.Verify` calls for [assertions](https://github.com/tidepool-org/platform/blob/BACK-2163-generate-api-challenge/appvalidate/validator.go#L210) and [attestations](https://github.com/tidepool-org/platform/blob/BACK-2163-generate-api-challenge/appvalidate/validator.go#L155). The validation is _not_ trivial - see: https://developer.apple.com/documentation/devicecheck/validating-apps-that-connect-to-your-server

I see. You're right, we definitely make more use of it than I had seen. Seems like a reasonable tradeoff then.

appvalidate/palmtree_secrets.go
ewollesen: Tests help other people understand what the code is doing. They can detect regressions, and generally improve the design of code. 

It'd be interesting to see some tests around `b64StdEncodingToURLEncoding`, because I'm not sure it's doing what it's supposed to, a test would give me confidence that it's doing the right things.

The code that handles continuing with invalid partner secrets would be another place that it would be good to have confidence that it can continue to function as expected without validated secret data.

ewollesen (approved):
Would prefer more test coverage, but I don't see anything dangerous. Thanks for fixing the Makefile.

PR #618, opened 2023-01-04
[BACK-2164] Add attestation verification route.

lostlevels:
Add attestation verification route under POST /v1/attestations/verifications.

This commit continues the attestation API to include verification of an
attestation. Once a user generates an attestation challenge, they must verify
the attestation object received from Apple which includes this challenge. Once
verified, the user will be able to make assertions in exchange for a secret.


toddkazakov (changes requested):
As discussed, we'd need some info level logging with the userId and keyId of the attestation/verification, because those parameters are not present in the API routes

toddkazakov:
@lostlevels I don't think we run the services with debug level logging in production, so it's best to log at `info` level.

toddkazakov (approved):

PR #619, opened 2023-01-05
[BACK-2290] Decline dexcom auth in clinic patient records

clintonium-119:
See [BACK-2290]

Also adds a clinic client to the auth service.

[BACK-2290]: https://tidepool.atlassian.net/browse/BACK-2290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/auth/service/api/v1/oauth.go
+++ b/auth/service/api/v1/oauth.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ant0ine/go-json-rest/rest"
 
+	clinicClient "github.com/tidepool-org/clinic/client"
 	confirmationClient "github.com/tidepool-org/hydrophone/client"
 
 	"github.com/tidepool-org/platform/auth"
@@ -151,6 +152,19 @@ func (r *Router) OAuthProviderRedirectGet(res rest.ResponseWriter, req *rest.Req
 	if errorCode := query.Get("error"); errorCode == oauth.ErrorAccessDenied {
 		html := fmt.Sprintf(htmlOnRedirect, redirectURLDeclined.String())
 		r.htmlOnRedirect(res, req, html)
+
+		if prvdr.Name() == "dexcom" {
+			// Decline dexcom request for any associated patient records
+			response, err := r.ClinicClient().DeclineDexcomConnectRequestWithResponse(ctx, clinicClient.UserId(restrictedToken.UserID))
+
+			if err != nil && response.StatusCode() != http.StatusNotFound {
+				log.LoggerFromContext(ctx).WithError(err).Error("unexpected error declining dexcom connect request")
+			}
+			if !(response.StatusCode() == http.StatusNoContent || response.StatusCode() == http.StatusNotFound) {
+				log.LoggerFromContext(ctx).WithError(err).Errorf("unexpected status code when declining dexcom connect request %v", response.StatusCode())
+			}
+		}
+
 		return
 	} else if errorCode != "" {
 		r.htmlOnError(res, req, errors.Newf("oauth provider return unexpected error %q", errorCode))
--- a/auth/service/service.go
+++ b/auth/service/service.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/tidepool-org/platform/apple"
 
+	clinicClient "github.com/tidepool-org/clinic/client"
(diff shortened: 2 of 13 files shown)


toddkazakov:
@clintonium-119 Handling errors in an OAuth2 flow seems off to me. The user is not actually rejecting the Dexcom connect invite we sent, but Dexcom's terms and conditions and/or granting the permissions. I think we should discuss this further. 

Also, I don't think we should add a dependency in the auth service to the clinic service. It seems strange for the auth service to know anything about clincs and their patients. A better approach would be to reject the hydrophone invite, which then updates the status in the clinic service.

clintonium-119:
@toddkazakov We can't just reject the hydrophone invite, and then update the status in the clinic service, since the user may choose to decline the dexcom authorization, and then proceed to accept the hydrophone invite.

I agree that dropping the dependancy is good, but the only way I can see of doing that is to make a separate call to the clinic service from the front-end redirect endpoint, and set the decline time on page load. 

I think there may be issues, though - the user is not authenticated at that point, and even if we could do it somehow by passing some appropriate url params to authorize the proper action (like we do for the signup confirmation key), it would make that call each time the url was refreshed.

Perhaps if we chat we can figure out a more elegant way to drop the dependancy here.  I am a bit confused by you referring to this as handling errors in the Oauth flow, though.

We're just trying to record the time that the user denies access the actions leading to this response (per the API docs)

```
If the user denies access in Step Two, the redirect_uri will be appended with ?error=access_denied instead of the authorization code. 
```
I agree that it would be good if we can avoid dexcom-specific logic in the oauth redirect handler, but I'm not sure of where else we can react to this event.

Thoughts?

toddkazakov:
According to the [OAuth2 spec](https://www.rfc-editor.org/rfc/rfc6749), `access_denied` can be returned not only when the user declines the authorization, but also when the authorization server (in our case Dexcom) denies the request for other reasons (not sure if it's possible). I guess I was just surprised to see the client acting on the error code returned in the response.

Is there a more elegant way to handle the rejection using CDC? E.g. by creating a dexcom data source/provider session and handling this in clinic worker? Do we to update the clinic service when a dexcom session expires?

clintonium-119:
Originally, I hadn't wanted to handle it there either, but couldn't think of another place to do it.

We do already apply a users `data_sources` updates via CDC to patients, but the issue is that in the case at hand, for new connection requests, we don't actually have an object in the `data_sources` collection at this point.  We only create that when the user authorizes a dexcom connect request (or they initiate it themselves from their profile page)

Even if we did update the `data_sources` documents to allow a `pending` state, we'd still need to update it to `declined` after the oauth redirect in order to trigger the CDC event, and so we'd still have a dependancy (between auth and data)

toddkazakov:
> so we'd still have a dependancy (between auth and data)

Don't we already have a dependency between auth and data?

clintonium-119:
I don't see the `auth` service depending on `data`.  Perhaps I'm missing it.  The `data` service does of course depend on `auth`

toddkazakov:
@clintonium-119 the data source is created by the Dexcom provider in `dexcom/provider/provider.go#OnCreate`. I'd prefer to use this mechanism to indirectly update the clinic service if it's feasible. I'm not very familiar with the code, so there might be caveats.

clintonium-119:
I'm not too familiar with it either, but I'll have a look. 

My original intent was not to extend the data_sources objects, simply because I didn't want to risk introducing any new bugs or wrinkles.

Just to make sure this is what you're requsting:

I'll have to add `pending` and `declined` states to the data source model (currently, we just manage `connected`, `disconnected`, and `error` states), and create a new pending data source document for the user when a patient dexcom connection request occurs.

Then, when the user declines dexcom authorization, I'll use the dexcom provider mechanism instead of the clinic api (from within the oauth redirect handler) to update that data source from `pending` to `declined`.

At that point, the CDC can listen for newly-declined data sources and update the patient records.

I can't think of any bugs that I'd need to watch out for as I introduce new states to the data sources, but I'll need to dive deeper into the code and have a look for all existing handling of the current states.

PR #620, opened 2023-01-12
[BACK-2165] Add app assertion verification route.

lostlevels:
This commit adds app assertion verification. The user can get an assertion verified only after being attested and only after generating an assertion challenge. Once the assertion is verified, the user will be granted access to some private or sensitive information.

### DIFF
--- a/appvalidate/appvalidate.go
+++ b/appvalidate/appvalidate.go
@@ -14,7 +14,7 @@ import (
 //go:generate mockgen -build_flags=--mod=mod -destination=./mock.go -package=appvalidate github.com/tidepool-org/platform/appvalidate Repository,ChallengeGenerator
 
 var (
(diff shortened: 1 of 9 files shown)


toddkazakov:
@lostlevels I think the assertion ought to be passed to and verified when we request the secret values from the 3rd party provider. Do we even a separate endpoint?

lostlevels:
I believe so. The idea behind my implementation was that during assertion verification the secret is also fetched, its just not implemented yet, though the route itself doesn't really indicate it. Should the route just be renamed to something along the lines of /something/secret and the assertion is passed in?

toddkazakov:
@lostlevels I just wanted to make sure I understand what's going on. We can leave the route as is and change it when we implement the secret handling.

toddkazakov (approved):

PR #622, opened 2023-02-03
[BACK-2378][BACK-2388] Bump mongo-driver package to latest version 1.11 to support mongodb versions up to 6.0.

lostlevels:


toddkazakov (approved):
I saw you bumped Go to 1.19 in one of the other repositories. Should we bump it here is well or is this not required? If not, then this PR LGTM.

lostlevels:
> I saw you bumped Go to 1.19 in one of the other repositories. Should we bump it here is well or is this not required? If not, then this PR LGTM.

No, I only bumped that one because the latest mongodb driver requires at least Go 1.18 for the any keyword and since I was updating the Go version anyways I thought I would bump it up to the latest(ish).

toddkazakov (dismissed):

toddkazakov (approved):

PR #625, opened 2023-03-01
[BACK-2328] dexcom validation updates

jh-bate:
https://tidepool.atlassian.net/browse/BACK-2328

- remove validation for dexcom device ID if its not sent
- allow AlertSchedules to be empty
- remove dead code


tjotala (approved):
LGTM

PR #626, opened 2023-03-01
[BACK-2327] allow g7 uploads

jh-bate:
also see #625

- include g7 in the allowed list


tjotala:
LGTM but let's hold off on this until https://github.com/tidepool-org/platform/pull/625 goes first, so we can confirm what is the Dexcom G7 device identifier. It is likely to be "g7" or something similar.

lostlevels (approved):

jh-bate:
superseded by #632

PR #629, opened 2023-03-08
[BACK-1758] Move data with type=upload to its own collection with dual read/writes.

lostlevels:
Move deviceData with type='upload' to its own collection named deviceDataSets

which acts as a collection for "parent" datum where its children are in the original deviceData collection. Split platform/data/store.DataRepository into 2 repositories, one for DataSet and one for data of these DataSets called DataSetRepository and DatumRepository, respectively.

### DIFF
--- a/data/store/mongo/mongo_data.go
+++ b/data/store/mongo/mongo_data.go
@@ -2,15 +2,14 @@ package mongo
 
 import (
 	"context"
+	"strings"
 	"time"
-
-	baseDatum "github.com/tidepool-org/platform/data/types"
-
-	"github.com/tidepool-org/platform/data/summary/types"
 
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.mongodb.org/mongo-driver/mongo/readconcern"
+	"go.mongodb.org/mongo-driver/mongo/writeconcern"
 
 	"github.com/tidepool-org/platform/data"
 	"github.com/tidepool-org/platform/data/store"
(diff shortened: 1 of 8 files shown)


toddkazakov (commented):
@lostlevels ~~If I understand correctly, when we deploy this PR we'll only write to and read from the new collection. Let's say there's a request for fetching an upload record by id that's been created before those changes. The record will only be present in the old collection, thus the request will return 404. Shouldn't the calling functions of the repositories fall back to the old collection for all reads?~~

Ignore my stupid comment above. Github didn't expand the diff of mongo_data.go and I assumed it's only deletions. Reviewing now.

data/store/mongo/mongo_data.go
```diff
@@ -4,258 +4,145 @@ import (
...
+	dataSets, err := d.DataSetRepository.GetDataSetsForUserByID(ctx, userID, filter, pagination)
+	if err != nil {
+		return nil, err
 	}
-	if pagination == nil {
-		pagination = page.NewPagination()
-	} else if err := structureValidator.New().Validate(pagination); err != nil {
-		return nil, errors.Wrap(err, "pagination is invalid")
+	if len(dataSets) > 0 {
+		return dataSets, nil
```
toddkazakov, line 44: Instead of returning early shouldn't you merge results from both collections?

data/store/mongo/mongo_data.go
```diff
@@ -4,258 +4,145 @@ import (
...
+	// Try reading from both new and old collections that hold dataSets, starting with the new one.
+	// Can read only from the new dataSet collection when migration completed.
+	dataSets, err := d.DataSetRepository.ListUserDataSets(ctx, userID, filter, pagination)
+	if err != nil {
+		return nil, err
 	}
-	if !filter.Deleted {
-		selector["deletedTime"] = bson.M{"$exists": false}
+	if len(dataSets) > 0 {
+		return dataSets, nil
```
toddkazakov, line 57: Instead of returning early shouldn't you merge results from both collections?

toddkazakov (changes requested):
Everything looks good, except the two methods which ought to list all data sets for a given user. Currently, if there are any data sets in the new collection those methods return the results only from the new collection, ignoring all data in the old collection. I think results from both collections should be merged instead.

lostlevels:
Squashed commits because rebasing with master was getting funky even with rerere.

toddkazakov (approved):

PR #630, opened 2023-03-08
BACK-2328 allow for 12hr time format

jh-bate:
validation was failing with times such as `8:00 AM` and `9:00 PM` being sent

- allow for existing 24hr format
- including 12hr format (am or pm)
    - ignore case and allow `am` or `AM` or `aM`
    - allow extra spacing between time and postfix  `8:00 AM` or `8:00   AM`

### DIFF
--- a/dexcom/alert.go
+++ b/dexcom/alert.go
@@ -4,6 +4,8 @@ import (
 	"regexp"
 	"sort"
 	"strconv"
+	"strings"
+	"time"
 
 	dataTypesSettingsCgm "github.com/tidepool-org/platform/data/types/settings/cgm"
 	"github.com/tidepool-org/platform/errors"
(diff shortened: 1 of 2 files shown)


dexcom/alert.go
```diff
@@ -627,15 +629,17 @@ func (a *AlertSetting) validateUrgentLowSoon(validator structure.Validator) {
 }
 
 func ParseAlertScheduleSettingsTime(value string) (int, int, bool) {
-	if submatch := alertScheduleSettingsTimeExpression.FindStringSubmatch(value); len(submatch) != 3 {
+	timeFormat := "15:04"
```
tjotala, line 632: This could be handled simply by extending the original regex pattern below to accommodate the optional AM/PM indicator with zero or more spaces in between:

```
var alertScheduleSettingsTimeExpression = regexp.MustCompile("^([0-9]{1,2}):([0-9]{2})(?i:\\s*(AM|PM))?$")
```

plus the actual check would need to accommodate 3 or 4 matches:

```
if submatch := alert... ; len(submatch) >= 3 {
```

tjotala: I should note that that check won't catch non-sensical 12-hour clock values like "18:05 AM", but I think it's good enough for this use.

tjotala (changes requested):

dexcom/alert.go
jh-bate: I guess my assumption was the proposed solution covers both cases but also covers the fact we are actually dealing with times here by parsing them as times ... and will also cover the case you mention above of non sensical 12 hour times.

if regex we then also need to convert the 12hr into 24hr format for consistency 

thoughts?

tjotala: Ah, yea, fair point. Yea, let's stick with your parsing solution, even though we don't actually use the resulting time (sensical or not) for anything AFAIK.

tjotala (approved):

PR #632, opened 2023-03-22
[BACK-2433] dexcom api v3

jh-bate:
v3 updates:

- [calibrations api docs](https://developer.dexcom.com/docs/dexcomv3/operation/getCalibrationsV3)
- [events api docs](https://developer.dexcom.com/docs/dexcomv3/operation/getEventsV3)
- [devices api docs](https://developer.dexcom.com/docs/dexcomv3/operation/getDevicesV3)
- [egvs api docs](https://developer.dexcom.com/docs/dexcomv3/operation/getEstimatedGlucoseValuesV3) 

general:
  - more data validation tests 
  - rationalize some of the type tests


dexcom/fetch/runner.go
```diff
@@ -363,8 +363,9 @@ func (t *TaskRunner) updateDeviceHash(device *dexcom.Device) bool {
 		t.deviceHashes = map[string]string{}
 	}
 
-	if device.SerialNumber != nil && t.deviceHashes[*device.SerialNumber] != deviceHash {
-		t.deviceHashes[*device.SerialNumber] = deviceHash
+	// TODO: SerialNumber is no longer sent, use transmitterId instead?
```
jh-bate, line 366: seeking feedback on this

dexcom/device.go
```diff
@@ -19,10 +19,14 @@ const (
 	DeviceDisplayDeviceShareReceiver       = "shareReceiver"
 	DeviceDisplayDeviceTouchscreenReceiver = "touchscreenReceiver"
 
-	DeviceTransmitterGenerationG4    = "g4"
-	DeviceTransmitterGenerationG5    = "g5"
-	DeviceTransmitterGenerationG6    = "g6"
-	DeviceTransmitterGenerationG6Pro = "g6 pro"
```
jh-bate, line 25: no listed `g6 pro` as per dexcom spec ... but might have to keep an eye out

dexcom/device.go
```diff
@@ -121,20 +128,11 @@ func (d *Devices) Normalize(normalizer structure.Normalizer) {
...
 	TransmitterGeneration *string         `json:"transmitterGeneration,omitempty" yaml:"transmitterGeneration,omitempty"`
 	DisplayDevice         *string         `json:"displayDevice,omitempty" yaml:"displayDevice,omitempty"`
-	SoftwareVersion       *string         `json:"softwareVersion,omitempty" yaml:"softwareVersion,omitempty"`
-	SoftwareNumber        *string         `json:"softwareNumber,omitempty" yaml:"softwareNumber,omitempty"`
-	Language              *string         `json:"language,omitempty" yaml:"language,omitempty"`
-	IsMmolDisplayMode     *bool           `json:"isMmolDisplayMode,omitempty" yaml:"isMmolDisplayMode,omitempty"`
-	IsBlindedMode         *bool           `json:"isBlindedMode,omitempty" yaml:"isBlindedMode,omitempty"`
-	Is24HourMode          *bool           `json:"is24HourMode,omitempty" yaml:"is24HourMode,omitempty"`
-	DisplayTimeOffset     *int            `json:"displayTimeOffset,omitempty" yaml:"displayTimeOffset,omitempty"`
-	SystemTimeOffset      *int            `json:"systemTimeOffset,omitempty" yaml:"systemTimeOffset,omitempty"`
```
jh-bate, line 137: this has all gone from v2 => v3 API

dexcom/event.go
```diff
@@ -176,9 +194,12 @@ func (e *Event) Parse(parser structure.ObjectParser) {
 	e.Type = parser.String("eventType")
 	e.SubType = parser.String("eventSubType")
 	e.Unit = parser.String("unit")
-	e.Value = parser.Float64("value")
-	e.ID = parser.String("eventId")
+	e.Value = parser.String("value")
```
jh-bate, line 197: value now a string - so we need to parse as float form some of the validation checks below

dexcom/event.go
```diff
@@ -196,44 +217,69 @@ func (e *Event) Validate(validator structure.Validator) {
...
+	validator.String("value", e.Value).Exists().NotEmpty()
+	if e.Value != nil {
+		floatVal, _ := strconv.ParseFloat(*e.Value, 64)
+		validator.Float64("value", &floatVal).Exists().InRange(EventValueCarbsGramsMinimum, EventValueCarbsGramsMaximum)
 	}
 }
 
 func (e *Event) validateExercise(validator structure.Validator) {
-	// HACK: Dexcom - value of -1 is invalid; ignore unit and value instead (per Dexcom)
-	if e.Value != nil && *e.Value == -1.0 {
```
jh-bate, line 215: I optimistically removed this ... we are now dealing with strings but again might have to do some focused testing for this

dexcom/event.go
```diff
@@ -196,44 +217,69 @@ func (e *Event) Validate(validator structure.Validator) {
...
-	validator.Float64("value", e.Value).EqualTo(0)
+	validator.String("value", e.Value).Exists().NotEmpty()
+}
+
+func (e *Event) validateNote(validator structure.Validator) {
+	validator.String("value", e.Value).Exists().NotEmpty()
+}
+
+func (e *Event) validateUnkown(validator structure.Validator) {
+	validator.String("value", e.Value).Exists().NotEmpty()
```
jh-bate, line 265: new types - just checking we have some sort of required value

dexcom/event.go
```diff
@@ -196,44 +217,69 @@ func (e *Event) Validate(validator structure.Validator) {
...
-		validator.Float64("value", e.Value).Exists().InRange(EventValueInsulinUnitsMinimum, EventValueInsulinUnitsMaximum)
+	validator.String("unit", e.Unit).Exists().OneOf(EventUnitInsulinUnits)
+	validator.String("value", e.Value).Exists().NotEmpty()
+	if e.Value != nil {
+		floatVal, _ := strconv.ParseFloat(*e.Value, 64)
+		validator.Float64("value", &floatVal).Exists().InRange(EventValueInsulinUnitsMinimum, EventValueInsulinUnitsMaximum)
+	}
+}
+
+func (e *Event) validateBG(validator structure.Validator) {
```
jh-bate, line 278: again a new type

dexcom/event_test.go
```diff
@@ -64,38 +45,6 @@ var _ = Describe("Event", func() {
 		Expect(dexcom.EventValueExerciseMinutesMinimum).To(Equal(0.0))
 	})
 
-	It("EventSubTypeHealthAlcohol is expected", func() {
```
jh-bate, line 67: have combined these into the one test below for the sake of brevity and sanity ... imo anyway

dexcom/fetch/translate.go
```diff
@@ -388,8 +376,26 @@ func translateEGVToDatum(egv *dexcom.EGV, unit *string, rateUnit *string) data.D
 				"threshold": dexcom.EGVValuePinnedMgdLMaximum,
 			}}
 		}
+	case dexcom.EGVUnitMmolL:
+		if *datum.Value < dexcom.EGVValuePinnedMmolLMinimum {
+			datum.Value = pointer.FromFloat64(dexcom.EGVValuePinnedMmolLMinimum - 0.1) //check this
```
jh-bate, line 381: seeking feedback on this - just went for arbitory minimal value

dexcom/fetch/translate.go
```diff
@@ -388,8 +376,26 @@ func translateEGVToDatum(egv *dexcom.EGV, unit *string, rateUnit *string) data.D
...
+			datum.Annotations = &metadata.MetadataArray{{
+				"code":      "bg/out-of-range",
+				"value":     "low",
+				"threshold": dexcom.EGVValuePinnedMmolLMinimum,
+			}}
+		} else if *datum.Value > dexcom.EGVValuePinnedMmolLMaximum {
+			datum.Value = pointer.FromFloat64(dexcom.EGVValuePinnedMmolLMaximum + 0.1) //check this
```
jh-bate, line 388: seeking feedback on this - just went for arbitory minimal value

dexcom/fetch/translate.go
```diff
@@ -507,3 +518,61 @@ func translateEventInsulinToDatum(event *dexcom.Event) data.Datum {
...
+}
+
+func translateEventNoteToDatum(event *dexcom.Event) data.Datum {
+	datum := dataTypesStateReported.New()
+
+	// TODO: Refactor so we don't have to clear these here
+	datum.ID = nil
+	datum.GUID = nil
+
+	//TODO: jhbate question, note has a subtype?
```
jh-bate, line 550: not that I have see from test data so far

dexcom/fetch/translate.go
```diff
@@ -507,3 +518,61 @@ func translateEventInsulinToDatum(event *dexcom.Event) data.Datum {
...
+			datum.States = &dataTypesStateReported.StateArray{{State: pointer.FromString(dataTypesStateReported.StateStateCycle)}}
+		case dexcom.EventSubTypeHealthAlcohol:
+			datum.States = &dataTypesStateReported.StateArray{{State: pointer.FromString(dataTypesStateReported.StateStateAlcohol)}}
+		}
+	}
+	if event.ID != nil {
+		datum.Origin = &origin.Origin{ID: pointer.CloneString(event.ID)}
+	}
+
+	//TODO: jhbate question, value should now be string??
```
jh-bate, line 571: tbd

dexcom/calibration.go
```diff
@@ -103,23 +114,35 @@ func NewCalibration() *Calibration {
...
-	validator.String("unit", c.Unit).Exists().OneOf(CalibrationUnits()...)
+	validator.String("transmitterId", c.TransmitterID).Exists().Using(TransmitterIDValidator)
+	validator.Int("transmitterTicks", c.TransmitterTicks).Exists().GreaterThan(0)
+	validator.String("displayDevice", c.DisplayDevice).Exists().NotEmpty()
+	validator.String("transmitterGeneration", c.TransmitterGeneration).Exists().OneOf(DeviceTransmitterGenerations()...)
+	validator.String("unit", c.Unit).OneOf(CalibrationUnits()...)
 	if c.Unit != nil {
 		switch *c.Unit {
 		case CalibrationUnitMgdL:
 			validator.Float64("value", c.Value).Exists().InRange(CalibrationValueMgdLMinimum, CalibrationValueMgdLMaximum)
```
tjotala, line 141: I realize this was existing code, but I guess I'm surprised we allowed a floating point value for what ought to be an integer value (mg/dL).

tjotala: Equally strange that the API doc claim it's an `int32` type. I doubt that's true for endpoint that provides mmol/L.

dexcom/calibration_test.go
```diff
@@ -20,7 +20,85 @@ var _ = Describe("Calibration", func() {
...
+				validator := validator.New()
+				testCalibration.Validate(validator)
+				Expect(validator.Error()).To(HaveOccurred())
+			},
+			Entry("required id is not set", func() *dexcom.Calibration {
+				device := test.RandomCalibration()
+				device.ID = nil
+				return device
+			}),
+			Entry("required id is not set", func() *dexcom.Calibration {
```
tjotala, line 48: This looks like a duplicate of the case above?

dexcom/device.go
tjotala: Yup, we found that empirically - it wasn't in the v2 spec, either.

dexcom/egv.go
```diff
@@ -171,33 +182,39 @@ func NewEGV(unit *string) *EGV {
 }
 
 func (e *EGV) Parse(parser structure.ObjectParser) {
+	e.ID = parser.String("recordId")
 	e.SystemTime = TimeFromRaw(parser.ForgivingTime("systemTime", TimeFormat))
 	e.DisplayTime = TimeFromRaw(parser.ForgivingTime("displayTime", TimeFormat))
 	e.Value = parser.Float64("value")
```
tjotala, line 188: Interesting, the old code never picked up the `Unit` field. Should be added here?

tjotala: Looks like the unit tests never covered that, so it was probably never discovered.

dexcom/egv_test.go
```diff
@@ -24,91 +22,141 @@ var _ = Describe("EGV", func() {
...
+		Expect(dexcom.EGVValuePinnedMmolLMaximum).To(Equal(22.20299))
 	})
 
 	It("EGVTransmitterTickMinimum is expected", func() {
 		Expect(dexcom.EGVTransmitterTickMinimum).To(Equal(0))
 	})
 
 	It("EGVsResponseRateUnits returns expected", func() {
-		Expect(dexcom.EGVsResponseRateUnits()).To(Equal([]string{"mg/dL/min"}))
+		Expect(dexcom.EGVsResponseRateUnits()).To(Equal([]string{"unknown", "mg/dL/min", "mmol/L/min"}))
```
tjotala, line 46: Consider using the enum values defined above here, like in the other tests.

dexcom/egv_test.go
```diff
@@ -24,91 +22,141 @@ var _ = Describe("EGV", func() {
...
+		Expect(dexcom.EGVsResponseRateUnits()).To(Equal([]string{dexcom.EGVUnitUnknown, dexcom.EGVUnitMgdLMinute, dexcom.EGVUnitMmolLMinute}))
 	})
 
 	It("EGVsResponseUnits returns expected", func() {
-		Expect(dexcom.EGVsResponseUnits()).To(Equal([]string{"mg/dL"}))
+		Expect(dexcom.EGVsResponseUnits()).To(Equal([]string{"unknown", "mg/dL", "mmol/L"}))
```
tjotala, line 51: Consider using the enum values defined above here, like in the other tests.

dexcom/event.go
tjotala: I wonder if the docs are correct, since the example below the API fields shows a numerical value: https://developer.dexcom.com/docs/dexcomv3/operation/getEventsV3/

tjotala: Put another way, do we need to accommodate this field being either a string or a float?

dexcom/event.go
```diff
@@ -196,44 +217,69 @@ func (e *Event) Validate(validator structure.Validator) {
 			e.validateHealth(validator)
 		case EventTypeInsulin:
 			e.validateInsulin(validator)
+		case EventTypeNotes:
+			e.validateNote(validator)
+		case EventTypeBG:
+			e.validateBG(validator)
+		case EventTypeUnknown:
+			e.validateUnkown(validator)
```
tjotala, line 225: Nit: typo in method name `e.validateUnkown(validator)` - but matches the actual method declaration :)

dexcom/event.go
```diff
@@ -196,44 +217,69 @@ func (e *Event) Validate(validator structure.Validator) {
...
+	validator.String("unit", e.Unit).Exists().OneOf(EventUnitInsulinUnits)
+	validator.String("value", e.Value).Exists().NotEmpty()
+	if e.Value != nil {
+		floatVal, _ := strconv.ParseFloat(*e.Value, 64)
+		validator.Float64("value", &floatVal).Exists().InRange(EventValueInsulinUnitsMinimum, EventValueInsulinUnitsMaximum)
+	}
+}
+
+func (e *Event) validateBG(validator structure.Validator) {
+	validator.String("unit", e.Unit).Exists().OneOf(EventUnitMgdL)
```
tjotala, line 279: Just a note to comment that I find it a bit odd they only appear to allow `mg/dL` for this type.

tjotala: (in the API docs, that is)

dexcom/fetch/runner.go
tjotala: I think there's not any other choice, assuming we actually get a `transmitterId`.

dexcom/device.go
```diff
@@ -37,15 +41,18 @@ func DeviceDisplayDevices() []string {
 
 func DeviceTransmitterGenerations() []string {
 	return []string{
+		DeviceTransmitterGenerationUnknown,
 		DeviceTransmitterGenerationG4,
 		DeviceTransmitterGenerationG5,
 		DeviceTransmitterGenerationG6,
-		DeviceTransmitterGenerationG6Pro,
+		DeviceTransmitterGenerationG6Plus,
```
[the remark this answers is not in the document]
jh-bate: changed as per https://developer.dexcom.com/docs/dexcomv3/operation/getDevicesV3/#!c=200&path=records/transmitterGeneration&t=response

Is there a chance that g6 pro is an undocumented device generation? I will add it back with a comment

dexcom/device_test.go
```diff
@@ -2,53 +2,86 @@ package dexcom_test
...
 
 	It("DeviceTransmitterGenerations returns expected", func() {
-		Expect(dexcom.DeviceTransmitterGenerations()).To(Equal([]string{"g4", "g5", "g6", "g6 pro"}))
+		Expect(dexcom.DeviceTransmitterGenerations()).To(Equal([]string{"unknown", "g4", "g5", "g6", "g6+", "dexcomPro", "g7"}))
+		Expect(dexcom.DeviceTransmitterGenerations()).To(Equal([]string{
+			dexcom.DeviceTransmitterGenerationUnknown,
+			dexcom.DeviceTransmitterGenerationG4,
+			dexcom.DeviceTransmitterGenerationG5,
+			dexcom.DeviceTransmitterGenerationG6,
+			dexcom.DeviceTransmitterGenerationG6Plus,
```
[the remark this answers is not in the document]
jh-bate: as above

dexcom/event.go
jh-bate: yeah - that will be me. Spelling is not a string point :)

dexcom/event.go
jh-bate: yeah ... I am wondering if this is caught up in a future proofing / migration for units. [In the docs ](https://developer.dexcom.com/docs/dexcomv3/endpoint-overview/#units) it says that units are specified with the response BUT also that 

>  At the moment, all glucose values are reported in units of mg/dL and trend rates in units of mg/dL/min

dexcom/egv.go
jh-bate: So maybe it wasn't specified when the work was done, although it is shown in dexcom's `v2`? I will add it in

dexcom/device.go
jh-bate: interestingly `"g4" "g5" "g6"` are the only documented ones for the v2 spec

dexcom/event.go
jh-bate: from the data I have seen in the v3 sandbox, from all generations of devices,  value is now def now a string
e.g. 

` "value": "30.00",
   "unit": "grams",`

dexcom/device.go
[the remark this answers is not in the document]
jh-bate: ahhh ok - read you comment else where so understand now

dexcom/egv.go
jh-bate: Ahhh ok figured it out.

v2 at the top level specified the `unit` and `rateUnit`. This has now been pushed down into each individual reading in v3. Thanks for the question though and I will clean that up

dexcom/fetch/translate.go
tjotala: Given that the current API may only return `mg/dL`, this seems like a reasonable low-risk approach.

dexcom/fetch/translate.go
tjotala: Seems reasonable.

dexcom/device.go
jh-bate: FYI `g6 pro` listed with a note that it isn't listed in the spec

dexcom/egv.go
```diff
@@ -75,9 +87,7 @@ func EGVTrends() []string {
 }
 
 type EGVsResponse struct {
-	RateUnit *string `json:"rateUnit,omitempty"`
-	Unit     *string `json:"unit,omitempty"`
```
jh-bate, line 79: unit and rateUnit now returned for each record rather than the set of records

dexcom/calibration.go
jh-bate: my assumption was for `mmol/L` units usage in the future? or  outside "US" usage

dexcom/fetch/translate.go
lostlevels, line 381: More of a question for my understanding. If the value is lower than the theshold minimum, we want to still keep it below the minimum but only by a small amount?

dexcom/test/calibration.go
```diff
@@ -42,14 +44,22 @@ func CloneCalibrations(datum *dexcom.Calibrations) *dexcom.Calibrations {
...
 func RandomCalibration() *dexcom.Calibration {
 	datum := dexcom.NewCalibration()
+	datum.ID = pointer.FromString(test.RandomString())
 	datum.SystemTime = RandomSystemTime()
 	datum.DisplayTime = RandomDisplayTime()
 	datum.Unit = pointer.FromString(test.RandomStringFromArray(dexcom.CalibrationUnits()))
 	switch *datum.Unit {
 	case dexcom.CalibrationUnitMgdL:
 		datum.Value = pointer.FromFloat64(test.RandomFloat64FromRange(dexcom.CalibrationValueMgdLMinimum, dexcom.CalibrationValueMgdLMaximum))
+	case dexcom.CalibrationUnitMmolL:
```
lostlevels, line 54: nitpick: since `case dexcom.CalibrationUnitMmolL:` and `case dexcom.CalibrationUnitUnknown:` assign the same values, could just combine the two into `case dexcom.CalibrationUnitMmolL, dexcom.CalibrationUnitUnknown:` - unless there's some future logic that the two will additionally do.

dexcom/event.go
```diff
@@ -196,44 +217,69 @@ func (e *Event) Validate(validator structure.Validator) {
...
 
 func (e *Event) validateCarbs(validator structure.Validator) {
 	validator.String("eventSubType", e.SubType).NotExists()
-	if e.Unit != nil || e.Value != nil {
-		validator.String("unit", e.Unit).Exists().OneOf(EventUnitCarbsGrams)
-		validator.Float64("value", e.Value).Exists().InRange(EventValueCarbsGramsMinimum, EventValueCarbsGramsMaximum)
+	validator.String("unit", e.Unit).Exists().OneOf(EventUnitCarbsGrams)
+	validator.String("value", e.Value).Exists().NotEmpty()
+	if e.Value != nil {
+		floatVal, _ := strconv.ParseFloat(*e.Value, 64)
```
lostlevels, line 240: I see this in a few places. Is it ok to not handle the error? Should the value to replaced if there is an error? Does it not matter?

dexcom/alert.go
```diff
@@ -304,6 +355,23 @@ func (a *AlertScheduleSettings) Parse(parser structure.ObjectParser) {
...
+	a.Active = parser.Bool("isActive")
+	a.Override = ParseOverrideSetting(parser.WithReferenceObjectParser("override"))
+}
+
+func ParseOverrideSetting(parser structure.ObjectParser) *OverrideSetting {
+	if !parser.Exists() {
+		return nil
+	}
+	datum := &OverrideSetting{}
+	parser.Parse(datum)
```
lostlevels, line 367: I see this in a couple places and seems a lot of it is based on previous code. Is there a reason not to handle any errors here? Is it because the object will be validated later which takes care of that?

lostlevels (dismissed):
Looks fine, added a few questions.

lostlevels (approved):
LGTM @jh-bate Still have questions from previous review though.

dexcom/alert.go
jh-bate: yes that is my assumption

dexcom/dexcom.go
```diff
@@ -10,10 +10,38 @@ import (
...
+	TimeFormat  = "2006-01-02T15:04:05"
+	TimeFormatZ = "2006-01-02T15:04:05-07:00"
+	// lastUploadDate
+	TimeFormatUTC = "2006-01-02T15:04:05Z"
+
 	SystemTimeNowThreshold = 24 * time.Hour
 )
 
+func timeFormats() map[int]string {
+	return map[int]string{
```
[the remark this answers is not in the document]
jh-bate: its possible but this was the simplest thing for now. I will add a comment

lostlevels (approved):

dexcom/event.go
jh-bate: hmm good point - it matters in that we should fail the validation if we can't parse the float value from the string. I have updated the code and added tests

dexcom/fetch/translate.go
jh-bate: yes - I don't understand the reasoning 100% but that seems to be the decision

jh-bate (commented):
@lostlevels - thanks for the reviews. I made some adjustments based on your feedback and also my testing with the sandbox. 🤞🏻 one last review :)

lostlevels (dismissed):

PR #636, opened 2023-04-19
Update tools container to fix build failures

Roukoswarf:
Since we also need to move to mongodb 6.x, and the clientside mongosh tools will work fine against older versions, this commit moves us to 6.x, instead of just the latest 4.x.

### DIFF
--- a/Dockerfile.tools
+++ b/Dockerfile.tools
@@ -16,7 +16,8 @@ CMD ["make", "service-start"]
 
 # Production
 FROM mongo:6.0.5 AS production
-ENV ENV="/home/tidepool/.bashrc" DEBIAN_FRONTEND="noninteractive"
+# this statically set $HOME is non-ideal, but is to combat it being hardcoded to /data/db upstream
+ENV HOME="/home/tidepool/" DEBIAN_FRONTEND="noninteractive"
 RUN apt -y update && \
     apt -y install ca-certificates tzdata && \
     adduser --disabled-password tidepool
--- a/Dockerfile.tools
+++ b/Dockerfile.tools
@@ -15,7 +15,7 @@ RUN ["make", "service-build"]
 CMD ["make", "service-start"]
 
 # Production
-FROM mongo:4.2.14 AS production
+FROM mongo:6.0.5 AS production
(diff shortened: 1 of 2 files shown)


jh-bate (commented):
Is there a production version of mongo that use and should we tie our versions to that?? Or a [process similar to the golang version update](https://github.com/tidepool-org/platform#upgrade) where we go through and use a consistent version. 

- `MONGODB=5.3.2` is used in [travis.yml](https://github.com/tidepool-org/platform/blob/master/.travis.yml#L14)
- `4.2.14` was used in tools ... which looks like an oversight 
- In the development repo [we suggest](https://github.com/tidepool-org/development#install-and-configure-mongodb) min is `5.0` and tested to `6.x`

Roukoswarf:
> Is there a production version of mongo that use and should we tie our versions to that?? Or a [process similar to the golang version update](https://github.com/tidepool-org/platform#upgrade) where we go through and use a consistent version.
> 
>     * `MONGODB=5.3.2` is used in [travis.yml](https://github.com/tidepool-org/platform/blob/master/.travis.yml#L14)
> 
>     * `4.2.14` was used in tools ... which looks like an oversight
> 
>     * In the development repo [we suggest](https://github.com/tidepool-org/development#install-and-configure-mongodb) min is `5.0` and tested to `6.x`

We use 4.4.19 in prod right now, but thats very soon going to be 5 or 6.

The tools container is just used for running some utilities, and should be much less sensitive to moving to 6.

The development repo readme suggests 5.x+ as technically we develop on 5.x+, but also technically, until recently our mongodb libs did not support 5.x or 6.x, we just never hit any removed/changed functions in our code that we know of.

Travis.yaml was updated from 4.4.14 (our version in prod at the time i set it) to 5.3.2 by jimmy when he updated the deps to support 5.x+.

Though, these build failures are all kinda secondary to the mongo version, as the root cause would be the age of the container itself, if we instead of moving to 6.x, went to 4.4.19, it would work fine, but since this container is mostly tools for interactive use, id rather get ahead and use 6.x now.

toddkazakov (approved):

PR #637, opened 2023-05-11
test is "p" in v3 path is still relevent

jh-bate:


jh-bate:
tested and not required

PR #639, opened 2023-05-22
stopgap for when no units sent

jh-bate:
From support 

> OK, looks like there are 17 users who are impacted by this. They are all showing the same error: records of type bloodGlucose are coming in without a unit

### DIFF
--- a/dexcom/egv.go
+++ b/dexcom/egv.go
@@ -5,6 +5,7 @@ import (
 
 	dataBloodGlucose "github.com/tidepool-org/platform/data/blood/glucose"
 	"github.com/tidepool-org/platform/data/types/settings/cgm"
+	"github.com/tidepool-org/platform/pointer"
 	"github.com/tidepool-org/platform/structure"
 	structureValidator "github.com/tidepool-org/platform/structure/validator"
 )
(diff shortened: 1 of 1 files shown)


tjotala (changes requested):
I think we need a similar change in https://github.com/tidepool-org/platform/blob/master/dexcom/event.go#L299 as well because at least in one sample case I saw the immediate parent of the validation error is GetEvents, not GetEGVs:

```
        "detail": "unable to get events",
        "caller": {
          "package": "github.com/tidepool-org/platform/dexcom/client",
          "function": "GetEvents",
          "file": "dexcom/client/client.go",
          "line": 80
        }
```

dexcom/egv.go
```diff
@@ -185,7 +185,7 @@ func (e *EGV) Parse(parser structure.ObjectParser) {
 	e.ID = parser.String("recordId")
 	e.SystemTime = TimeFromString(parser.String("systemTime"))
 	e.DisplayTime = TimeFromString(parser.String("displayTime"))
-	e.Unit = parser.String("unit")
+	e.Unit = UnitFromParser(parser, EGVUnitMgdL)
 	e.RateUnit = parser.String("rateUnit")
```
tjotala, line 189: Likely need same default handling for `rateUnit` as well

tjotala: Also for the `unit` in calibrations (https://github.com/tidepool-org/platform/blob/master/dexcom/calibration.go)

jh-bate: I agree with `rateUnit` but question the unit in `calibration`.

Looking at both our code and the dexcom API it is allowed for the units to potentially not be set 

- https://github.com/tidepool-org/platform/blob/cd87510bb586c45a52c69f11680cba48a357ff6a/dexcom/calibration.go#L144-L153
- https://developer.dexcom.com/docs/dexcomv3/operation/getCalibrationsV3/#!c=200&path=records/unit&t=response

tjotala: Sorry, I should have been more clear. I did notice that the calibration record will pass validation as-is.

In the other cases (EGVs, events) though, the proposed solution has the parser filling in the missing unit with a default value, thereby passing the subsequent validation. That also means that the subsequent DB record will end up with that filled-in (default) unit.

tjotala (approved):
LGTM as-is.

PR #640, opened 2023-05-24
[BACK-2523] allow for missing event value

jh-bate:


dexcom/event.go
```diff
@@ -208,17 +208,22 @@ func (e *Event) Parse(parser structure.ObjectParser) {
 		switch *e.Type {
 		case EventTypeCarbs:
 			e.Unit = StringOrDefault(parser.String("unit"), EventUnitCarbsGrams)
+			e.Value = StringOrDefault(parser.String("value"), EventValueCarbsGramsMinimum)
```
tjotala, line 211: It would be great if we could track how often this actually happens. Right now we know it's N=1. Can we add logging here or in the `StringOrDefault` method?

tjotala (approved):
Adding log would be a nice bonus, but not critical.

PR #641, opened 2023-05-25
[BACK-2531] Dexcom followup

jh-bate:
- add basic logging for when a default value is used 
- add update to calibration units also 
- allow for the case where the value is an empty string and we should return the default


tjotala (commented):
LGTM, though for some reason the previous fix did not appear to resolve the issue as it is still happening.

dexcom/default.go
```diff
@@ -4,14 +4,20 @@ import (
...
 	"github.com/tidepool-org/platform/pointer"
+	"github.com/tidepool-org/platform/structure"
 )
 
 // NOTE: currently we are finding that g7 data occasionally doesn't have the required units
 // data. This is a workaround until that issue is resolved
-func StringOrDefault(givenValue *string, defaultValue interface{}) *string {
+func StringOrDefault(parser structure.ObjectParser, reference string, defaultValue interface{}) *string {
+	givenValue := parser.String(reference)
 	if givenValue == nil && defaultValue != nil {
```
tjotala, line 14: I think we may need to expand this to also detect an empty string value (i.e. "") - in addition to the `nil`. Unless the string parser turns an empty string value into a nil?

tjotala (approved):
LGTM.

PR #643, opened 2023-06-22
[BACK-2550] Handle stuck/broken tasks in task service automatically

Roukoswarf:


task/queue/queue.go
```diff
@@ -191,13 +200,23 @@ func (q *Queue) runTask(ctx context.Context, tsk *task.Task) {
 
 func (q *Queue) startManager(ctx context.Context) {
 	q.waitGroup.Add(1)
-	go func() {
+
+	// pick a starting random time in a future cycle to ensure multiple daemons don't do this exactly at the same
+	// time, it is not an error condition if it does, but could stress the db if the collection gets large
+	nextUnstickTime := pointer.FromAny(time.Now().Add(time.Duration(rand.Int63n(int64(q.delay * 15)))))
```
lostlevels, line 206: Not that there's anything inherently wrong with this but is there any reason why this needs to be a pointer?

lostlevels (approved):

task/queue/queue.go
Roukoswarf: Probably an artifact of a previous setup I had, the following goroutine only runs a single thread, ill move the initial unstick time into the goroutine instead, before the internal loop.

lostlevels (approved):

PR #645, opened 2023-07-27
start of fetching of dataRange

jh-bate:
WIP

### DIFF
--- a/dexcom/data_range.go
+++ b/dexcom/data_range.go
@@ -1,0 +1,123 @@
+package dexcom
+
+import (
+	"github.com/tidepool-org/platform/structure"
+	structureValidator "github.com/tidepool-org/platform/structure/validator"
+)
+
+type DataRange struct {
+	Start *DateRange `json:"start,omitempty"`
+	End   *DateRange `json:"end,omitempty"`
+}
+
+type DateRange struct {
+	SystemTime  *Time `json:"systemTime,omitempty"`
+	DisplayTime *Time `json:"displayTime,omitempty"`
+}
+
+type DataRangeResponse struct {
+	RecordType    *string    `json:"recordType,omitempty"`
+	RecordVersion *string    `json:"recordVersion,omitempty"`
+	UserID        *string    `json:"userId,omitempty"`
+	Calibrations  *DataRange `json:"calibrations,omitempty"`
+	Egvs          *DataRange `json:"egvs,omitempty"`
+	Events        *DataRange `json:"events,omitempty"`
+}
+
+func ParseDataRangeResponse(parser structure.ObjectParser) *DataRangeResponse {
+	if !parser.Exists() {
+		return nil
+	}
+	datum := NewDataRangeResponse()
+	parser.Parse(datum)
+	return datum
+}
+
+func NewDataRangeResponse() *DataRangeResponse {
+	return &DataRangeResponse{}
+}
+
+func (d *DataRangeResponse) Parse(parser structure.ObjectParser) {
+	d.UserID = parser.String("userId")
+	d.RecordType = parser.String("recordType")
+	d.RecordVersion = parser.String("recordVersion")
+	d.Calibrations = ParseDataRange(parser.WithReferenceObjectParser("calibrations"))
+	d.Egvs = ParseDataRange(parser.WithReferenceObjectParser("egvs"))
+	d.Events = ParseDataRange(parser.WithReferenceObjectParser("events"))
+}
+
+func (d *DataRangeResponse) Validate(validator structure.Validator) {
+	if calibrationsValidator := validator.WithReference("calibrations"); d.Calibrations != nil {
+		d.Calibrations.Validate(calibrationsValidator)
+	} else {
+		calibrationsValidator.ReportError(structureValidator.ErrorValueNotExists())
+	}
+
+	if egvsValidator := validator.WithReference("egvs"); d.Egvs != nil {
+		d.Egvs.Validate(egvsValidator)
+	} else {
+		egvsValidator.ReportError(structureValidator.ErrorValueNotExists())
+	}
+
+	if eventsValidator := validator.WithReference("events"); d.Events != nil {
+		d.Events.Validate(eventsValidator)
+	} else {
+		eventsValidator.ReportError(structureValidator.ErrorValueNotExists())
+	}
+}
+
+func ParseDataRange(parser structure.ObjectParser) *DataRange {
+	if !parser.Exists() {
+		return nil
+	}
+	datum := NewDataRange()
+	parser.Parse(datum)
+	return datum
+}
+
+func NewDataRange() *DataRange {
+	return &DataRange{}
+}
+
+func (d *DataRange) Parse(parser structure.ObjectParser) {
+	d.Start = ParseNewDateRange(parser.WithReferenceObjectParser("start"))
+	d.End = ParseNewDateRange(parser.WithReferenceObjectParser("end"))
+}
+
+func (d *DataRange) Validate(validator structure.Validator) {
+	validator = validator.WithMeta(d)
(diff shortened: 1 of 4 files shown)


dexcom/data_range.go
```diff
@@ -0,0 +1,143 @@
...
+	d.Events = ParseDataRange(parser.WithReferenceObjectParser("events"))
+}
+
+func (d *DataRangeResponse) GetOldestStartDate() time.Time {
+	oldest := time.Time{}
+	if d.Calibrations.Start != nil {
+		oldest = d.Calibrations.Start.DisplayTime.Time
+	}
+	if d.Events.Start != nil {
+		if d.Events.Start.DisplayTime.Time.Before(oldest) {
```
tjotala, line 57: I think this logic will fail if `d.Calibrations.Start` was `nil`, which leaves `oldest` still zero. In that case, no `d.Events.Start.DisplayTime` could be *before* `oldest`.

dexcom/fetch/runner.go
```diff
@@ -399,9 +399,19 @@ func (t *TaskRunner) updateDataSet(dataSetUpdate *data.DataSetUpdate) error {
 }
 
 func (t *TaskRunner) fetchSinceLatestDataTime() error {
-	startTime := initialDataTime
+
+	var startTime time.Time
+	var err error
 	if t.dataSource.LatestDataTime != nil && startTime.Before(*t.dataSource.LatestDataTime) {
```
tjotala, line 405: With `startTime` now initialized to zero, isn't the 2nd expression always true?

dexcom/fetch/runner.go
```diff
@@ -399,9 +399,19 @@ func (t *TaskRunner) updateDataSet(dataSetUpdate *data.DataSetUpdate) error {
...
 		startTime = *t.dataSource.LatestDataTime
+	} else {
+		startTime, err = t.fetchDataRangeStart()
```
tjotala, line 408: It occurs to me that we could further utilize the dataRange API response, since it also returns the end times. We could use that to avoid making the subsequent calls, since we could already know there's no new data to fetch.

dexcom/data_range.go
jh-bate: have updated so that "oldest" begins at "now" and is updated as we get better info

dexcom/fetch/runner.go
```diff
@@ -381,27 +379,22 @@ func (t *TaskRunner) updateDeviceHash(device *dexcom.Device) bool {
...
+	var err error
+	if t.dataSource.LatestDataTime != nil {
 		startTime = *t.dataSource.LatestDataTime
+	} else {
+		startTime, err = t.fetchDataRangeStart()
+		if err != nil {
+			return err
+		}
+	}
+	if startTime.IsZero() {
```
jh-bate, line 394: updated so that when there is no start time we don't fetch anything

ewollesen:
I checked out this PR, and cherry-picked the non-merge commits onto the most recent master. The result is: https://github.com/tidepool-org/platform/tree/eric/fix-fetch-dexcom

My hope is that this proves a better place to re-ignite this review.

@jh-bate if you wanted to do something like this (or similar commands, as you see fit):
```
git fetch origin
git checkout back-325-data-range
git reset --hard eric/fix-fetch-dexcom
git push --force-with-lease origin back-325-data-range
```
That should bring this PR up to date with the current master, and thereby make reviewing it a lot easier.

dexcom/data_range_test.go
```diff
@@ -0,0 +1,121 @@
...
+				dataRange := test.RandomDataRange()
+				dataRange.Start.SystemTime = nil
+				return dataRange
+			}),
+			Entry("required start display time is not set", func() *dexcom.DataRange {
+				dataRange := test.RandomDataRange()
+				dataRange.Start.DisplayTime = nil
+				return dataRange
+			}),
+		)
```
ewollesen, line 49: Consider adding a test where it handles malformed time values. I suspect given the current code, they'll be silently dropped. Is that the behavior we want? I really don't know, I guess it could be, but I feel like that's an indicator that something's wrong and we might need to retry from the top.

dexcom/fetch/runner.go
```diff
@@ -538,6 +532,23 @@ func (t *TaskRunner) fetchData(startTime time.Time, endTime time.Time) (data.Dat
 	return datumArray, nil
 }
 
+func (t *TaskRunner) fetchDataRangeStart() (time.Time, error) {
+	t.logger.Debug("fetching dataRange ...")
+	response, err := t.DexcomClient().GetDataRange(t.context, t.tokenSource)
```
ewollesen, line 537: Contexts, by their nature, are specific to a given function call, and should be passed as the first argument to the method. https://go.dev/wiki/CodeReviewComments#contexts

ewollesen: Looks like this is just how TaskRunner is setup though, so I guess disregard. It goes against the usual convention, but as long as our tasks aren't multithreaded, it should be fine I guess.

dexcom/data_range.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil
+	}
+	datum := NewDateRange()
+	parser.Parse(datum)
+	return datum
+}
+
+func NewDateRange() *DateRange {
+	return &DateRange{
+		SystemTime:  NewTime(),
```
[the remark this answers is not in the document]
jh-bate: If you look through the other files I believe it follows the same convention of all response variables being pointers

dexcom/data_range.go
```diff
@@ -0,0 +1,151 @@
...
+		DisplayTime: NewTime(),
+	}
+}
+
+func (c *DateRange) Parse(parser structure.ObjectParser) {
+	c.SystemTime = TimeFromString(parser.String("systemTime"))
```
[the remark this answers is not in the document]
jh-bate: yes that is what I understand - it would be an error once we try to parse it

dexcom/data_range_test.go
jh-bate: if the time is malformed the parsing would return an error and the the validation will fail as the required date isn't set. From memory finding the date issues wasn't too hard

dexcom/data_range.go
```diff
@@ -0,0 +1,149 @@
...
+func (d *DataRangeResponse) Parse(parser structure.ObjectParser) {
+	d.UserID = parser.String("userId")
+	d.RecordType = parser.String("recordType")
+	d.RecordVersion = parser.String("recordVersion")
+	d.Calibrations = ParseDataRange(parser.WithReferenceObjectParser("calibrations"))
+	d.EGVs = ParseDataRange(parser.WithReferenceObjectParser("egvs"))
+	d.Events = ParseDataRange(parser.WithReferenceObjectParser("events"))
+}
+
+func (d *DataRangeResponse) GetOldestStartDate() time.Time {
```
ewollesen, line 51: I think GetOldestStartDate should return an error if none of the three time values are found.

```suggestion
func (d *DataRangeResponse) GetOldestStartDate() (time.Time, error) {
```

Right now, time.Now() is returned, then [it's paired with a minute ago](https://github.com/tidepool-org/platform/blob/back-325-data-range/dexcom/fetch/runner.go#L400), resulting in a negative range request to Dexcom, that even if they accept it and run it, is still requesting data for a span of time that's at most 1 minute. It's a waste of resources.

Instead, if none of these three values is found, we should return an error, and not waste more of our time, or Dexcom's.

The above should never happen anyway, as that would be an invalid DataRangeResponse, and should be caught before getting here, but why not just be safe and return an error?

ewollesen (changes requested):
I think I'd just like to see GetOldestStartDate return an error, otherwise, I'm happy. Nice work Jamie! Thanks for updating the PR and all.

ewollesen (approved):

jh-bate:
super seeded by later work

PR #649, opened 2023-08-15
Add EHR data sync tasks

toddkazakov:


Roukoswarf (commented):
Look correct.

Only question is should this be a toplevel package?
This is a pattern which was used for the dexcom task, which imo shouldnt be repeated.
Pieces which generally get used across platform, like pointer, auth, etc make sense to put in the top level as long as they don't cause circular deps on anything lower.

But this EHR task wont be pulled in by any code outside the task service, why not live there?

Roukoswarf (approved):

PR #651, opened 2023-08-15
Don't assume that 127.0.0.[2-255] aren't in use.

ewollesen:
A pair of tests were testing error paths for non-responsive hosts. To do so, they were attempting to connect to 127.0.0.2 and 127.0.0.3. However, some hosts are configured to respond to 127.0.0.0/8 as localhost, and with such a configuration, these tests could fail.

One alternative is to configure mongodb to bind specifically to 127.0.0.1. Another would be a test helper that tests the expected IP to make sure nothing responds before running the test, possibly skipping it if something responds (or trying another address?) But this solution seems like the simplest, easiest change to make, that covers the most possible configurations.


ewollesen:
Moving this PR to tidepool-org for CI.

PR #656, opened 2023-08-22
[BACK-2633] soft failure when running dexcom tasks

jh-bate:
Allow "soft failure" for dexcom task runner

- only implemented for dexcom tasks
- allow the task to be "retried" three times 
- on the third it will either pass or actually fail the task and not retry again 

FYI context from Darin:

```
I think, IIRC, what I would recommend is to:
• Add a new dynamic property of the Data field of task. TaskCreate (I think) that captures an "error count" (or "retry count" or similar).
• Only invoke tsk. AppendError if the "error count" exceeds some maximum. Otherwise, just exit that task run, but do not mark the task as failed.
• Add this functionality to fetch. Runner. Run in dexcom/fetch/runner.go at line 130 to allow a number of refresh token failures before truly failing the task.
• You'll probably want to add similar functionality throughout fetch. TaskRunner. Run in dexcom/fetch/runner. go in lines 193-213 to be more tolerant of any/all intermittent Dexcom API-related failures.
• Reset the error count upon a fully successful task run.
This should allow for any transient Dexcom API failures to not cause the task to immediately fail. You'll likely need to play around with the maximum "error count" to find a value that detects true failures without causing a large number of unnecessary requests. Perhaps start at a value of 5.
```


dexcom/fetch/runner.go
```diff
@@ -124,17 +124,16 @@ func (r *Runner) Run(ctx context.Context, tsk *task.Task) bool {
...
 
 		if serverSessionToken, sErr := r.AuthClient().ServerSessionToken(); sErr != nil {
-			tsk.AppendError(errors.Wrap(sErr, "unable to get server session token"))
+			ErrorOrRetryTask(tsk, errors.Wrap(sErr, "unable to get server session token"))
 		} else {
 			ctx = auth.NewContextWithServerSessionToken(ctx, serverSessionToken)
-
 			if taskRunner, tErr := NewTaskRunner(r, tsk); tErr != nil {
-				tsk.AppendError(errors.Wrap(tErr, "unable to create task runner"))
+				ErrorOrRetryTask(tsk, errors.Wrap(tErr, "unable to create task runner"))
```
toddkazakov, line 134: If this block is executed, this means that there is bug in the code. It's better not to fail the task here, but keep on retrying (until the bug is eventually fixed).

dexcom/fetch/runner.go
```diff
@@ -124,17 +124,16 @@ func (r *Runner) Run(ctx context.Context, tsk *task.Task) bool {
...
 			} else if tErr = taskRunner.Run(ctx); tErr != nil {
-				tsk.AppendError(errors.Wrap(tErr, "unable to run task runner"))
+				ErrorOrRetryTask(tsk, errors.Wrap(tErr, "unable to run task runner"))
```
toddkazakov, line 136: I think this is the only place where we should fail the task if retries are exceeded.

toddkazakov, line 136: We should log the error here, otherwise we won't know why a task has been retried.

dexcom/fetch/runner.go
```diff
@@ -124,17 +124,16 @@ func (r *Runner) Run(ctx context.Context, tsk *task.Task) bool {
 	}
 
 	if !skipToAvoidDexcomBackup {
-		tsk.ClearError()
+		ResetTask(tsk)
 
 		if serverSessionToken, sErr := r.AuthClient().ServerSessionToken(); sErr != nil {
-			tsk.AppendError(errors.Wrap(sErr, "unable to get server session token"))
+			ErrorOrRetryTask(tsk, errors.Wrap(sErr, "unable to get server session token"))
```
toddkazakov, line 130: We shouldn't fail the task if we can't obtain a session token from our authentication service, otherwise we'll fail all tasks if the auth service is down for whatever reason. Retries should be unlimited too.

toddkazakov (dismissed):

dexcom/fetch/runner.go
jh-bate: the error is logged when we fail the task originally not all errors caused a task failure within `taskRunner.Run`

Roukoswarf (approved):

dexcom/fetch/task.go
```diff
@@ -24,8 +27,48 @@ func NewTaskCreate(providerSessionID string, dataSourceID string) (*task.TaskCre
...
+	return true
+}
+
+func incrementTaskRetryCount(t *task.Task) {
+	if t.Data[dexcomTaskRetryField] != nil {
+		count, ok := t.Data[dexcomTaskRetryField].(int)
+		if ok {
+			count++
+			t.Data[dexcomTaskRetryField] = count
+		}
```
lostlevels, line 70: Is it worth resetting this to 1 if it's not an int?

lostlevels (approved):

PR #657, opened 2023-08-23
add endpoints for care partner alerting

ewollesen:
These are preliminary, and might move, but it's a small start.


ewollesen:
The travis failure appears to be a time value that was just outside of tolerance, probably that test is a bit flappy and/or sensitive? I'll re-run it.

data/carepartner/alerts/alerts.go
```diff
@@ -0,0 +1,85 @@
...
+	sCtx.RespondWithStatusAndData(http.StatusNotImplemented, nil)
+}
+
+// AuthWrapper lets the base service inject authentication.
+//
+// Cuz alerts don't care how users are authenticated.
+type AuthWrapper = func(service.HandlerFunc) service.HandlerFunc
+
+// PathBuilder builds a path whose root is pre-determined.
+type PathBuilder = func(segments ...string) string
```
toddkazakov, line 70: I'm not a big fan of this. We have other examples of a similar pattern which makes searching for env variables impossible (segments are underscores, so you can't just grep to find all uses of an env var). 

I think readability is more important, especially in this codebase which has a lot of abstraction and indirection, so I'd stick to static paths.

data/carepartner/alerts/alerts.go
```diff
@@ -0,0 +1,85 @@
...
+	lgr := log.LoggerFromContext(ctx)
+
+	lgr.Info("store an Apple push notification token")
+	sCtx.RespondWithStatusAndData(http.StatusNotImplemented, nil)
+}
+
+// AuthWrapper lets the base service inject authentication.
+//
+// Cuz alerts don't care how users are authenticated.
+type AuthWrapper = func(service.HandlerFunc) service.HandlerFunc
```
toddkazakov, line 67: The comment I left for §PathBuilder§ applies here too. I think it's better if we are explicit regarding who's authorized to access those endpoints, so it's obvious for the reader.

ewollesen: I guess I'd have to re-arch some things then. It'd be my preference to keep the alerts routes in the alerts package, but in order to use the existing `Authenticate()`, I'd have to move the alerts routes into the v1 package. Which feels like a coupling I'd prefer to avoid. That's the "real" reason for the AuthWrapper. As this is all one codebase, I'm not really saying that alerts shouldn't have to care how the authentication is being done, but it also shouldn't be reinventing that, but at the same time, with the way the v1 package is architected, I can't use `Authenticate()` from alerts, without creating a circular dependency.

Would you be opposed to lifting the `Authenticate()` out of v1 if I could find a reasonable way? That would let me keep the routes for alerts in the alerts package, and still have an explicit auth, granted it would still be in a different package, but at least which package would be clearly called out (as the function call would be something like `v1auth.Authenticate()` or so.)

data/carepartner/alerts/alerts.go
ewollesen: I understand where you're coming from with the grepping... I was thinking it might help avoid a situation where two packages are trying to register the same route. Maybe the routing library takes care of that, or maybe it just says "last to register wins", or who knows. Those kinds of bugs can be a real pain, and with the static routes that are in there now, most of the routes are hidden, ie the `DataSetRoutes`, `SourcesRoutes`, etc are off somewhere else, and point who knows where.

So for getting a view of which endpoints go where, it might almost be better to use the kind of scheme I've proposed because then data/service/api/v1/v1.go would at least be able to tell you where each package is rooted. The grepping is still an issue I guess, if you happen to grep for a complete route, but grepping for `/v1/carepartners/` for example would lead you to the v1.go file, where you could see that every route under that prefix is handled by the `alerts` package.

I'm not hard set on the idea, so I'll follow the existing pattern if after considering the above you aren't convinced. :)

data/carepartner/alerts/alerts.go
ewollesen: Re-reading this myself, I don't care about the routes being in the alerts package... that's a separation I can live with. It's almost better... they're just handlers, they don't know about routes. I can grok that.

data/carepartner/alerts/alerts.go
ewollesen: After more thought, I'm gonna just do static routes as you've asked. I was on the fence as to whether it was over complicating things or not, and your comments are a clear indicator that yes, it is.

data/carepartner/alerts/alerts.go
toddkazakov: You can use `service/api/RequireUser`/`service/api/RequireServer`/`service/api/Require`. I'm not sure why `data` has its own authentication middleware.

data/carepartner/alerts/alerts.go
toddkazakov: The `/v1/carepartner` prefix will never change for those routes, so there's no need build those dynamically. I think this is a code smell. 

You don't need to add your route table in `/data/service/api/v1`. You can define your routes in  `/carepartner/api/v1.go` and include those in `data/service/api/standard.go`. This way "data" related routes and "alerts" related routes will be separate.  

> I was thinking it might help avoid a situation where two packages are trying to register the same route.

When you don't have a top-level view of all routes and those are built dynamically it's easier to define a duplicate elsewhere because you can't even grep for the full route. If duplicate routes are a problem we should make the router panic when a duplicate is added.

ewollesen: > The /v1/carepartner prefix will never change for those routes, so there's no need build those dynamically. I think this is a code smell.

It was never about building them dynamically, it was about safely delegating control to divide the routing space. The v1 router could say, "OK carepartner, here's your slice of the routing domain, do with it what you will", while knowing that it couldn't do anything but generate routes under /v1/carepartner, which would allow the v1 package to ensure that there wouldn't be conflicts.

With the current setup, the carepartner package can add routes wherever it likes and there can be collisions. But if all the routes are defined clearly in v1 I don't see it as a problem. That said, I don't think they are at the moment (see DataSetRoutes, SourceRoutes and friends). If/when they are, it could still be a potential problem, but at least it should be easily found, and you're right, it's overall simpler to just have all the routes in one place, and greppable. That's good value. I'm for that.

> When you don't have a top-level view of all routes and those are built dynamically it's easier to define a duplicate elsewhere because you can't even grep for the full route. If duplicate routes are a problem we should make the router panic when a duplicate is added.

I don't see that it exists now (again see DataSetRoutes, SourceRoutes, etc), and from my reading, the router is from go-json-rest, so modifying it to panic it probably not a realistic option. I checked, and it says:

> // This validates the Routes and prepares the Trie data structure.
> // It must be called once the Routes are defined and before trying to find Routes.
> // The order matters, if multiple Routes match, the first defined will be used.

So there will be no warnings when multiple routes match.

IMO, one doesn't need a view of every route in existence in once place, so long as the trail is clearly marked, ie v1 delegates control to sub packages that can't add routes outside of their delegated prefixes.

But again, I agree with you that it's simpler to just have all the routes in v1 (or somewhere) clearly visible and searchable. So I'll redesign to that. :)

data/carepartner/alerts/alerts.go
ewollesen: It has a slightly different signature, but it's doing basically the same thing as `service/api/Require`. I think it should be calling `Require` under the hood, and could probably be modified to operate that way, but I'll handle that in a separate PR unless you don't think it's worth messing with?

data/carepartner/alerts/alerts.go
```diff
@@ -0,0 +1,85 @@
...
+//
+// All routes are prefixed with base.
+func Routes(path PathBuilder, auth AuthWrapper) []service.Route {
+	return []service.Route{
+		service.MakeRoute("GET", path(""), auth(List)),
+		service.MakeRoute("GET", path(":alertId"), auth(Get)),
+		service.MakeRoute("POST", path(""), auth(Create)),
+		service.MakeRoute("PUT", path(":alertId"), auth(Update)),
+		service.MakeRoute("POST", path(":alertId/acks"), auth(Ack)),
+		service.MakeRoute("POST", path(":userId/tokens/apple"), auth(CreateAppleToken)),
```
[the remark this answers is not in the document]
ewollesen: I don't think the :userId should even be necessary, as it's an authenticated endpoint, it should be acquired via the auth info.

toddkazakov:
I think it's a good practice to include the user id in the path segment. In
this case I think we'd want expose the alerts for a specific user to our
support team (we have an internal tool called orca).

ewollesen:
> I think it's a good practice to include the user id in the path segment. In this case I think we'd want expose the alerts for a specific user to our support team (we have an internal tool called orca).

OK, I can add that back. I've generally not included it, so that there's no chance of accidental data leaks where one user sees data for another user via an improperly configured handler. Support teams would usually have an "impersonate" feature to work around that.

But yes, I'll put it back.

alerts/config.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (m DurationMinutes) Duration() time.Duration {
+	return time.Duration(m)
+}
+
+// ValueWithUnits binds a value to its units.
+//
+// Other types can extend it to parse and validate the Units.
+type ValueWithUnits struct {
```
toddkazakov, line 95: You may want to reuse https://github.com/tidepool-org/platform/blob/master/data/types/blood/glucose/glucose.go instead

alerts/config.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil
+	case UnitsMillimollsPerLiter:
+		return nil
+	}
+	return fmt.Errorf("invalid units: %q", t.Units)
+}
+
+const (
+	// UnitsMilligramsPerDeciliter are a common blood-glucose measurement unit in the USA.
+	UnitsMilligramsPerDeciliter string = "mg/dL"
```
toddkazakov, line 126: Those constants are already defined [here](https://github.com/tidepool-org/platform/blob/master/data/blood/glucose/glucose.go#L10-L23)

alerts/config.go
```diff
@@ -0,0 +1,129 @@
...
+type ValueWithUnits struct {
+	Value float64 `json:"value"`
+	Units string  `json:"units"`
+}
+
+// Threshold is a value measured in either mg/dL or mmol/L.
+type Threshold ValueWithUnits
+
+// UnmarshalJSON adds validation of Units.
+func (t *Threshold) UnmarshalJSON(b []byte) error {
```
toddkazakov, line 104: A little odd to see validation logic as a custom unmarshaler. Doesn't this limit you to a single error per request? 

You may want to take a look at existing validation logic in the other platform services.

data/service/api/v1/alerts.go
```diff
@@ -0,0 +1,78 @@
...
+	repo := sCtx.AlertsRepository()
+
+	if err := checkPermissions(details, r.PathParam("userID")); err != nil {
+		sCtx.RespondWithError(platform.ErrorUnauthorized())
+		return
+	}
+
+	cfg := &alerts.Config{}
+	if err := json.NewDecoder(r.Body).Decode(cfg); err != nil {
+		sCtx.RespondWithError(platform.ErrorJSONMalformed())
```
toddkazakov, line 56: This will return a "json is malformed" error when validation fails, which isn't correct.

alerts/config.go
ewollesen: Thanks! I never would have seen these. :)

Hmm, I'll need to look further. I might be able to re-use some of the validation, but the actual types themselves are very different. Using Glucose would include [types.Base](https://github.com/tidepool-org/platform/blob/bdd98e0d295604a1504c6aaf6e9f2a2b8083f9d2/data/types/base.go#L33) and that's a whole lot of stuff I don't need/want.

Plus, in Glucose, both Units and Value are optional fields. Here, they're not optional, which means I can use even less of the validation. I'll take a look though. No sense in having validations in two places if it can be avoided.

alerts/config.go
ewollesen: Thanks for the link!

alerts/config.go
ewollesen: > A little odd to see validation logic as a custom unmarshaler. 

Odd how? It makes sense to me. Part of unmarshaling is validation, both syntactically, and semantically.

> Doesn't this limit you to a single error per request?

No, though that is how I've coded it. I could still continue after receiving an error, grouping them up and returning a grouped error at the end. I find it's usually just simpler to return at the first error though. The code stays simpler that way.
 
> You may want to take a look at existing validation logic in the other platform services.

Will do. I looked around a bit before this, but clearly didn't look hard enough. Now that I know where some other validations are, I'll look at them more closely.

data/service/api/v1/alerts.go
ewollesen: I don't see how that's not correct? Can you expand on that? Which error message would you use?

alerts/config.go
ewollesen: `glucose.Glucose` embeds `blood.Blood`, which embeds `types.Base`, and that's a _whole_ lot of stuff I'm definitely not interested in sharing.

Additionally, `Units` and `Value` are optional in `types.Base`, but they aren't optional here. I'm thinking I can re-use the `structure.Validator` though.

alerts/config.go
toddkazakov: > Odd how?

I don't think overloading marshaling/unmarshaling logic is a standard practice when working with web services, or at least I haven't seen this pattern in popular projects.

> Part of unmarshaling is validation, both syntactically, and semantically.

Marshaling deals with the way the data is formatted for transmission. I agree that there is validation step, e.g. when checking that a number fits in `int16`, but I think errors should only be raised when an incorrect data structure/primitive is used. Failing to unmarshal should mean that the data format is incorrect - e.g. when a client sends xml instead of json. 

Data validation comes after unmarshaling and ought to establish whether the received values lie within the semantic bounds we have defined.

> No, though that is how I've coded it. 

That's what I meant. Platform services usually return multiple errors - it's handy during development/debugging.

data/service/api/v1/alerts.go
toddkazakov: Let's say an alert threshold that we have received is beyond some bounds we have defined. This doesn't mean the json is malformed. The returned response ought to indicate which field failed validation and why.

ewollesen: I guess we just see that differently. I'm not opposed to semantic validation in there. It makes sense to me, as I consider that JSON to be malformed. It does not unmarshal into a valid go object (because the integer is out of range.) To me, that's malformed. While I can see your point, I think it's too fine a distinction to bother adding the distinction and not worth the added complexity.

But it's moot in this case, as you've shown me the existing validation package, and I'm glad to use that. :)

ewollesen: Maybe part of it is that I don't see UnmarshalJSON as validating only the JSON, I see it as saying, I have type Foo, and UnmarshalJSON should take in some JSON, and produce a valid Foo. Therefore, if there's any incompatability in doing so, that indicates that the JSON provided is invalid. It might be "valid" JSON, but that doesn't make it a valid Foo.

ewollesen:
@toddkazakov I think this is in a good place to review. But I also think it's too big. What do you think? I can gladly chop it into a few separate PRs if that makes it easier for you to review.

I'm going to remove the draft flag. You let me know if it looks daunting and you want me to break it into smaller PRs.

alerts/config.go
```diff
@@ -0,0 +1,158 @@
...
+	"context"
+	"time"
+
+	"github.com/tidepool-org/platform/data/blood/glucose"
+	"github.com/tidepool-org/platform/structure"
+)
+
+// Config models a user's desired alerts.
+type Config struct {
+	OwnerID         string         `json:"ownerID" bson:"ownerID"`
```
toddkazakov, line 16: Would it be better to use `userId` attribute for the owner? This would allow to standardize on a single field when trying to find records owned by a given user.

alerts/config.go
```diff
@@ -0,0 +1,158 @@
...
+	InvitorID       string         `json:"invitorID" bson:"invitorID"`
```
toddkazakov, line 17: This should be the userId of the user that's being "followed". If the account is custodial (soon we'll introduce the ability for regular users to create custodial accounts) the invitor id and followee id might be different. I think it's better to revise the attribute name.

Additionally, I don't think we should use uppercase ID for consistency with other json/bson models.

alerts/config.go
```diff
@@ -0,0 +1,158 @@
...
+	UrgentLow       *WithThreshold `json:"urgentLow,omitempty" bson:"urgentLow,omitempty"`
+	Low             *Deluxe        `json:"low,omitempty" bson:"low,omitempty"`
+	High            *Deluxe        `json:"high,omitempty" bson:"high,omitempty"`
+	NotLooping      *WithDelay     `json:"notLooping,omitempty" bson:"notLooping,omitempty"`
+	NoCommunication *WithDelay     `json:"noCommunication,omitempty" bson:"noCommunication,omitempty"`
+}
+
+func (c Config) Validate(validator structure.Validator) {
+	validator.String("ownerID", &c.OwnerID).LengthGreaterThan(0)
```
toddkazakov, line 26: You should use the user id validator here https://github.com/tidepool-org/platform/blob/104a8bf918ee3e8e2298c3de29d2289c127a5a31/user/user.go#L100

alerts/config.go
```diff
@@ -0,0 +1,158 @@
...
+	validator.String("invitorID", &c.InvitorID).LengthGreaterThan(0)
```
toddkazakov, line 27: You should use the user id validator here https://github.com/tidepool-org/platform/blob/104a8bf918ee3e8e2298c3de29d2289c127a5a31/user/user.go#L100

alerts/config.go
```diff
@@ -0,0 +1,158 @@
...
+	DelayMixin `bson:",inline"`
+}
+
+func (d WithDelay) Validate(validator structure.Validator) {
+	d.Base.Validate(validator)
+	d.DelayMixin.Validate(validator)
+}
+
+// Deluxe extends Base with both DelayMixin and ThresholdMixin.
+type Deluxe struct {
```
toddkazakov, line 104: I have trouble understanding where the name of this struct comes from. Can you please clarify or revise?

data/service/api/v1/alerts.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/tidepool-org/platform/alerts"
+	"github.com/tidepool-org/platform/data/service"
+	"github.com/tidepool-org/platform/permission"
+	"github.com/tidepool-org/platform/request"
+	platform "github.com/tidepool-org/platform/service"
+)
+
+func AlertsRoutes() []service.Route {
+	return []service.Route{
```
toddkazakov, line 17: Is it possible to expose a `GET` route for fetching an invite by id? This way it can be used by the QA team for testing purposes.

data/store/mongo/mongo_alerts.go
```diff
@@ -0,0 +1,55 @@
...
+	_, err := r.DeleteMany(ctx, filter, nil)
+	if err != nil {
+		return fmt.Errorf("upserting AlertsConfig: %w", err)
+	}
+	return nil
+}
+
+// AlertsConfigDocument wraps alerts.Config to provide an ID for mongodb.
+type AlertsConfigDocument struct {
+	ID             string `bson:"_id"`
```
toddkazakov, line 41: You're overriding the autogenerated ObjectId `_id` with a string identifier of the concatenated "ownerId:invitorId". There's a unique constraint on that index that cannot be removed. This implies the alert configuration is unique for the user id tuple. Is there a possibly down that we may allow multiple alerts/alert configurations?

permission/permission.go
```diff
@@ -7,10 +7,13 @@ import (
 type Permission map[string]interface{}
 type Permissions map[string]Permission
 
-const Owner = "root"
-const Custodian = "custodian"
-const Write = "upload"
-const Read = "view"
+const (
+	Alerting  = "alerting"
```
toddkazakov, line 11: I think it's better to define this as `follow`, because if I am not mistaken, in addition to alerting we will allow the follower to read (a subset?) of the device the data of the followee.

toddkazakov (commented):
Overall looks very good, only a few minor issues/questions.

alerts/config.go
ewollesen: Could go either way. My intention was to make it clearer what the roles of each of the two people were. They're both users after all.

alerts/config.go
ewollesen: Re invitor vs follower, that makes sense, thanks for the clarification. 

What would you recommend for a the attribute name? `FolloweeId` feels stranged... `FollowedId` maybe? I'll go with that I think. It's a least a little less impersonal than `SourceId`, heh.

alerts/config.go
ewollesen: Oh, thanks, will do. Hadn't seen that.

alerts/config.go
ewollesen: It's not a good name. Do you have a suggestion for something better?

Here "deluxe" is referring to the fact that this type of alert has all of the possible options, namely it has `DelayMixin` and `ThresholdMixin`.

data/service/api/v1/alerts.go
ewollesen: Absolutely, though I was planning for a follow-on PR for that, as this one is already quite large.

data/store/mongo/mongo_alerts.go
ewollesen: You tell me? Aside from a "never say never" situation, I don't know what that would be...

Is it bad form in mongo to override the generated IDs? I can move away from that... I was unsure about it in the first place, but I _do_ want to ensure that, at least for now, there is only a single alerts config for any given combination of owner and invitor.

permission/permission.go
ewollesen: @ps2 do you have any objection? We had discussed "alerting" as the term earlier. Personally, I don't feel strongly either way.

alerts/config.go
ewollesen: It's also just shorter than `BaseWithDelayAndThreshold`.

data/service/api/v1/alerts.go
ewollesen: To clarify, I will add the GET route now, for easier testing in QA.

permission/permission.go
ps2: I think either works for me as well. We could also be specific and say "CarePartner".

alerts/config.go
ewollesen: I'm going to stick with `ID` in go, but I'll use `Id` in the json/bson. Perhaps that what you meant anyway. After giving it some thought, this seems like the place to make that shift.

permission/permission.go
ewollesen: Will do. Going with Follow, as I feel it describes what it's allowing better than CarePartner.

ewollesen:
Updates from code review, and a rebase onto current master.

toddkazakov (approved):
LGTM

ewollesen:
Let's see if this still works:

/deploy dev1 data

PR #660, opened 2023-09-11
[BACK-37] Jellyfish data migration

jh-bate:
goal is to allow dedup while jellyfish is deprecated and uploader is migrated to only use platform API

- find existing data ingested via jellyfish and generate duplicator hash
- also see https://github.com/tidepool-org/jellyfish/pull/191


migrations/back-37/back-37.go
```diff
@@ -0,0 +1,202 @@
...
+				"_deduplicator": bson.M{"$exists": false},
+			}
+
+			var jellyfishResult bson.M
+			jellyfishDocCursor, err := m.dataRepository.Find(context.Background(), selector)
+			if err != nil {
+				logger.WithError(err).Error("Unable to find jellyfish results")
+			}
+			for jellyfishDocCursor.Next(context.Background()) {
+				jellyfishDocCursor.Decode(&jellyfishResult)
```
lostlevels, line 110: might want to check for error here

migrations/back-37/back-37.go
```diff
@@ -0,0 +1,202 @@
...
+
+				dupQuery := bson.M{
+					"_userId":       userID,
+					"_active":       true,
+					"_id":           jellyfishResult["_id"],
+					"_deduplicator": bson.M{"$exists": true},
+				}
+				dupCursor, err := m.dataRepository.Find(context.Background(), dupQuery)
```
lostlevels, line 118: would this ever return a document since the  *_id* is unique and the original query on line 101 already checks for no dedeupe hash? Or can the doc be modified in b/t this query and the one on line 105?

lostlevels (changes requested):

migrations/back-37/Hasher.go
```diff
@@ -0,0 +1,52 @@
...
+
+// used for BACK-37 to set hash for jellyfish datum for migrating to platform API
+func CreateHash(bsonData bson.M) string {
+	// represents data from base id fields
+	var baseData = func(bsonData bson.M) []string {
+		return []string{
+			bsonData["_userId"].(string),
+			bsonData["deviceId"].(string),
+			bsonData["time"].(string),
+			bsonData["type"].(string),
```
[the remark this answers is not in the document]
lostlevels: meant *time* field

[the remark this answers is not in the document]
jh-bate: crap - yeah it is. Thanks for that

migrations/back-37/back-37.go
jh-bate: the `_id` is yet another hash of the datum's "id" fields e.g. for the smbg `var idFields = ['type', 'deviceId', 'time', 'value'];`

see https://github.com/tidepool-org/jellyfish/blob/35165c8dce520390213fd2fbd7b0228931d976e3/lib/schema/smbg.js#L22 for example 

So if you had a duplicate value they would have the same `_id` from what I understand? 

@darinkrauss ??

lostlevels: There are two fields, the internal mongodb `_id` and our own generated `id` I believe. I think you want the `id` here.

migrations/back-37/Hasher.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/tidepool-org/platform/data/deduplicator/deduplicator"
+	"github.com/tidepool-org/platform/data/types"
+)
+
+// used for BACK-37 to set hash for jellyfish datum for migrating to platform API
+func CreateHash(bsonData bson.M) string {
+	// represents data from base id fields
+	var baseData = func(bsonData bson.M) []string {
+		return []string{
+			bsonData["_userId"].(string),
```
darinkrauss, line 18: [nit] I wouldn't be surprised if there is old data in the production database that is missing one or more of these fields. If there is, then I believe the forced typecast will panic and crash the migration process. Do we know with 100% certainty that these four fields are on every record and are of the expected type?

migrations/back-37/Hasher.go
```diff
@@ -0,0 +1,54 @@
...
+	"go.mongodb.org/mongo-driver/bson"
+
+	"github.com/tidepool-org/platform/data/deduplicator/deduplicator"
+	"github.com/tidepool-org/platform/data/types"
+)
+
+// used for BACK-37 to set hash for jellyfish datum for migrating to platform API
+func CreateHash(bsonData bson.M) string {
+	// represents data from base id fields
+	var baseData = func(bsonData bson.M) []string {
```
darinkrauss, line 16: Any reason this is a function? Why not simply calculate and use the result below.

migrations/back-37/Hasher.go
```diff
@@ -0,0 +1,54 @@
...
+	default:
+		theHash = makeHash(baseData(bsonData)...)
+	}
+	return theHash
+}
+
+func makeHash(fields ...string) string {
+	hash, err := deduplicator.GenerateIdentityHash(fields)
+	if err != nil {
+		panic(err)
```
darinkrauss, line 51: Do we want to panic here? Perhaps an error message and continue on with the rest of the migration? (I'd hate to see the migration get almost to the end and then fail.)

migrations/back-37/Hasher.go
darinkrauss: Plus, if you have a value of `baseData` here, then you can simply append to it in the switch and then pull out the `makeHash` below the switch and DRY up the code.

migrations/back-37/back-37.go
```diff
@@ -0,0 +1,207 @@
...
+				"_active": true,
+				// uploads aren't de-duped.
+				"type":          bson.M{"$ne": "upload"},
+				"_deduplicator": bson.M{"$exists": false},
+			}
+
+			var jellyfishResult bson.M
+			jellyfishDocCursor, err := m.dataRepository.Find(context.Background(), selector)
+			if err != nil {
+				logger.WithError(err).Error("Unable to find jellyfish results")
```
darinkrauss, line 107: Presumably we'd want to go to the next user rather than keep going, yes?

migrations/back-37/back-37.go
```diff
@@ -0,0 +1,207 @@
...
+			logger.Debugf("Finding jellyfish records for user ID %s", userID)
+			selector := bson.M{
+				"_userId": userID,
+				"_active": true,
+				// uploads aren't de-duped.
+				"type":          bson.M{"$ne": "upload"},
+				"_deduplicator": bson.M{"$exists": false},
+			}
+
+			var jellyfishResult bson.M
```
darinkrauss, line 104: This can be pushed inside the `for` loop below to narrow the scope.

migrations/back-37/back-37.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func (m *Migration) migrateJellyfishDocuments() (int, int, int) {
+	logger := m.Logger()
+
+	logger.Debug("Finding distinct users")
+
+	var hashUpdatedCount, archivedCount, errorCount int
+
+	userIDs, err := m.dataRepository.Distinct(context.Background(), "_userId", bson.M{})
```
darinkrauss, line 88: I don't remember, is there an index on `_userId`? I'm pretty sure there is. If not, this would have to do a scan through all records.

darinkrauss: Also, consider adding the same selector as below so we are only getting distinct user ids for those users with the selector. Specifically:
```
			selector := bson.M{
				"_active": true,
				// uploads aren't de-duped.
				"type":          bson.M{"$ne": "upload"},
				"_deduplicator": bson.M{"$exists": false},
			}
```

migrations/back-37/back-37.go
```diff
@@ -0,0 +1,207 @@
...
+	} else {
+		logger.Debugf("Finding jellyfish records for %d users", len(userIDs))
+
+		for _, userID := range userIDs {
+			logger.Debugf("Finding jellyfish records for user ID %s", userID)
+			selector := bson.M{
+				"_userId": userID,
+				"_active": true,
+				// uploads aren't de-duped.
+				"type":          bson.M{"$ne": "upload"},
```
darinkrauss, line 100: Are `upload` records in the data collection now? I thought they were moved.

darinkrauss: Or are they still around, but just not used in the data collection?

migrations/back-37/back-37.go
```diff
@@ -0,0 +1,207 @@
...
+						}
+					}
+					hashUpdatedCount++
+				}
+			}
+			if err := jellyfishDocCursor.Err(); err != nil {
+				logger.WithError(err).Error("error while fetching data. Please re-run to complete the migration.")
+				errorCount++
+			}
+			jellyfishDocCursor.Close(context.Background())
```
darinkrauss, line 151: Consider moving this as a `defer` just below its creation.

migrations/back-37/back-37.go
```diff
@@ -0,0 +1,207 @@
...
+				if err != nil {
+					logger.WithError(err).Errorf("Could not query for duplicate datum %s.", jellyfishResult["_id"])
+					errorCount++
+					continue
+				}
+				if dupCursor.Next(context.Background()) {
+					if !m.DryRun() {
+						if err := m.archiveDocument(jellyfishResult["_id"]); err != nil {
+							logger.WithError(err).Error("Unable to archive jellyfish document")
+							errorCount++
```
darinkrauss, line 133: If it errored, then should it count add to the `archivedCount`? Same as below with `hashUpdatedCount`.

migrations/back-37/back-37.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	_, err := m.dataRepository.UpdateOne(context.Background(), bson.M{"_id": jfDatum["_id"]}, deduplicatorUpdate)
+	return err
+}
+
+func (m *Migration) archiveDocument(objectId interface{}) error {
+	archiveUpdate := bson.M{
+		"$set": bson.M{
+			"_active":       false,
+			"_archivedTime": time.Now().UnixNano() / int64(time.Millisecond),
```
darinkrauss, line 202: I believe this is "archivedTime", not "_archivedTime".

migrations/back-37/Hasher.go
darinkrauss: Also, even if the fields are there, you may want to do the same checks as in `Base.IdentifyFields()` to ensure there aren't any missing strings or zero times.

darinkrauss: Same down below in the `switch` statement.

migrations/back-37/back-37.go
```diff
@@ -0,0 +1,207 @@
...
+			// converted value as it is used to calculate the hash
+			val := jfDatum["value"].(float64)
+			mgdlVal := val*18.01559 + 0.5
+			mgdL := glucose.MgdL
+			jfDatum["value"] = glucose.NormalizeValueForUnits(&mgdlVal, &mgdL)
+
+			deduplicatorUpdate = bson.M{
+				"$set": bson.M{
+					"_deduplicator": bson.M{
+						"hash": CreateHash(jfDatum),
```
darinkrauss, line 179: Do we want to set the deduplicator name and version? Same below in `default` clause.

darinkrauss (changes requested):

migrations/back-37/back-37.go
darinkrauss: Ignore for now.

darinkrauss: Consider total number of distinct users in memory at the same time.

migrations/back-37/back-37.go
jh-bate: they have moved

jh-bate:
superseded by https://github.com/tidepool-org/platform/pull/665

PR #661, opened 2023-09-12
SNYK ignore known issue for 6 months

jh-bate:
see https://security.snyk.io/vuln/SNYK-GOLANG-GITHUBCOMDISINTEGRATIONIMAGING-5880692

in an effort to keep the build clean (or before we potential remove the imaging service) this ignores the above issue for now ... well for the next 6 mths anyway


toddkazakov (approved):
Safe to ignore - the service isn't used

PR #663, opened 2023-09-29
[BACK-2670] Run tasks in a separate worker pool for each task type

toddkazakov:


Roukoswarf (dismissed):
Looks good, needs some time in QA to be sure, but it looks right and I expect its fine.

Roukoswarf (approved):

PR #664, opened 2023-10-02
[BACK-2681] improvements to alert config endpoints

ewollesen:
Modifications to the alerts endpoints for safety and simplicity:

- Delete no longer reads a config from its body
- Upsert no longer reads UserID or FollowedID from its body
- Improved BSON storage format (documents are better nested/cleaner)
- Fixed capitalization of `enabled` field
- Added custom marshaling for `DurationMinutes`
- Clarified determination of UserID for alerts.Config objects

BACK-2681


toddkazakov (dismissed):

alerts/config.go
```diff
@@ -29,29 +41,33 @@ type Config struct {
...
 }
 
 // Base describes the minimum specifics of a desired alert.
 type Base struct {
 	// Enabled controls whether notifications should be sent for this alert.
-	Enabled bool
+	Enabled bool `json:"enabled"`
 	// Repeat is measured in minutes.
-	Repeat DurationMinutes `json:"repeat"`
+	Repeat DurationMinutes `json:"repeat,omitempty"`
```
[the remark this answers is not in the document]
ewollesen: I believe I understand what you're saying. However, in this case, the omitempty only affects the JSON representation. The BSON representation seen by Mongo does not have the omitempty flag, and so I believe should be OK here. I'll write a test to confirm.

That said, I hadn't considered the effect of omitempty on a Mongo upsert operation, so I should review the bson struct tags for that behavior in other places. A few tests would also be good, to prevent regressions or unexpected behavior.

ewollesen:
rebased onto master, for real this time

/deploy qa2 data

alerts/config.go
```diff
@@ -5,116 +5,179 @@ package alerts
...
 	// be triggered.
 	Threshold `json:"threshold"`
 }
 
-func (t ThresholdMixin) Validate(validator structure.Validator) {
-	t.Threshold.Validate(validator)
+func (d UrgentLowAlert) Validate(validator structure.Validator) {
+	d.Base.Validate(validator)
+	d.Threshold.Validate(validator)
+	validateThresholdInNativeUnits(validator, d.Threshold, 40, 55, glucose.MgdL)
```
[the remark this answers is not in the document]
darinkrauss: Also, as discuss in our meeting, consider validating in the threshold's declared type (rather than doing any transformation to validate).

alerts/config.go
```diff
@@ -5,116 +5,179 @@ package alerts
...
+type UrgentLowAlert struct {
+	Base `bson:",inline"`
 	// Threshold is compared the current value to determine if an alert should
 	// be triggered.
 	Threshold `json:"threshold"`
 }
 
-func (t ThresholdMixin) Validate(validator structure.Validator) {
-	t.Threshold.Validate(validator)
+func (d UrgentLowAlert) Validate(validator structure.Validator) {
```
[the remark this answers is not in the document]
darinkrauss: Same for the rest of the structs/functions.

alerts/config.go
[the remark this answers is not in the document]
darinkrauss: Also, consider constants for the `40` and `55` so the code becomes self-documenting.

alerts/config.go
```diff
@@ -5,116 +5,179 @@ package alerts
...
+func validateThresholdInNativeUnits(validator structure.Validator, threshold Threshold, min, max float64, units string) {
+	nativeMin := glucose.Convert(min, glucose.MgdL, threshold.Units)
+	nativeMax := glucose.Convert(max, glucose.MgdL, threshold.Units)
+	validator.Float64("threshold.value", &threshold.Value).InRange(nativeMin, nativeMax)
+}
+
+// NotLoopingAlert extends Base with a delay.
+type NotLoopingAlert struct {
+	Base  `bson:",inline"`
+	Delay DurationMinutes `json:"delay,omitempty"`
```
darinkrauss, line 128: I don't have context into any previous discussions, what are we trying to accomplish with a custom `DurationMinutes` type here, rather than just an `*int`?

alerts/config.go
```diff
@@ -5,116 +5,179 @@ package alerts
...
 }
 
-func (d WithThreshold) Validate(validator structure.Validator) {
+func (d NotLoopingAlert) Validate(validator structure.Validator) {
 	d.Base.Validate(validator)
-	d.ThresholdMixin.Validate(validator)
+	dur := d.Delay.Duration()
+	validator.Duration("delay", &dur).InRange(0, 2*time.Hour)
```
darinkrauss, line 134: [nit] Consider constants.

darinkrauss: Same for the rest of the code.

alerts/config.go
```diff
@@ -5,116 +5,179 @@ package alerts
...
+	validator.Duration("delay", &dur).InRange(0, 6*time.Hour)
+}
+
+// LowAlert extends Base with threshold and a delay.
+type LowAlert struct {
+	Base `bson:",inline"`
+	// Threshold is compared the current value to determine if an alert should
+	// be triggered.
+	Threshold `json:"threshold"`
+	Delay     DurationMinutes `json:"delay,omitempty"`
```
darinkrauss, line 155: Not sure if this got turned off or is no longer used, but is the code formatter working? IIRC, I think the struct tags should be aligned here. Perhaps it is no longer a thing.

data/service/api/v1/alerts.go
```diff
@@ -31,23 +30,25 @@ func DeleteAlert(dCtx service.Context) {
 	details := request.DetailsFromContext(ctx)
 	repo := dCtx.AlertsRepository()
 
-	if err := checkAuthentication(details, r.PathParam("userID")); err != nil {
```
darinkrauss, line 34: Given that all three of these endpoints have the same path parameters (`userID` and `followedID`) could we DRY this up? Perhaps in a common authorization function? This might simplify some of the extra functions at the end of this file into one "validateAndAuthorize" function that might return the `userID`, `followedID`, and an error.

data/blood/glucose/glucose.go
```diff
@@ -60,3 +61,49 @@ func NormalizeValueForUnits(value *float64, units *string) *float64 {
 	}
 	return value
 }
+
+const (
+	// MgdL_Per_MmolL renames MmolLToMgdLConversionFactor in an alternate way.
+	MgdL_Per_MmolL = MmolLToMgdLConversionFactor
```
darinkrauss, line 67: I know, underscores make it more readable, but we lose any consistency we have with the rest of the code. Something to consider.

darinkrauss: Also, why a new name?

data/blood/glucose/glucose.go
```diff
@@ -60,3 +61,49 @@ func NormalizeValueForUnits(value *float64, units *string) *float64 {
...
+// Convert between common blood glucose measurement units.
+//
+// Results are rounded to five decimal places. However, if fromUnits and
+// toUnits are the same, the value is returned unchanged. If either fromUnits
+// or toUnits aren't handled, Convert panics.
+//
+// Convert is modeled to return results as similar as possible to
+// NormalizeValueForUnits, but without the hassle of pointer values, and with
+// the ability to convert between arbitrary units (as supported).
+func Convert(value float64, fromUnits, toUnits string) float64 {
```
darinkrauss, line 81: [nit] If the function has multiple different types, then I prefer to be explicit about argument for make it easier to determine the type for each parameter (without having to use extra brain power to say "wait, there is no type specified here, okay, then keep reading until we find a type"). I know, extreme nit. :)

data/blood/glucose/glucose.go
```diff
@@ -60,3 +61,49 @@ func NormalizeValueForUnits(value *float64, units *string) *float64 {
...
+	if from == to {
+		return value
+	}
+	if from == MgdL && to == MmolL {
+		return Round5(value * MmolL_Per_MgdL)
+	}
+	if from == MmolL && to == MgdL {
+		return Round5(value * MgdL_Per_MmolL)
+	}
+	panic(fmt.Errorf("unhandled conversion %q => %q", fromUnits, toUnits))
```
darinkrauss, line 92: Please do not use `panic` within the code. (Outside of test code it is used in only a very, very few specific uses cases.) Return an error and, if possible and appropriate, log.

data/blood/glucose/glucose.go
```diff
@@ -60,3 +61,49 @@ func NormalizeValueForUnits(value *float64, units *string) *float64 {
...
+func Convert(value float64, fromUnits, toUnits string) float64 {
+	from, to := normalizeUnitsCase(fromUnits), normalizeUnitsCase(toUnits)
+	if from == to {
+		return value
+	}
+	if from == MgdL && to == MmolL {
+		return Round5(value * MmolL_Per_MgdL)
+	}
+	if from == MmolL && to == MgdL {
+		return Round5(value * MgdL_Per_MmolL)
```
darinkrauss, line 90: I'm not sure this is going to do what you want. You are rounding a mg/dL value to 5 when only mmol/L should be rounded to 5. This is why I think we should stick with original values and use explicit limits for both mg/dL and mmol/L rather than doing conversion first to apply the limit to a common unit.

See problem here. https://go.dev/play/p/qRTDdyktgq8

darinkrauss (commented):
@ewollesen Take most of these comment with a grain of salt. I don't have a good handle on the latest code standards/conventions so a lot of my comments go back to when I last worked on the Tidepool backend. Only one or two actual code concerns.

data/service/api/v1/alerts.go
```diff
@@ -20,8 +19,8 @@ import (
 func AlertsRoutes() []service.Route {
 	return []service.Route{
 		service.MakeRoute("GET", "/v1/alerts/:userID/:followedID", Authenticate(GetAlert)),
-		service.MakeRoute("POST", "/v1/alerts/:userID", Authenticate(UpsertAlert)),
-		service.MakeRoute("DELETE", "/v1/alerts/:userID", Authenticate(DeleteAlert)),
+		service.MakeRoute("POST", "/v1/alerts/:userID/:followedID", Authenticate(UpsertAlert)),
```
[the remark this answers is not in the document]
ewollesen: It is, yes. It is the user id of person being followed. Or to put it another way, it is the user id of the person sharing their data.

I had originally named it "invitorID", but after some discussion with Todd, we changed it to "followedID". You can peruse that [history here](https://github.com/tidepool-org/platform/pull/657#discussion_r1331482131).

alerts/config.go
```diff
@@ -5,116 +5,179 @@ package alerts
...
 
 func (c Config) Validate(validator structure.Validator) {
 	validator.String("UserID", &c.UserID).Using(user.IDValidator)
 	validator.String("FollowedID", &c.FollowedID).Using(user.IDValidator)
-	if c.Low != nil {
-		c.Low.Validate(validator)
+	c.Alerts.Validate(validator)
+}
+
+func (i Alerts) Validate(validator structure.Validator) {
```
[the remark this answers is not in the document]
ewollesen: Ooh, yes, this is a typo. Thanks.

alerts/config.go
```diff
@@ -5,116 +5,179 @@ package alerts
...
+	if i.Low != nil {
```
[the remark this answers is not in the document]
ewollesen: I will try to re-order them that way, barring any dependency issues.

alerts/config.go
```diff
@@ -5,116 +5,179 @@ package alerts
...
+	if i.NoCommunication != nil {
+		i.NoCommunication.Validate(validator)
 	}
 }
 
 // Base describes the minimum specifics of a desired alert.
 type Base struct {
 	// Enabled controls whether notifications should be sent for this alert.
-	Enabled bool
+	Enabled bool `json:"enabled"`
```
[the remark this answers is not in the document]
ewollesen: It is, however the defaults are just fine. Adding a tag would seem to make the code unnecessarily more complicated/harder to read, IMO. This isn't a strongly held opinion on my part though, so if it still bugs you, let me know, and I'll add the tag.

alerts/config.go
```diff
@@ -5,116 +5,179 @@ package alerts
...
+	RepeatMin = 15 * time.Minute
+	// RepeatMax is the maximum duration for a repeat setting.
+	RepeatMax = 4 * time.Hour
+)
 
-func (d DelayMixin) Validate(validator structure.Validator) {
-	dur := d.Delay.Duration()
-	validator.Duration("delay", &dur).GreaterThan(0 * time.Minute)
+func validateRepeat(value time.Duration, errorReporter structure.ErrorReporter) {
+	if value == 0 {
```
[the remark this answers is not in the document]
ewollesen: Sure, I can do that. I generally don't find it useful to make constants for Zero values in Go, as 99% of the time it's either a default or an invalid value. I'm just trained to read Go code that way, and I find it removes a bit of cruft. Since that's not a shared experience, I can define a const.

data/service/api/v1/alerts.go
ewollesen: I've studied this a bit, and even created the function in question. What I came up with, looked like this:

```go
// checkAlertsPrerequisites handles HTTP administration tasks.
//
// Such as authentication, authorization, and consistency checks.
func checkAlertsPrerequisites(dCtx service.Context) (string, string, *platform.Error) {
	r := dCtx.Request()
	ctx := r.Context()
	details := request.DetailsFromContext(ctx)

	if err := checkAuthentication(details); err != nil {
		return "", "", platform.ErrorUnauthorized()
	}

	followedID := r.PathParam("followedID")
	userID := userIDWithServiceFallback(details, r.PathParam("userID"))
	pc := dCtx.PermissionClient()
	if err := checkUserAuthorization(ctx, pc, userID, followedID); err != nil {
		return "", "", platform.ErrorUnauthorized()
	}

	if err := checkUserIDConsistency(details, r); err != nil {
		return "", "", platform.ErrorUnauthorized()
	}

	return userID, followedID, nil
}
```
This function is responsible for a lot of things. Namely:

1. authenticating the request
2. authorizing the user
3. checking the consistency of the userIDs

To do that, it has knowledge of:

1. the data package's `Context`
2. the rest package's `Request`
3. the platform package's `Context`
4. the platform package's `Error`
5. the request's path structure

That seems like a lot for what's supposed to be a helper function. It would also be suspect to increased complexity if a future handler with slightly different requirements were to be added.[^1]

I would prefer to keep things simpler, even if it means repeating myself. 

The three existing helpers (`checkUserAuthorization`, `checkAuthentication`, and `checkUserIDConsistency`) know only a small part of all of that which goes into a handler function. As a result, they're easier to read and reason about. They're easier to write tests that exercise them as well, as there's less setup required, and fewer internal logical paths to trace.

P.S. It's true that the handler functions (`GetAlert`, `DeleteAlert`, and `UpsertAlert`) themselves know about all of the things that `checkAlertsPrerequisites` would, but they have no choice in the matter, that's complexity inherent to how HTTP handlers in the platform package are written. If I can contain/limit that complexity to just those handlers, keeping the helper functions smaller, simpler, and focused on a single responsibility, then that strikes me as a good choice.

1. The most straightforward path would be to modify the `checkAlertsPrerequisites` function to handle the new needs, but that would make an already complex function more so, and perhaps even worse, it could easily have unforeseen effects on the behavior of the other handler functions. Adding a separate `checkAlertsPrequisites2` of some sort, just makes things more complex as well.

data/blood/glucose/glucose.go
ewollesen: > I know, underscores make it more readable, but we lose any consistency we have with the rest of the code. Something to consider.

Yes, I generally hate underscores in Go names. This was an intentional choice for readability. As I see it, since this is important code, readability trumps convention. For that matter, I personally _don't_ find that `AThingWithALongName` is less readable than `A_Thing_With_A_Long_Name`, but in this specific case, with the MgdL and MmolL, I found the underscores clearer.

> Also, why a new name?

I thought it was more readable. It took me a while, and several readings of the code, to figure out how to use `MmolLToMgdLConversionFactor` correctly. Assuming I have MmolL, do I multiply or divide to get MgdL? It's not clear to me from the name. 

With MgdL_Per_MmolL, from the name alone I know that this value is a ratio of:
```
  Mg/dL
  -----
  Mmol/L
```
So then, if I have MmolL, and want to get MgdL, I know that I can multiply my value by MgdL_Per_MmolL and the MmolL will cancel out, leaving me with MgdL.

This could all be just a result of how my brain works, or how I learned to do these conversion in school or something, I dunno. If you don't find the new names more readable, I can leave them out.

data/service/api/v1/alerts.go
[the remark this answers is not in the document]
darinkrauss: Consider `followedUserID` to make it clear it is a user id that is being followed.

alerts/config.go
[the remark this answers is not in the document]
darinkrauss: In this past, I used the struct tags as an indicator of where it can be serialized to. So, if the BSON tags were present it was a message that my future self to remember that this struct ends up serialized in BSON.

While it's probably obvious, many of my suggestions are based upon habits I built up over time to help out "future" me have an easier time picking up old code again and preventing future mistakes.

alerts/config.go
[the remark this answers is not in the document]
darinkrauss: Yea, that makes sense where using 0 (or the zero-value) is obvious and it is not being used as a "special value".

In this case I don't think it is obvious what 0 actually means, at least to me. Since 0 is the zero-value for a `time.Duration`, if the field is missing in JSON, does it mean "use a system-defined default value to repeat after", "never repeat", or something else? 

For me to know the answer I have to scan backwards through the code and find the comment `// A value of 0 (the default) disables repeat notifications.` However, if the constant was something like `RepeatDurationNeverRepeat` or `RepeatDurationDisabled` then the code becomes self-documenting and a separate manual comment becomes less critical for understanding. (Oh, yes, before I forget, I suck at naming things, but you get the idea. ;) )

Alternatively, you could be completely clear by using a separate boolean to indicate if repeating was enabled/disabled. That makes it even clearer, but at the expense of a bit of complexity.

data/service/api/v1/alerts.go
darinkrauss: Okay, I see your point. That does make it more complex. Another option you could consider is adding much of the code as a middleware function, just like the `Authenticate` middleware used in `AlertRoutes`.

One of the reasons I mentioned this was that I noticed that the "authentication/authorization/validation" code is different for each of `DeleteAlert`, `GetAlert`, and `UpsertAlert`, but it seems like it should be all the same. For example, in `GetAlert` the check for `checkUserIDConsistency` is after the `checkUserAuthorization` (which we really want last since it is the most expensive operation). Also, the `checkUserAuthorization` is after `DecodeRequestBody` in `UpsertAlert` (though one could argue that validating the request body should go before all authentication/authorization since it is validating the incoming request before trying to do anything with it).

data/blood/glucose/glucose.go
darinkrauss: Well, I'll leave that up to you to decide. However, consider this from a perspective of a future engineer looking at this code and seeing two constants referring to the same value. Future me would ask "why two constants here"?

I agree that `MmolLToMgdLConversionFactor` is poorly named. I told you I suck at naming things. ;)

Fortunately, it looks like `MmolLToMgdLConversionFactor` (and `MmolLToMgdLPrecisionFactor`) is only used in `glucose.go` and `glucose_test.go`, so consider just renaming it to something more descriptive. This will directly address the root of the problem (a poorly named constant). :100:

data/service/api/v1/alerts.go
ewollesen: The middleware idea is tempting, but I prefer to have middleware deal with the concerns of HTTP (or whatever transport layer). Keeping the transport layers and business logic layers separate yields benefits in terms of clearer logic and easier testing. 

The orderings of the helpers do differ a little, and most of those differences were guesses at an performance optimization that I've never measured, but do believe with probability 0.99 don't matter anyway :). So I think they could all be done in the same order and it's very likely that no one would ever know the difference.

> Also, the checkUserAuthorization is after DecodeRequestBody in UpsertAlert (though one could argue that validating the request body should go before all authentication/authorization since it is validating the incoming request before trying to do anything with it).

I wouldn't necessarily agree with that. Though the argument is largely academic, as a successful request needs to perform all checks and operations anyway, so the order doesn't matter.

But to continue with the thought experiment... I would generally want to perform the operations in shortest time to completion order, as that lets me abort at the soonest opportunity, thereby optimizing for minimal resource consumption. I generally think of json un/marshaling as being relatively slow, due to the use of reflect and the amount of object allocation involved.

In this case, the permission check is slower than typical, as it requires a round trip to the gateway service. I don't have a good feel for how it compares to json un/marshaling.

I'd also consider that in the event that a request fails the authentication or authorization steps, it increases the likelihood that the request body is either malformed (or even malicious). So leaving the expensive json un/marshaling operation for last seems like a good idea.

data/blood/glucose/glucose.go
ewollesen: I'm not a fan of having two names for the same thing at all, but I was unwilling to touch _anything_ in the other functions, given the sensitivity of the method. In hindsight, that was silly, and it's better to make those changes (if I think they're warranted) and then call out dangerous areas to code reviewers for careful review.

If you're not opposed to `MmolLToMgdLConversionFactor` being renamed to `MgdL_Per_MmolL`, I would gladly make that switch and collapse them back down to 1 constant.

Naming is hard, I won't deny it, or pretend that I'm somehow better than anyone else at it. I can only explain my reasoning and hope it resonates with future readers.

alerts/config.go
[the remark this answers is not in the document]
ewollesen: > In this past, I used the struct tags as an indicator of where it can be serialized to. So, if the BSON tags were present it was a message that my future self to remember that this struct ends up serialized in BSON.

OK. I'll add them. FWIW, my previous self screams never trust that, probably for reasons similar too, but opposite of yours. :)

data/blood/glucose/glucose.go
ewollesen: Can you go deeper into what's wrong here? The fact that you don't get exactly the same value back is an expected behavior of unit conversions with limited precision. Is there some other problem with the logic?

If you take it one step further, and go back to mmol/L again, you'll see you get the same answer as the first time:

https://go.dev/play/p/rqGmk8Ajw25

So it does enter a steady state. The conversions aren't 1:1 by their nature, so you can never have perfect bi-directional conversion. That's why the choice of precision is important, it's saying that if there are any differences in conversion, at this precision, they're small enough that they don't matter.

Or am I misunderstanding something else about your comment?

ewollesen: As a followup, in terms of significant digits (6), if we round the result to 6 significant digits, it becomes 140.000, and that works out.

alerts/config.go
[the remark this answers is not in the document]
ewollesen: > In this case I don't think it is obvious what 0 actually means, at least to me. Since 0 is the zero-value for a `time.Duration`, if the field is missing in JSON, does it mean "use a system-defined default value to repeat after", "never repeat", or something else?

That's interesting to me. When I read this code, 0 is very clearly "don't repeat". I can't even imagine what else it could mean. But I have no problem with adding a const. Brains are strange things. :)

data/blood/glucose/glucose.go
ewollesen: This was poor choice on my part, you're right.

data/blood/glucose/glucose.go
ewollesen: Either way, given our discussion with @ps2 today, this code will be removed, so no worries.

alerts/config.go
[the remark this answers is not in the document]
darinkrauss: Yea, I hear you. I suppose if we aren't doing it consistently then it is less-than-helpful.

data/service/api/v1/alerts.go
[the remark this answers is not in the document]
darinkrauss: Yea, that would have been clearer if implemented at the beginning, but we didn't think about that. At this point it would be a very large change to a large chunk of code, so probably something we should avoid.

data/store/mongo/mongo_alerts.go
```diff
@@ -54,18 +54,18 @@ func (r *alertsRepo) EnsureIndexes() error {
...
+				SetName("UserIdFollowedUserIdTypeUnique"),
 		},
 	})
 }
 
 func (r *alertsRepo) filter(cfg *alerts.Config) interface{} {
 	return &alerts.Config{
-		UserID:     cfg.UserID,
-		FollowedID: cfg.FollowedID,
+		UserID:         cfg.UserID,
```
darinkrauss, line 68: Have we deployed this code anywhere? That is, have the previous indexes already been created and/or is the data stored with `followedId` anywhere? If so, we may have "bad data" for those environments. Not usually a big deal for QA environments, but something to be aware of.

alerts/config.go
```diff
@@ -5,116 +5,168 @@ package alerts
...
+	// be triggered.
+	Threshold `json:"threshold"`
+	Delay     DurationMinutes `json:"delay,omitempty"`
 }
 
-func (d WithDelayAndThreshold) Validate(validator structure.Validator) {
-	d.Base.Validate(validator)
-	d.DelayMixin.Validate(validator)
-	d.ThresholdMixin.Validate(validator)
+func (a HighAlert) Validate(validator structure.Validator) {
```
darinkrauss, line 165: Just pointing out that a number of struct functions don't have the lowercased first letter of the struct name as the receiver name. (You indicated that an earlier one was a typo.) Not a big deal, but just making you aware. Okay if you don't want to change them now. (If you do, though, go through all of the structs in the PR and update them all.)

data/service/api/v1/alerts.go
darinkrauss: While I don't think a code change is required, I'm going to have to disagree on "Though the argument is largely academic, as a successful request needs to perform all checks and operations anyway, so the order doesn't matter."

In most cases, I validate input *before* doing anything else. There are a couple of reasons for this:
- If the input is invalid, why bother doing anything else or invoking any other microservices or the database? (If a bad actor is trying to hack the API, then catching invalid input before impacting another microservice/database is a good thing. Think of a DDOS attack with bad input - would you rather have just this microservice impacting by the attack or this microservice and the permissions microservice and/or the database?)
- Unexpected consequences can occur if any input is used prior to its validation.

With both of the above, I try to validate all input first as a matter of habit.

Either way, its up to you.

darinkrauss (approved):
A couple additional comments/questions, but nothing that is blocking approval.

alerts/config.go
```diff
@@ -135,6 +187,11 @@ func (m *DurationMinutes) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+func (m *DurationMinutes) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
ewollesen: Yes, that's fine. The docs call for 1-minute granularity, and well-behaved clients should be adhering to that. Those that send other values can live with the poorly-defined behavior :)

Or it can perhaps be filed under "be broad in what you accept, and strict in what you emit" if you prefer. :)

alerts/config.go
```diff
@@ -151,8 +208,25 @@ type ValueWithUnits struct {
...
+func (t Threshold) Validate(v structure.Validator) {
+	v.String("units", &t.Units).OneOf(glucose.MgdL, glucose.MmolL)
+	// This is a sanity check. Client software will likely further constrain these values. The
+	// broadness of these values allows clients to change their own min and max values
+	// independently, and it sidesteps rounding and conversion conflicts between the backend and
+	// clients.
+	var max, min float64
+	switch t.Units {
+	case glucose.MgdL, glucose.Mgdl:
+		max = glucose.MgdLMaximum
```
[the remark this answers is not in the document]
ewollesen: Heh, no good reason.

alerts/config.go
ewollesen: Oh, my pattern is to use the first letter of the noun. So:

```
func (c *httpClient) Get() {}
```

So the `a` there is intentional on my part.

alerts/config.go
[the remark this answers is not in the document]
darinkrauss: Sure. I just wanted to make you aware in case some downstream operation was generated based upon that initial incoming data that was in fractional minutes. 

For example, if upon initial create operation (immediately after `UnmarshalJSON` of fractional minutes), we created a some other data (e.g. a cron job) using those fractional minutes (e.g. the cron job would fire off every second). 

If we don't do this (and won't in the future), then it isn't an issue.

PR #671, opened 2023-10-17
rename a function and a struct for clarity

ewollesen:
My motivation with this change is to prevent a potential authentication error wherein someone accidentally breaks assumptions about the authentication process due to a misunderstanding about the purpose of the Details struct.

So let's get more concrete. The Authenticate function in data/service/api/v1/authenticate_middleware.go was being used to check that authentication was previously performed. The Authenticate function itself did no actual authentication. Intead, it checked for the existence of a Details value in the request's Context. If a Details value was found, then it was assumed that the request was previously authenticated. All of that is OK, as long as the Details object is only added to a request Context after successful authentication.

The problem I'm trying to avoid with this renaming, is some future me (or you!) coming along with some piece of data that needs to be passed to a request handler, and I see this handy Details value that gets passed via the Context, and I think "I'll just add it into this Details thing. Easy-peasey. All the while not realizing that the mere existence of the Details value indicates to request handlers that the request has been authenticated.

This renaming can't prevent the scenario above from happening, but it should make it obvious that these values, now called AuthDetails, are specifically for authentication information. Hopefully that's enough of a warning to scare future me away from touching them for non-authentication purposes.


toddkazakov (approved):

ewollesen:
Closed in favor of #672

PR #673, opened 2023-10-19
[BACK-2702] Add reason for setting last updated and outdated since

toddkazakov:
Adds a field which stores the event that trigerred the summary update. This will allow clinic-worker to conditionally trigger EHR flows if the summary was updated after a dataset was closed.


data/datum.go
```diff
@@ -21,6 +21,7 @@ type Datum interface {
 	GetPayload() *metadata.Metadata
 
 	GetType() string
+	GetActive() bool
```
toddkazakov, line 24: When the function returns a boolean it's better to prefix the attribute with `Is` instead of `Get` so it reads naturally when you use it in an if statement - e.g. `if datum.IsActive() { ... }`

data/service/api/v1/datasets_data_create.go
```diff
@@ -129,13 +129,13 @@ func DataSetsDataCreate(dataServiceContext dataService.Context) {
 }
 
 // The following 2 functions are pulled out of the above so that they can be unit tested independently.
-func MaybeUpdateSummary(ctx context.Context, registry *summary.SummarizerRegistry, updatesSummary map[string]struct{}, userId string) map[string]*time.Time {
+func MaybeUpdateSummary(ctx context.Context, registry *summary.SummarizerRegistry, updatesSummary map[string]struct{}, userId, reason string) map[string]*time.Time {
```
toddkazakov, line 132: This function should not be defined in `datasets_data_create.go` - it's better if it's located in the summary package (as it's also used when updating a dataset).

data/summary/types/summary.go
```diff
@@ -177,14 +172,22 @@ func NewConfig() Config {
...
 
-func (s *Summary[T, A]) SetOutdated() {
+func (s *Summary[T, A]) SetOutdated(reason string) {
+	set := mapset.NewSet[string](reason)
+	if len(s.Dates.OutdatedReason) > 0 {
+		set.Append(s.Dates.LastUpdatedReason...)
+	}
+
+	s.Dates.LastUpdatedReason = nil
+	s.Dates.OutdatedReason = set.ToSlice()
```
ewollesen, line 182: Is this worth bringing in an external dependency? The golang-set module looks like a nice module, I'll grant. But this could just be:

```go
func (s *Summary[T, A]) SetOutdated(reason string) {
	set := map[string]struct{}{}
	uniq := []string{}
	for _, r := range s.Dates.OutdatedReason {
		if _, found := set[r]; !found {
			set[r] = struct{}{}
			uniq = append(uniq, r)
		}
	}

	s.Dates.LastUpdatedReason = nil
	s.Dates.OutdatedReason = uniq
	s.Dates.OutdatedSince = pointer.FromAny(time.Now().Add(setOutdatedBuffer).UTC().Truncate(time.Millisecond))
	s.Dates.HasOutdatedSince = true
}
```
You could fix up my poor names a bit, and even extract it to a function if you want, but the point is, for such a simple set operation, I don't think bringing in an external dependency is worth it.

toddkazakov: I'm aware that I can use a map to track unique elements, but I find the version with the external dependency much easier to read and understand.

data/datum.go
Roukoswarf: Debated this while I was writing it, as SetActive's opposite should be GetActive, even if it makes very little wording sense.
But happy to change it with a vote to weigh it down the other way, done.

data/summary/types/summary.go
Roukoswarf: This is such a common pattern we keep having all over the place, I for one welcome this external dependency, we can remove more than just the pattern here, but in a few other places in summary code next pass.

data/service/api/v1/datasets_data_create.go
Roukoswarf: done.

data/service/api/v1/datasets_data_create.go
```diff
@@ -158,9 +158,13 @@ func CheckDatumUpdatesSummary(updatesSummary map[string]struct{}, datum data.Dat
 	twoYearsPast := time.Now().UTC().AddDate(0, -24, 0)
 	oneDayFuture := time.Now().UTC().AddDate(0, 0, 1)
 
-	for _, typ := range types.DeviceDataTypes {
-		if datum.GetType() == typ && datum.GetTime().Before(oneDayFuture) && datum.GetTime().After(twoYearsPast) {
-			updatesSummary[types.DeviceDataToSummaryTypes[typ]] = struct{}{}
+	// we only update summaries if the data is both of a relevant type, and being uploaded as "active"
+	// it also must be recent enough, within the past 2 years, and no more than 1d into the future
+	if datum.GetActive() {
+		for _, typ := range types.DeviceDataTypes {
```
toddkazakov, line 137: Consider making types.DeviceDataTypes a set, so we can avoid the quadratic complexity here.

data/service/api/v1/datasets_update.go
```diff
@@ -91,17 +93,27 @@ func DataSetsUpdate(dataServiceContext dataService.Context) {
...
-	MaybeUpdateSummary(ctx, dataServiceContext.SummarizerRegistry(), all, *dataSet.UserID, types.OutdatedReasonUploadCompleted)
 
 	if err = dataServiceContext.MetricClient().RecordMetric(ctx, "data_sets_update"); err != nil {
 		lgr.WithError(err).Error("Unable to record metric")
 	}
 
 	dataServiceContext.RespondWithStatusAndData(http.StatusOK, dataSet)
 }
+
+func CheckDataSetUpdatesSummary(ctx context.Context, repository store.DataRepository, updatesSummary map[string]struct{}, dataSetID string) {
```
toddkazakov, line 111: This looks like business logic to me - doesn't belong to the api package.

data/store/mongo/mongo_datum.go
```diff
@@ -543,6 +545,30 @@ func validateAndTranslateSelectors(selectors *data.Selectors) (bson.M, error) {
...
 
+func (d *DatumRepository) CheckDataSetContainsType(ctx context.Context, dataSetID string, typ string) (bool, error) {
+	twoYearsPast := time.Now().UTC().AddDate(0, -24, 0)
+	oneDayFuture := time.Now().UTC().AddDate(0, 0, 1)
+
+	selector := bson.M{
+		"_active":  true,
+		"uploadId": dataSetID,
+		"type":     typ,
+		"time": bson.M{"$gt": twoYearsPast,
```
toddkazakov: This query doesn't do exactly what the function name says it does - consider passing the time bounds as a parameter

data/summary/store/store.go
```diff
@@ -238,12 +238,12 @@ func (r *Repo[T, A]) GetOutdatedUserIDs(ctx context.Context, page *page.Paginati
 	if errors.Is(err, mongo.ErrNoDocuments) {
 		return nil, nil
 	} else if err != nil {
-		return nil, errors.Wrap(err, "unable to get outdated summaries")
+		return nil, fmt.Errorf("unable to get outdated summariesL %w", err)
```
lostlevels, line 241: `summariesL` typo

data/summary/types/summary.go
```diff
@@ -72,7 +79,8 @@ type Config struct {
 }
 
 type Dates struct {
-	LastUpdatedDate time.Time `json:"lastUpdatedDate" bson:"lastUpdatedDate"`
+	LastUpdatedDate   time.Time `json:"lastUpdatedDate" bson:"lastUpdatedDate"`
```
lostlevels, line 84: For my understanding, is this always populated?

lostlevels (dismissed):
Just had the question and the typo.

data/summary/store/store.go
Roukoswarf: good catch, fixed.

data/summary/types/summary.go
Roukoswarf: yes, always. its created with a 0 value on backfill to push backfills to the front of the queue, but otherwise its always filled with real values, any Zero default values are intentional

lostlevels (dismissed):

lostlevels (approved):

PR #675, opened 2023-10-25
fix dexcom task retry logic

jh-bate:


dexcom/fetch/task.go
```diff
@@ -52,11 +52,12 @@ func FailTask(l log.Logger, t *task.Task, err error) error {
...
-			return count >= 3
-		}
+	if t.Data[dexcomTaskRetryField] == nil {
+		t.Data[dexcomTaskRetryField] = 0
+	}
+	count, ok := t.Data[dexcomTaskRetryField].(int)
+	if ok {
+		return count >= 3
 	}
 	return true
```
tjotala, line 62: Should we reset it to 0 in case it fails to parse? That way it would automatically recover from (what appears to be) bad data in the DB.

darinkrauss: A few thoughts:
- Consider logging this failure as an error so we can detect if/when it happens.
- Is there a non-linear fallback, just in case there is a bug in the code and this happens repeatedly?
- You might consider shortening this code to eliminate the extra parse by returning immediately after setting the field to 0.

jh-bate:
super seeded by later work

PR #677, opened 2023-10-27
[BACK-2702] Allow closing datasets created by jellyfish so we can recalculate summaries correctly

toddkazakov:
Currently, jellyfish doesn't close legacy datasets after upload completion. Due to this, we can't detect when the upload has finished. 

When a new batch is uploaded, jellyfish marks the summary as oudated with a 2 minute buffer, meaning the summary will be recalculated after 2 minutes. If an upload doesn't finish within this timeframe, the summary may be calculated mid-upload and may not be accurate. 

In order to fix the issue above and to reduce the time required to recalculate a summary after an upload has finished and the subsequent push of updated reports to the EHR, this PR allows Uploader to "close" jellyfish datasets (which currently don't have a deduplicator) so summaries can be recalculated right after upload. This is important for in-clinic uploads.

This may have some unexpected side-effects when https://github.com/tidepool-org/jellyfish/pull/195 is merged (e.g. running the deduplicator on jellyfish datasets).


data/service/api/v1/datasets_update.go
```diff
@@ -84,16 +84,20 @@ func DataSetsUpdate(dataServiceContext dataService.Context) {
...
-			return
-		} else if deduplicator == nil {
-			dataServiceContext.RespondWithInternalServerFailure("Deduplicator not found")
-			return
-		} else if err = deduplicator.Close(ctx, dataServiceContext.DataRepository(), dataSet); err != nil {
-			dataServiceContext.RespondWithInternalServerFailure("Unable to close", err)
-			return
+		// Allow closing datasets without deduplicator (i.e. those created by jellyfish)
+		// so we can mark summaries as outdated.
+		if dataSet != nil && dataSet.HasDeduplicatorName() {
```
darinkrauss, line 89: If I'm understanding things correctly, I don't think this will cause issues with the current Jellyfish migration/deprecation plan. The plan does not currently include any updates to the data set (just data). Thus, a Jellyfish data set will not have a deduplicator name and so this code will not be executed.

If, however, we do need to update the data set with the deduplicator name, then this will cause a couple of potential issues:

1. The Jellyfish data set and data may include some fields normally only present in Platform data (written by `DataRepository.UpdateDataSet` and `DataRepository.ActiveDataSetData`). We'd need to do a survey of exactly what fields are added by Jellyfish or Platform and compare the two. Even if there are field differences they may not cause any real issue, just inconsistencies.
2. The Platform code that performs data duplication will execute on Jellyfish data (where other Jellyfish data is duplicated based upon this Jellyfish data using the Platform deduplication mechanism). I don't *think* this will cause an issue, though, since Jellyfish will have already deduplicated the data (by dropping any new duplicates) prior to closing of the Jellyfish data set. Thus, I *think* that this data deduplication step would end up just being a NOP.

In either case, we should verify this definitely won't cause issues and add it as one or more test cases, but at first glance there doesn't appear to be any (or if there are some they would be minor).

cc: @jh-bate

darinkrauss (approved):
LGTM. I think we are okay with the Jellyfish migration/deprecation plan.

PR #679, opened 2023-11-01
[BACK-2737] adds an alerts client

ewollesen:
This client was part of [hydrophone-126](https://github.com/tidepool-org/hydrophone/pull/126), but is extracted out to platform. It makes more sense for it to be in platform, where it's service endpoints are defined.

BACK-2500

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,20 +12,22 @@ go_import_path: github.com/tidepool-org/platform
 env:
   global:
     - MONGODB=6.0.11
-    - MONGOSH=1.10.6
+    - MONGOSH=2.0.2
     - DIST=jammy
 
 before_install:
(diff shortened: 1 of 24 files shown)


.travis.yml
```diff
@@ -11,21 +11,23 @@ go_import_path: github.com/tidepool-org/platform
 
 env:
   global:
-    - MONGODB=6.0.11
-    - MONGOSH=1.10.6
+    - MONGODB=7.0.2
```
toddkazakov, line 14: This should match the version we use in production which is 6

toddkazakov (changes requested):

.travis.yml
ewollesen: Fixed

toddkazakov (approved):

PR #688, opened 2023-12-19
combined changes for jellyfish data migration

jh-bate:
[back-37](https://tidepool.atlassian.net/browse/BACK-37) - Migrate all existing data to add required Platform deduplication hash fields and ensure data adheres to platform standards
[back-2773](https://tidepool.atlassian.net/browse/BACK-2773) - Update platform to validate certain fields in pumpSettingsOverride and overridePresets
[back-2728](https://tidepool.atlassian.net/browse/BACK-2728) - Support tandem pumpSettings in the platform
[back-2520](https://tidepool.atlassian.net/browse/BACK-2520) - Data model changes to support Control-IQ sleep schedules


migrations/20231128_jellyfish_migration/utils/data_migration.go
```diff
@@ -0,0 +1,386 @@
...
+
+type DataMigrationQueryFn func(m *DataMigration) bool
+type DataMigrationUpdateFn func(m *DataMigration) (int, error)
+
+type groupedErrors map[string][]ErrorData
+
+func NewMigration(ctx context.Context, settings *Settings, checker MongoInstanceCheck, dataC *mongo.Collection, lastID *string) (*DataMigration, error) {
+	var err error
+	if settings == nil {
+		err = errors.Join(err, errors.New("missing required settings"))
```
darinkrauss, line 117: Why not just return the errors here rather than using a local variable?

darinkrauss: BTW, do we want to check `dataC`, too?

migrations/20231128_jellyfish_migration/utils/data_migration.go
```diff
@@ -0,0 +1,386 @@
...
+	ctx                  context.Context
+	dataC                *mongo.Collection
+	settings             *Settings
+	updates              []mongo.WriteModel
+	groupedDiffs         map[string][]UpdateData
+	groupedErrors        groupedErrors
+	errorsCount          int
+	updatedCount         int
+	fetchedCount         int
+	lastUpdatedID        string
```
darinkrauss, line 104: Suggest you just make this a `*string` to avoid any of the `pointer.String` stuff.

darinkrauss: Besides, when not using a pointer, the default is an empty string, which you have a special case to check for in `jellyfishQuery`. Just use a pointer and get rid of an empty string logic.

migrations/20231128_jellyfish_migration/utils/data_migration.go
```diff
@@ -0,0 +1,386 @@
...
+				log.Println(err)
+				os.Exit(1)
+			}
+			f.WriteString(string(errJSON) + "\n")
+		}
+		m.groupedErrors[group] = []ErrorData{}
+	}
+}
+
+func (m *DataMigration) writeAudit(groupLimit *int) {
```
darinkrauss, line 319: Is groupLimit actually used by any callers?

migrations/20231128_jellyfish_migration/utils/data_migration.go
```diff
@@ -0,0 +1,386 @@
...
+
+func (c MigrationStats) report() {
+	if c.Applied == 0 && c.Fetched > 0 {
+		log.Printf("elapsed [%s] for [%d] items fetched with [%d] errors\n", c.Elapsed, c.Fetched, c.Errored)
+		return
+	}
+	log.Printf("elapsed [%s] for [%d] fetched [%d] updates applied with [%d] errors\n", c.Elapsed, c.Fetched, c.Applied, c.Errored)
+}
+
+func (m *DataMigration) writeErrors(groupLimit *int) {
```
darinkrauss, line 291: Is groupLimit actually used by any callers?

migrations/20231128_jellyfish_migration/utils/data_migration.go
```diff
@@ -0,0 +1,386 @@
...
+			if len(errors) < *groupLimit {
+				continue
+			}
+		}
+		f, err := m.createFile("error", group, "%s.log")
+		if err != nil {
+			log.Println(err)
+			os.Exit(1)
+		}
+		defer f.Close()
```
darinkrauss, line 306: This will leave all of the files open until the function exits. It won't close each file after it is written to (which I think was your intention).

migrations/20231128_jellyfish_migration/utils/data_migration.go
```diff
@@ -0,0 +1,386 @@
...
+}
+
+func (m *DataMigration) GetLastID() string {
+	return m.lastUpdatedID
+}
+
+func (m *DataMigration) OnError(data ErrorData) {
+	m.errorsCount++
+	m.groupedErrors[data.ItemType] = append(m.groupedErrors[data.ItemType], data)
+	if m.settings.StopOnErr {
```
darinkrauss, line 267: @You can probably put this block at the top of the function.

migrations/20231128_jellyfish_migration/utils/data_migration.go
```diff
@@ -0,0 +1,386 @@
...
+}
+
+func (m *DataMigration) completeUpdates() bool {
+	capacity := m.settings.capacity
+	if capacity != nil {
+		stats := m.GetStats()
+
+		percent := (float64(stats.Fetched) * float64(100)) / float64(*capacity)
+
+		log.Printf("processed %.0f %% of %d records and applied %d changes", percent, *capacity, stats.Applied)
```
darinkrauss, line 227: Would we want a log message for when `capacity == nil`, too? Just without the capacity and percent?

migrations/20231128_jellyfish_migration/utils/data_migration.go
```diff
@@ -0,0 +1,386 @@
...
+	m.settings.WriteBatchSize = m.mongoInstanceChecker.GetWriteBatchSize()
+	return *m.settings
+}
+
+func (m *DataMigration) GetDataCollection() *mongo.Collection {
+	return m.dataC
+}
+
+func (m *DataMigration) Execute(
+	queryFn DataMigrationQueryFn,
```
darinkrauss, line 170: Using functions a bit harder to understand, but okay, I suppose. This is for testing right? Next time consider interfaces.

migrations/20231128_jellyfish_migration/utils/jellyfish_data.go
```diff
@@ -0,0 +1,165 @@
...
+
+func jellyfishQuery(settings Settings, userID *string, lastFetchedID *string) (bson.M, *options.FindOptions) {
+	selector := bson.M{
+		"_deduplicator": bson.M{"$exists": false},
+	}
+	if settings.Rollback {
+		selector = bson.M{
+			settings.RollbackSectionName: bson.M{"$exists": true},
+		}
+	}
```
darinkrauss, line 24:
```suggestion
	var selector bson.M
	if settings.Rollback {
		selector = bson.M{
			settings.RollbackSectionName: bson.M{"$exists": true},
		}
	} else {
		selector = bson.M{
  		    "_deduplicator": bson.M{"$exists": false},
		}
	}
```

migrations/20231128_jellyfish_migration/utils/jellyfish_data.go
```diff
@@ -0,0 +1,165 @@
...
+				if rollback, ok := item[settings.RollbackSectionName].(primitive.A); ok {
+					cmds := []bson.M{}
+					for _, cmd := range rollback {
+						if cmd, ok := cmd.(bson.M); ok {
+							cmds = append(cmds, cmd)
+						}
+					}
+					if len(cmds) > 0 {
+						m.SetUpdates(UpdateData{
+							Filter:   bson.M{"_id": itemID, "modifiedTime": item["modifiedTime"]},
```
darinkrauss, line 92: Modified time is so we don't stop on other changes happening simultaneously-ish?

darinkrauss: So, I guess if this happens ad an update is excluded because it changed by something else, we'd pick this up on the next go around?

darinkrauss (changes requested):
Okay, finally finished. Lots of comments with some probably needing a change and others maybe not.

jh-bate:
supersedded by #741

jh-bate:
superseded by https://github.com/tidepool-org/platform/pull/741

PR #690, opened 2024-01-03
Makefile: export TIMESTAMP to prevent skew

ewollesen:
Some travis builds have failed because TIMESTAMP experiences a skew. Since timestamp is only created in the one place, this should fix the issue.

https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html

This will probably wait until after @darinkrauss finishes his cleanup passes?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,7 @@
 TIMESTAMP ?= $(shell date +%s)
+# ensure that we use the same timestamps in sub-makes. We've seen cases where
+# these can vary by 1 second
+export TIMESTAMP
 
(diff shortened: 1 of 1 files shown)


tjotala (approved):
LGTM! :shipit:

darinkrauss (approved):
LGTM! Feel free to merge as you wish. I can pull it into the Dependabot updates the next time around.

darinkrauss:
@ewollesen I ended up not incorporating this into the Dependabot updates.

PR #691, opened 2024-01-03
[BACK-2416] Add read-only mode for services that doesn't create indexes.

lostlevels:
Adds prefix to read directly from auth service instead of going through the virtualservice.

Opening a new PR because rebasing / merging with current master was too messy with the recent master changes.


toddkazakov:
@lostlevels it would be great to have the capability to run a shadow environment for other testing purposes down the road. Let's get this tested in QA? Can you please come up with a test strategy?

PR #693, opened 2024-01-08
set CreatedUserID on device data set uploads.

ewollesen:


ewollesen:
Doing more spelunking, and I don't know that this will help anything. It's very unclear what's going on. We don't validate that it exists, but we also no longer set it (we used to, a LONG time ago).

toddkazakov (approved):

PR #694, opened 2024-01-09
add basic Dexcom API metrics

ewollesen:
The oauth.Client.SendOauthRequest method is modified to accept a
[]request.ResponseInspect. One of these inspectors generates prometheus
metrics calls to track Dexcom API usage by number of calls, path, and status
code.

This is done in response to use seeing some 429 (too many requests) errors in our logs. With the visibility these metrics provide, we hope to better understand our usage of the Dexcom API, so that we can take steps to reduce our usage if necessary, or at least have a meaningful discussion with Dexcom about how much we'd like to use their API.

BACK-2820

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -80,8 +80,8 @@ func (c *Client) RequestStreamWithHTTPClient(ctx context.Context, method string,
 
 	for _, inspector := range inspectors {
 		if err = inspector.InspectResponse(res); err != nil {
-			drainAndClose(res.Body)
-			return nil, err
+			log.LoggerFromContext(ctx).WithError(err).Warn("inspecting response")
+			continue
 		}
 	}
 
--- a/client/client.go
+++ b/client/client.go
@@ -79,17 +79,14 @@ func (c *Client) RequestStreamWithHTTPClient(ctx context.Context, method string,
 	}
 
 	for _, inspector := range inspectors {
-		if err = inspector.InspectResponse(res); err != nil {
-			log.LoggerFromContext(ctx).WithError(err).Warn("inspecting response")
-			continue
-		}
+		inspector.InspectResponse(res)
 	}
 
 	return c.handleResponse(ctx, res, req)
 }
 
 func (c *Client) RequestDataWithHTTPClient(ctx context.Context, method string, url string, mutators []request.RequestMutator, requestBody interface{}, responseBody interface{}, inspectors []request.ResponseInspector, httpClient *http.Client) error {
-	headerInspector := request.NewHeadersInspector()
+	headerInspector := request.NewHeadersInspector(log.LoggerFromContext(ctx))
 	body, err := c.RequestStreamWithHTTPClient(ctx, method, url, mutators, requestBody, append(inspectors, headerInspector), httpClient)
 	if err != nil {
 		return err
--- a/dexcom/client/client.go
+++ b/dexcom/client/client.go
@@ -132,13 +132,12 @@ func (c *Client) sendRequest(ctx context.Context, method, url string, mutators [
 type promDexcomInstrumentor struct{}
 
 // InspectResponse implements request.ResponseInspector.
-func (i *promDexcomInstrumentor) InspectResponse(r *http.Response) error {
+func (i *promDexcomInstrumentor) InspectResponse(r *http.Response) {
 	labels := prometheus.Labels{
 		"code": strconv.Itoa(r.StatusCode),
 		"path": r.Request.URL.Path,
 	}
 	promDexcomCounter.With(labels).Inc()
-	return nil
 }
 
 // promDexcomCounter instruments the Dexcom API paths and status codes called.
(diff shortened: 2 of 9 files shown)


dexcom/client/client.go
```diff
@@ -26,9 +29,6 @@ func New(cfg *client.Config, tknSrcSrc oauth.TokenSourceSource) (*Client, error)
 		return nil, err
 	}
 
-	// NOTE: Dexcom authorization server does not support HTTP Basic authentication
```
ewollesen, line 29: This is a no-op, found via staticcheck.

client/client.go
```diff
@@ -80,8 +80,8 @@ func (c *Client) RequestStreamWithHTTPClient(ctx context.Context, method string,
 
 	for _, inspector := range inspectors {
 		if err = inspector.InspectResponse(res); err != nil {
-			drainAndClose(res.Body)
-			return nil, err
+			log.LoggerFromContext(ctx).WithError(err).Warn("inspecting response")
```
toddkazakov, line 83: This change was the only red-flag for me in this PR, but I think it's ok, because `drainAndClose` is called in `handleResponse`. @darinkrauss is more familiar with the code than me, so I'll let him take a final a look.

ewollesen: Yeah, that change isn't strictly required, but as they're written before this, response inspectors that return errors, will cause the entire request to return an error, regardless of whether the request was otherwise successful. Any additional inspectors (like possibly the metrics inspector being added in this PR) might not run at all, causing inaccurate metrics.

darinkrauss: If we log the error (rather than failing) then the `drainAndClose` isn't necessary as it should be done in `handleResponse`.

Though logging (versus failing) changes the nature of the protocol, I don't see a problem with changing it since there isn't any previous usage.

darinkrauss: Oh, and the `continue` is unnecessary.

dexcom/client/client.go
darinkrauss: When this was first implemented this was required because the Dexcom API did not correctly conform to the OAuth2 spec and this allowed the OAuth library to work with Dexcom. (I don't remember exactly what was broken, though.) Since we've updated this dependency since then and the function has become a no-op, and the Dexcom API appears to still work, it appears we can remove it. (Though, I suppose, in theory, this could be the cause of some of the Dexcom failures.)

request/inspector.go
```diff
@@ -7,6 +7,17 @@ import (
 )
 
 type ResponseInspector interface {
+	// InspectResponse is passed a response to inspect.
+	//
+	// An inspector must not modify the response. Doing so could impact later
```
darinkrauss, line 12: I don't think this is required. It is how we define the inspector and since we don't require this right now, I wouldn't make this claim. Just my two cents, though.

request/inspector.go
```diff
@@ -7,6 +7,17 @@ import (
...
+	// inspectors.
+	//
+	// The state of the response's body is undefined. There could be multiple
+	// inspectors before or after any given inspector, so when reading the
+	// body, it's probably a good idea to restore it when done.
+	//
+	// Any error returned will simply be logged. This might be removed in the
```
darinkrauss, line 19: Why? It seems like you are changing the intention of the interface without a need.

client/client.go
darinkrauss: Hmm, I'm not sure I agree with this change. If an inspector is important and shouldn't  be affected by errors from other inspectors, then it should be listed first.

However, I won't cause my disagreement to block this PR.

darinkrauss (dismissed):
Approved even though I disagree with some of the changes.

request/inspector.go
ewollesen: This is just an attempt at documenting what users of this interface can expect.

Since there's no known ordering, any inspector that modifies the Response, could have an impact on later inspectors. As a result, warning folks to _not_ modify the Response seems like the safe path.

The comment aims to make it clear to someone writing their own ResponseInspector, that they probably shouldn't modify the Response, and if they do, they should be extra careful to do so in a way that doesn't interfere with other response inspectors.

request/inspector.go
```diff
@@ -1,28 +1,42 @@
...
-	return &HeadersInspector{}
+func NewHeadersInspector(logger log.Logger) *HeadersInspector {
+	return &HeadersInspector{logger: logger}
 }
 
-func (h *HeadersInspector) InspectResponse(res *http.Response) error {
+func (h *HeadersInspector) InspectResponse(res *http.Response) {
 	if res == nil {
-		return errors.New("response is missing")
+		fmt.Fprintf(os.Stdout, "I'm going to write an error into %+v!\n", h.logger)
```
darinkrauss, line 34: Is this `fmt.Fprintf` intentional?

darinkrauss (changes requested):
I think there is an extraneous `fmt.Fprintf`?

request/inspector.go
ewollesen: Doh! Will fix.

darinkrauss (approved):
LGTM!

PR #695, opened 2024-01-11
break AddBin down into smaller functions

ewollesen:
Namely, addBinAfter, addBinBefore, and replaceBin.

I found this breakdown a little easier to reason about, as it separates the if-then structures and exposes what I feel is a more obvious, and more task-oriented flow.

It might also be interesting too, to have a "readjust" function, that given a slice of buckets, simply re-orders them and inserts/removes gaps as necessary. Then AddBin just becomes something like:

    buckets = append(buckets, newBucket)
    readjustBuckets(buckets)

The signature for removeExcessBuckets has an extra argument, because I couldn't quite figure out how to make the compiler happy without it.

### DIFF
--- a/data/summary/types/summary.go
+++ b/data/summary/types/summary.go
@@ -153,8 +153,6 @@ func CreateBucket[A BucketDataPt[T], T BucketData](t time.Time) *Bucket[A, T] {
 	return bucket
 }
 
-type Buckets[T BucketData, S BucketDataPt[T]] []*Bucket[S, T]
-
 type Stats interface {
 	CGMStats | BGMStats
 }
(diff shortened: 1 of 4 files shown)


data/summary/types/summary.go
```diff
@@ -341,60 +283,58 @@ const MaxBucketGap = -time.Hour * HoursAgoToKeep
...
 	*buckets = append(*buckets, newBucket)
 
-	removeExcessBuckets(buckets, newBucket)
+	removeExcessBuckets(buckets)
 
 	return nil
 }
 
-// addBinBefore readjutss buckets to that newBucket is at the start.
+// addBinBefore readjusts buckets to that newBucket is at the start.
```
ewollesen, line 309: Hehe, thanks. 👍🏻

PR #697, opened 2024-02-07
[BACK-2866] Fix Dexcom task failing for newly created Dexcom connections

darinkrauss:
- Fix Dexcom task failing for newly created Dexcom connections
- Only set CreatedUserID in UsersDataSetsCreate if user AuthDetails
- Always add error to Dexcom tasks whether or not task should fail
- Update tests
- Update .gitignore and .dockerignore
- https://tidepool.atlassian.net/browse/BACK-2866

### DIFF
--- a/data/service/api/v1/users_datasets_create.go
+++ b/data/service/api/v1/users_datasets_create.go
@@ -69,7 +69,9 @@ func UsersDataSetsCreate(dataServiceContext dataService.Context) {
 	}
 
 	dataSet.SetUserID(&targetUserID)
-	dataSet.SetCreatedUserID(pointer.FromString(details.UserID()))
+	if details.IsUser() {
+		dataSet.SetCreatedUserID(pointer.FromString(details.UserID()))
+	}
 
 	dataSet.Normalize(normalizer)
 
--- a/dexcom/fetch/task.go
+++ b/dexcom/fetch/task.go
@@ -35,9 +35,9 @@ func NewTaskCreate(providerSessionID string, dataSourceID string) (*task.TaskCre
 }
 
 func ErrorOrRetryTask(t *task.Task, err error) {
+	t.AppendError(err)
 	if t.IsFailed() {
-		if shouldTaskError(t) {
-			t.AppendError(err)
+		if shouldTaskFail(t) {
 			return
 		}
 		incrementTaskRetryCount(t)
@@ -51,7 +51,7 @@ func FailTask(l log.Logger, t *task.Task, err error) error {
 	return err
 }
 
-func shouldTaskError(t *task.Task) bool {
+func shouldTaskFail(t *task.Task) bool {
 	if t.Data[dexcomTaskRetryField] != nil {
 		count, ok := t.Data[dexcomTaskRetryField].(int)
 		if ok {
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,6 @@
 *.coverprofile
 _bin/
 _tmp/
+coverprofile.out
 debug
 deploy/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.coverprofile
+*.env
 *.test
 /_bin
 /_data
@@ -7,3 +8,5 @@
 /deploy
 /.idea
 .DS_Store
+.envrc
+coverprofile.out
--- a/data/service/api/v1/users_datasets_create_test.go
+++ b/data/service/api/v1/users_datasets_create_test.go
@@ -5,46 +5,73 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/ant0ine/go-json-rest/rest"
 	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/ant0ine/go-json-rest/rest"
 
 	"github.com/tidepool-org/platform/alerts"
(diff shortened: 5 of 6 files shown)


dexcom/fetch/task.go
```diff
@@ -51,7 +51,7 @@ func FailTask(l log.Logger, t *task.Task, err error) error {
 	return err
 }
 
-func shouldTaskError(t *task.Task) bool {
+func shouldTaskFail(t *task.Task) bool {
```
toddkazakov, line 54: I don't think this function handles correctly the case where the `t.Data[dexcomTaskRetryField]` is not set. When `t.Data[dexcomTaskRetryField]` is `nil`, this function should return `false` instead of `true`, so the retry counter can be set to `1`.

toddkazakov (changes requested):
I think there's one edge case that should be fixed

jh-bate (dismissed):
nothing else that I can see - will approve on the understanding that Todd's comment needs to be addressed first

data/service/api/v1/users_datasets_create.go
```diff
@@ -69,7 +69,9 @@ func UsersDataSetsCreate(dataServiceContext dataService.Context) {
 	}
 
 	dataSet.SetUserID(&targetUserID)
-	dataSet.SetCreatedUserID(pointer.FromString(details.UserID()))
+	if details.IsUser() {
+		dataSet.SetCreatedUserID(pointer.FromString(details.UserID()))
```
toddkazakov, line 73: FYI, `details.UserID` will be set to the id backend client service account, which is a real user in Keycloak.

dexcom/fetch/task.go
darinkrauss: Yes, that is correct and I intentionally did not fix that yet. 

If we make that fix right now, then every Dexcom task that hits a 429 will start down the "retry path" and, after retrying 3 times, will be marked as failed. I really don't want that to happen quite yet. I need to first straighten out what errors cause a formal retry (where the retry counter is incremented; e.g. Dexcom auth errors), what errors retry for "free" (no increment; e.g. Dexcom 429s), and what errors fail immediately (e.g bad data). Furthermore, I need to add code that resets the retry counter on success (as there is no code to do that currently).

data/service/api/v1/users_datasets_create.go
darinkrauss: Hmm, that's not what I'm seeing. I'm getting an empty string which causes a failure during later validation (`CreatedUserID` of `nil` is fine, but an empty string is not).

darinkrauss: I'll double check.

data/service/api/v1/users_datasets_create.go
```diff
@@ -69,7 +69,9 @@ func UsersDataSetsCreate(dataServiceContext dataService.Context) {
...
+	}
```
ewollesen, line 74: Would it be simpler, given the validation failure, to do:
```suggestion
if userID := details.UserID(); userID != "" {
    dataSet.SetCreatedUserID(pointer.FromString(userID))
}
```

darinkrauss: Per discussions in Slack, will leave as-is. Using the system-generated user ID for service session tokens will cause downstream issues. Plus, existing platform code that references the `UserID` from `AuthDetails` checks `IsUser` beforehand.

data/service/api/v1/users_datasets_create.go
darinkrauss: While the service session token does include the `UserID` of the backend client service account, it is not being used because the session token is marked as `IsServer`. Find the relevant code that ignores the `UserID` for a server session token at https://github.com/tidepool-org/platform/blob/f7976305c9b2324d441c82a45f051e03ce0ff95d/auth/client/external.go#L208.

dexcom/fetch/runner.go
```diff
@@ -26,10 +26,10 @@ import (
 )
 
 const (
-	AvailableAfterDurationMaximum = 75 * time.Minute
-	AvailableAfterDurationMinimum = 45 * time.Minute
+	AvailableAfterDurationMaximum = 135 * time.Minute
+	AvailableAfterDurationMinimum = 105 * time.Minute
```
darinkrauss, line 30: Doubled average period between Dexcom task executions from 60 minutes to 120 minutes.

dexcom/fetch/runner.go
```diff
@@ -26,10 +26,10 @@ import (
...
 	DataSetSize                   = 2000
-	TaskDurationMaximum           = 5 * time.Minute
+	TaskDurationMaximum           = 10 * time.Minute
```
darinkrauss, line 32: Doubled task duration to account for a couple of long running errors observed in logs.

ewollesen: > Doubled task duration to account for a couple of long running errors observed in logs.

I think that'd be useful to capture in the commit message, to help future us. Same with your other comment in this PR. Just $0.02 though.

toddkazakov (approved):
Should `.vscode/launch.json` be added to gitignore? Other than that, it looks ok.

ewollesen:
Merging this on behalf of Darin, since this is a critical update, and Darin is OOO. Squashing as I believe that's what Darin said he normally does. Will combine commit messages in a way that makes as much sense as possible.

PR #698, opened 2024-02-12
[BACK-2780] Add new user profiles endpoint.

lostlevels:
Since a lot of this is copy paste from shoreline, for reviewers:

* ~~`shoreline/user/user.go` => `platform/user/full_user.go` - (because there's already a type called `User`).~~ `platform/user.go` `user.User` extended with shoreline user fields.
* ~~`shoreline/user/hasher.go` => `platform/user/hasher.go`~~
* `shoreline/user/storage.go` => `platform/user/user_accessor.go` (somewhat like the interface of the Storage repository, but simplified).
* `shoreline/user/migrationStore.go` => `platform/user/keycloak/user_accessor.go` (Takes the logic around User creation / manipulation and removes the fallback / mongodb stuff), implements `platform/user.UserAccessor` interface
* `shoreline/keycloak/client.go` => `platform/user/keycloak/client.go` (Wrapper around gocloak).


user/profile.go
```diff
@@ -0,0 +1,231 @@
...
+	Birthday       Date     `json:"birthday"`
+	DiagnosisDate  Date     `json:"diagnosisDate"`
+	DiagnosisType  string   `json:"diagnosisType"`
+	TargetDevices  []string `json:"targetDevices"`
+	TargetTimezone string   `json:"targetTimezone"`
+	About          string   `json:"about"`
+}
+
+type ClinicProfile struct {
+	Name      string   `json:"diagnosisDate"`
```
toddkazakov, line 90:
```suggestion
	Name      string   `json:"name"`
```

user/profile.go
```diff
@@ -0,0 +1,231 @@
...
+	Role      []string `json:"role"`
+	Telephone string   `json:"telephone"`
+}
+
+// TODO: may or may not have the "profile_" prefix in the keycloak attribute name as it is somewhat redundant.
```
toddkazakov, line 95: Let's strip the `profile_` prefix

auth/service/api/v1/profile.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/tidepool-org/platform/service/api"
+	structValidator "github.com/tidepool-org/platform/structure/validator"
+	"github.com/tidepool-org/platform/user"
+)
+
+func (r *Router) ProfileRoutes() []*rest.Route {
+	return []*rest.Route{
+		rest.Get("/v1/users/:userId/profile", api.RequireAuth(r.GetProfile)),
+		rest.Get("/v1/users/legacy/:userId/profile", api.RequireAuth(r.GetLegacyProfile)),
+		// The following modification routes required custodian access in seagull, but I'm not sure that's quite right - it seems it should be if the user can modify the userId.
```
toddkazakov, line 19: This is very strange. I looked into seagull, and it seems that it really requires `custodian` permissions, but users can update their own profiles, so maybe gatekeeper returns `custodian` in the response when calling userInGroup with the same user for the grantor and grantee e.g. `/access/todd/todd`?

auth/service/api/v1/profile.go
```diff
@@ -0,0 +1,146 @@
...
+		rest.Put("/v1/users/:userId/profile", r.requireWriteAccess("userId", r.UpdateProfile)),
```
toddkazakov, line 20: Seagull also exposes a `POST` endpoint.

permission/client/client.go
```diff
@@ -46,3 +46,41 @@ func (c *Client) GetUserPermissions(ctx context.Context, requestUserID string, t
...
+
+func (c *Client) HasWritePermissions(ctx context.Context, granteeUserID, grantorUserID string) (has bool, err error) {
+	if granteeUserID != "" && granteeUserID == grantorUserID {
+		return true, nil
+	}
+	perms, err := c.GetUserPermissions(ctx, granteeUserID, grantorUserID)
+	if err != nil {
+		return false, err
+	}
+	return len(perms[permission.Custodian]) > 0 || len(perms[permission.Write]) > 0 || len(perms[permission.Owner]) > 0, nil
```
toddkazakov, line 84: If gatekeeper returns:
```
{
   custodian: {}
}
```

We will deserialize this to `map[string]Permission` where `Permission` is `map[string]interface{}`.  If I am not mistaken `len(perms[permission.Custodian])` check whether that second map is empty, which will always be true, no?

permission/permission.go
```diff
@@ -17,6 +17,10 @@ const (
 
 type Client interface {
 	GetUserPermissions(ctx context.Context, requestUserID string, targetUserID string) (Permissions, error)
+	// Not sure whether to put these methods in platform/permission or go-common/clients
```
toddkazakov, line 20: Go-common is very difficult to maintain. It's better to keep those here.

user/full_user.go
```diff
@@ -0,0 +1,705 @@
...
+	Roles         []string `json:"roles,omitempty"`
+	Attributes    struct {
+		TermsAcceptedDate []string `json:"terms_and_conditions,omitempty"`
+	} `json:"attributes"`
+}
+
+/*
+ * Incoming user details used to create or update a `User`
+ */
+type NewUserDetails struct {
```
toddkazakov, line 99: Are those structs used? Please delete them if they are not used.

user/full_user.go
```diff
@@ -0,0 +1,705 @@
...
+	Roles          []string
+	TermsAccepted  *string
+	EmailVerified  *bool
+}
+
+type Profile struct {
+	FullName string `json:"fullName"`
+}
+
+var (
```
toddkazakov, line 126: The validation logic doesn't conform to platform validators and should be removed to avoid confusion.

user/full_user.go
```diff
@@ -0,0 +1,705 @@
...
+	RoleCustodialAccount: {},
+	RoleMigratedClinic:   {},
+	RolePatient:          {},
+}
+
+var custodialAccountRoles = []string{RoleCustodialAccount, RolePatient}
+
+// FullUser is the rull representation of a user. It is a
+// temporary type until I can figure out how much the
+// existing user.User type can be extended to include these
```
toddkazakov, line 42: Please extend `user.User` to avoid bringing cruft from shoreline.

user/hasher.go
```diff
@@ -0,0 +1,57 @@
...
+	"crypto/sha1"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"math/big"
+	"strconv"
+	"time"
+)
+
+func generateUniqueHash(strings []string, length int) (string, error) {
```
toddkazakov, line 14: I don't think those hashes are used anywhere, so please delete them.

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,231 @@
...
+
+	if user.IsMigrated {
+		return m.updateKeycloakUser(ctx, user, details)
+	}
+	// expected all users to be migrated(?)
+	return nil, userLib.ErrUserNotMigrated
+}
+
+func (m *keycloakUserAccessor) assertEmailsUnique(ctx context.Context, userId string, emails []string) error {
+	// for _, email := range emails {
```
toddkazakov, line 93: Delete if unused.

toddkazakov (changes requested):
Overall it looks good. I found few issues:
- There is a lot of cruft that was copied from shoreline. I believe the majority of those functions are not used. There's no reason to keep unused code in platform, because it only leads to confusion. 
- There is probably a gap in the requirements that I discovered when I read the seagull code to try and figure out why the profile endpoints require custodian permissions. In order to fully deprecate and migrate existing seagull profiles from Mongo to Keycloak, we have to provide an alternative implementation of [this functionality](https://github.com/tidepool-org/seagull/blob/master/lib/routes/seagullApi.js#L241-L374). For each user sharing their data with [a user with a given id](https://github.com/tidepool-org/seagull/blob/master/lib/seagullService.js#L53) seagull fetches the user object from shoreline and merges it with the user profile attributes stored in the profile object in mongo. The response is sanitized depending on the actual permissions of the requesting user. This deserves its own ticket, but please do some research what the linked code does and then write the requirements. The alternative implementation should be much simpler, because the user account and profiles will be stored in a single service (Keycloak).

user/profile.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+// Date is a string of type YYYY-mm-dd, the reason this isn't just a type definition
+// of a time.Time is to ignore timezones when marshaling.
+type Date string
+
+// UserProfile represents the user modifiable attributes of a user. It is named
+// somewhat redundantly as UserProfile instead of Profile because there already
+// exists a type Profile in this package.
+type UserProfile struct {
```
toddkazakov, line 22: MRN field is missing

user/full_user.go
lostlevels: They do not, but if we are going to (eventually) add the user creation routes from shoreline to platform these would be use. I can remove for now.

permission/client/client.go
lostlevels: Yes I think I only handled the case where the user created his own account.

auth/service/api/v1/profile.go
lostlevels: The seagull code was very confusing with the layers of nesting but custodian check is *i believe*

* check if this is a service call, if yes allow.
* check if the requester in the request equals the user in the url param, if yes allow.
* check if the requester in the request has any permissions to the user in the url param OR the user in the url param has any permissions to the requester, if yes allow.
* otherwise disallow

toddkazakov (dismissed):
Looks ok, but needs to be thoroughly tested. @lostlevels please open a new ticket for reimplementing the missing seagull endpoint I mentioned in my previous review.

auth/service/api/v1/profile.go
```diff
@@ -0,0 +1,260 @@
...
+	for userID, trustPerms := range mergedUserPerms {
+		if userMatchesQueryOnPermissions(*trustPerms, filter) {
+			filteredUserPerms[userID] = trustPerms
+		}
+	}
+
+	results := make([]*user.User, 0, len(mergedUserPerms))
+	// just doing sequentially fetching of users for now
+	for userID, trustPerms := range filteredUserPerms {
+		// Does this mean all users should already be migrated
```
toddkazakov, line 122: Yes, we should migrate all users to mongo

auth/service/api/v1/profile_filter.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	return true
+}
+
+func stringToBoolean(value string) bool {
+	value = strings.TrimSpace(strings.ToLower(value))
+	return slices.Contains(TRUES, value)
+}
+
+func parseUsersQuery(query url.Values) *usersProfileFilter {
```
toddkazakov, line 104: My impressions is that this query is not used at all. Can we check blip and uploader?

lostlevels: On checking blip and uploader it seems they are indeed no query string parameters used.

[blip](https://github.com/tidepool-org/blip/blob/develop/app/core/api.js#L426)
[platform-client](https://github.com/tidepool-org/platform-client/blob/develop/index.js#L334)
[uploader](https://github.com/tidepool-org/uploader/blob/master/lib/core/api.js#L274)
[uploader](https://github.com/tidepool-org/uploader/blob/master/lib/core/api.js#L315)

Can you confirm that no query string parameters are passed for `getAssociatedUsersDetails` @krystophv @gniezen @jh-bate ?

auth/service/api/v1/profile.go
```diff
@@ -103,64 +104,58 @@ func (r *Router) GetUsersWithProfiles(res rest.ResponseWriter, req *rest.Request
...
+			}
+			if err != nil {
+				return err
+			}
+			trustorPerms := trustPerms.TrustorPermissions
+			if trustorPerms == nil || len(*trustorPerms) == 0 {
+				profile = profile.ClearPatientInfo()
+			} else {
+				if trustorPerms.HasAny(permission.Custodian, permission.Read, permission.Write) {
+					// TODO: need to read seagull.value.settings - confirm this is actually used
```
toddkazakov, line 140: Settings are not part of the profile object so probably there's no need to do anything?

auth/service/api/v1/profile.go
```diff
@@ -103,64 +104,58 @@ func (r *Router) GetUsersWithProfiles(res rest.ResponseWriter, req *rest.Request
...
+				}
+				if trustorPerms.Has(permission.Custodian) {
+					// TODO: need to read seagull.value.preferences - confirm this is actually used
```
toddkazakov, line 143: Preferences are not part of the profile object so probably there's no need to do anything?

auth/service/api/v1/profile.go
lostlevels: I see the settings and preferences returned in the old seagull code, but don't know if they are actually used by clients. @krystophv @clintonium-119 @gniezen Do you know if the returned result of `paltform-client getAssociatedUsersDetails` ever uses the returned users' settings or preferences in `blip` or `uploader`?

user/fallback_user_accessor.go
```diff
@@ -0,0 +1,123 @@
...
+	for i := range arbritraryRetryLimit {
+		err = f.upsertUserProfile(ctx, id, profile)
+		if errors.Is(err, ErrUserProfileMigrationInProgress) {
+			time.Sleep(time.Second * time.Duration(i+1))
+			continue
+		}
+		if err != nil {
+			return err
+		}
+	}
```
toddkazakov, line 74: You're missing a `break` here

user/keycloak/client.go
```diff
@@ -0,0 +1,615 @@
...
+	"time"
+
+	"github.com/Nerzal/gocloak/v13"
+	"github.com/Nerzal/gocloak/v13/pkg/jwx"
+	"github.com/go-resty/resty/v2"
+	"github.com/kelseyhightower/envconfig"
+	"golang.org/x/oauth2"
+
+	"github.com/tidepool-org/platform/pointer"
+	userLib "github.com/tidepool-org/platform/user"
```
toddkazakov: This doesn't follow platform import conventions

user/keycloak/client.go
```diff
@@ -0,0 +1,615 @@
...
+)
+
+const (
+	tokenPrefix            = "kc"
```
toddkazakov: This is not used

user/keycloak/client.go
```diff
@@ -0,0 +1,615 @@
...
+	tokenPartsSeparator    = ":"
```
toddkazakov: This is not used

user/user.go
```diff
@@ -143,3 +205,9 @@ func ValidateID(value string) error {
 }
 
 var idExpression = regexp.MustCompile(`^([0-9a-f]{10}|[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12})$`)
+
+// IsValidUserID return true if the string is in a human readable uuid hex 8-4-4-4-12 format or legacy alphanumeric 10 characters
+func IsValidUserID(id string) bool {
+	ok, _ := regexp.MatchString(`^([a-fA-F0-9]{10})$|^([a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12})$`, id)
```
toddkazakov: Why is the regexp duplicated with the one above?

user/profile.go
```diff
@@ -0,0 +1,561 @@
...
+		}
+		if val := pointer.ToString(up.Clinic.NPI); val != "" {
+			addAttribute(attributes, "clinic_npi", val)
+		}
+	}
+
+	return attributes
+}
+
+func ProfileFromAttributes(username string, attributes map[string][]string, roles []string) (profile *UserProfile, ok bool) {
```
toddkazakov: What is the meaning of the `ok` bool return argument here? Returning a `nil` instead of having multiple return arguments is equivalent.

user/profile.go
```diff
@@ -0,0 +1,561 @@
...
+}
+
+// LegacyUserProfile represents the old seagull format for a profile.
+type LegacyUserProfile struct {
+	FullName        string                `json:"fullName,omitempty"` // string pointer because some old profiles have empty string as full name
+	Patient         *LegacyPatientProfile `json:"patient,omitempty"`
+	Clinic          *ClinicProfile        `json:"clinic,omitempty"`
+	MigrationStatus migrationStatus       `json:"-"`
+	// The Email and Emails fields are legacy properties that will be populated from the keycloak user if the profile is finished migrating, otherwise from the seagull collection
+	Email  string   `json:"email,omitempty"`
```
toddkazakov: What is the purpose of `email` and `emails`? Are they used by any of the services? If not, I suggest removing them.

user/profile.go
```diff
@@ -0,0 +1,561 @@
...
+	TargetTimezone string   `json:"targetTimezone,omitempty"`
+	About          string   `json:"about,omitempty"`
+	MRN            string   `json:"mrn,omitempty"`
+	BiologicalSex  string   `json:"biologicalSex,omitempty"`
+
+	Custodian *Custodian `json:"custodian,omitempty"`
+	// The PRESENCE of a clinic object in a profile is used by blip to determine which page to show so this needs to be returned in the response.
+	// There are clinicians/legacy clinics with completely empty values within the clinic object but are still clinicians/clinics.
+	Clinic *ClinicProfile `json:"clinic,omitempty"`
+	Email  string         `json:"-"` // This is used when returning profiles in the legacy format. It is not stored in the profile, but is populated from the keycloak username and not returned in the new profiles route.
```
toddkazakov: If this field is not used, I suggest removing it

user/profile.go
```diff
@@ -0,0 +1,561 @@
...
+	}
+)
+
+// Date is a string of type YYYY-mm-dd, the reason this isn't just a type definition
+// of a time.Time is to ignore timezones when marshaling.
+type Date string
+
+// UserProfile represents the user modifiable attributes of a user. It is named
+// somewhat redundantly as UserProfile instead of Profile because there already
+// exists a type Profile in this package.
```
toddkazakov: Is this still true? If not, let's rename it to profile.

user/profile.go
```diff
@@ -0,0 +1,561 @@
...
+	Birthday       Date     `json:"birthday,omitempty"`
+	DiagnosisDate  Date     `json:"diagnosisDate,omitempty"`
+	DiagnosisType  string   `json:"diagnosisType,omitempty"`
+	TargetDevices  []string `json:"targetDevices,omitempty"`
+	TargetTimezone string   `json:"targetTimezone,omitempty"`
+	About          string   `json:"about,omitempty"`
+	IsOtherPerson  jsonBool `json:"isOtherPerson,omitempty"`
+	MRN            string   `json:"mrn,omitempty"`
+	BiologicalSex  string   `json:"biologicalSex,omitempty"`
+	// The Email and Emails fields are legacy properties that will be populated from the keycloak user if the profile is finished migrating, otherwise from the seagull collection
```
toddkazakov: Are those used? If not let's remove them. I am little confused why those exist in two levels.

user/profile.go
```diff
@@ -0,0 +1,561 @@
...
+
+func (up *UserProfile) ToAttributes() map[string][]string {
+	attributes := map[string][]string{}
+
+	if up.FullName != "" {
+		addAttribute(attributes, "full_name", up.FullName)
+	}
+	if up.Custodian != nil && up.Custodian.FullName != "" {
+		addAttribute(attributes, "custodian_full_name", up.Custodian.FullName)
+		// The "has_custodian" attribute is only added so that filtering on users is simpler via the keycloak API - because
```
toddkazakov, line 272: Is this attribute used? What service does the filtering?

user/keycloak/client.go
```diff
@@ -0,0 +1,615 @@
...
+		ID:            &user.ID,
+		Username:      &user.Username,
+		Enabled:       &user.Enabled,
+		EmailVerified: &user.EmailVerified,
+		FirstName:     &user.FirstName,
+		LastName:      &user.LastName,
+		Email:         &user.Email,
+	}
+
+	attrs := map[string][]string{
```
toddkazakov: When a users profile is updated, this will clear out any attributes that are not defined here or in the profile. This is not desirable.

user/legacy_raw_seagull_profile.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+var (
+	ErrSeagullFieldNotFound  = errors.New("seagull field not found within value object string")
+	ErrSeagullMarshalValue   = errors.New(`unable to encode seagull "value" to JSON`)
+	ErrSeagullUnmarshalValue = errors.New(`unable to decode seagull "value" from JSON`)
+)
+
+// LegacySeagullDocument is the database model representation of the legacy
+// seagull collection object. The value is a raw stringified JSON blob. TODO:
```
toddkazakov: Please add a ticket instead of TODO

auth/service/api/v1/profile.go
```diff
@@ -0,0 +1,246 @@
...
+
+type trustPermissions struct {
+	TrustorPermissions *permission.Permission
+	TrusteePermissions *permission.Permission
+}
+
+func (r *Router) ProfileRoutes() []*rest.Route {
+	return []*rest.Route{
+		rest.Get("/v1/users/:userId/profile", r.requireMembership("userId", r.GetProfile)),
+		rest.Get("/v1/users/:userId/users", r.requireMembership("userId", r.GetUsersWithProfiles)),
```
toddkazakov: The permission check here seems incorrect. If user A shares their account with user B, then user B (making the request to this endpoint) should only be able to see the list of users that user A is sharing their account with, only if B has a custodian/root relationship.

toddkazakov: Do we also need an additional route with `/metadata` prefix?

auth/service/api/v1/profile.go
```diff
@@ -0,0 +1,246 @@
...
+		return nil, err
+	}
+	if profile == nil {
+		return nil, user.ErrUserProfileNotFound
+	}
+	return profile, nil
+}
+
+// GetProfile returns the user's profile in the new, non seagull, format
+func (r *Router) GetProfile(res rest.ResponseWriter, req *rest.Request) {
```
toddkazakov, line 78: This endpoint uses `hasMembership` guard which checks whether a sharing relationship exists in either direction. If user A shares their account with B, user B shouldn't be able to see `profile.patient`. You're missing a call to `profile.ClearPatientInfo()`.

auth/service/api/v1/profile.go
```diff
@@ -0,0 +1,246 @@
...
+	if err := group.Wait(); err != nil {
+		r.handleUserOrProfileErr(responder, err)
+		return
+	}
+
+	responder.Data(http.StatusOK, results)
+}
+
+// GetLegacyProfile returns user profiles in the legacy seagull format.
+func (r *Router) GetLegacyProfile(res rest.ResponseWriter, req *rest.Request) {
```
toddkazakov, line 187: This endpoint uses `hasMembership` guard which checks whether a sharing relationship exists in either direction. If user A shares their account with B, user B shouldn't be able to see `profile.patient`. You're missing a call to `profile.ClearPatientInfo()`.

auth/service/api/v1/profile.go
```diff
@@ -0,0 +1,246 @@
...
+	responder := request.MustNewResponder(res, req)
+	ctx := req.Context()
+	userID := req.PathParam("userId")
+
+	profile := &user.UserProfile{}
+	if err := request.DecodeRequestBody(req.Request, profile); err != nil {
+		responder.Error(http.StatusBadRequest, err)
+		return
+	}
+	if err := structValidator.New(log.LoggerFromContext(ctx)).Validate(profile); err != nil {
```
toddkazakov: This call is redundant. `DecodeRequestBody` already calls `Validate`.

auth/service/api/v1/profile.go
```diff
@@ -0,0 +1,246 @@
...
+		rest.Put("/v1/users/:userId/profile", r.requireCustodian("userId", r.UpdateProfile)),
+		rest.Post("/v1/users/:userId/profile", r.requireCustodian("userId", r.UpdateProfile)),
+
+		rest.Put("/v1/users/legacy/:userId/profile", r.requireCustodian("userId", r.UpdateLegacyProfile)),
+		rest.Put("/metadata/:userId/profile", r.requireCustodian("userId", r.UpdateLegacyProfile)),
+
+		rest.Post("/v1/users/legacy/:userId/profile", r.requireCustodian("userId", r.UpdateLegacyProfile)),
+		rest.Post("/metadata/:userId/profile", r.requireCustodian("userId", r.UpdateLegacyProfile)),
+
+		rest.Delete("/v1/users/:userId/profile", r.requireCustodian("userId", r.DeleteProfile)),
```
toddkazakov: This route shouldn't even exist - seagull doesn't delete profiles, it returns 501

auth/service/api/v1/permission.go
```diff
@@ -0,0 +1,116 @@
...
+			}
+			handlerFunc(res, req)
+		}
+	}
+}
+
+// requireWriteAccess aborts with an error if the request isn't a server request
+// or the authenticated user doesn't have access to the user id in the url param,
+// targetParamUserID
+func (r *Router) requireWriteAccess(targetParamUserID string, handlerFunc rest.HandlerFunc) rest.HandlerFunc {
```
toddkazakov: This isn't used anywhere. Please delete it.

permission/permission.go
```diff
@@ -15,9 +25,17 @@ const (
...
 type Client interface {
 	GetUserPermissions(ctx context.Context, requestUserID string, targetUserID string) (Permissions, error)
 	UpdateUserPermissions(ctx context.Context, requestUserID string, targetUserID string, permissions Permissions) error
+	// GroupsForUser returns permissions that have been shared with granteeUserID. It is keyed by the user that has shared something with granteeUserID. It includes the user themself.
+	GroupsForUser(ctx context.Context, granteeUserID string) (Permissions, error)
+	// UsersInGroup returns permissions that the user with id sharerID has shared with others, keyed by user id. It includes the user themself.
+	UsersInGroup(ctx context.Context, sharerID string) (Permissions, error)
+	HasMembershipRelationship(ctx context.Context, granteeUserID, grantorUserID string) (has bool, err error)
+	HasCustodianPermissions(ctx context.Context, granteeUserID, grantorUserID string) (has bool, err error)
+	HasWritePermissions(ctx context.Context, granteeUserID, grantorUserID string) (has bool, err error)
```
toddkazakov: This is not used, please remove it

permission/client/client.go
```diff
@@ -63,3 +63,92 @@ func (c *Client) UpdateUserPermissions(ctx context.Context, sharerUserID string,
...
+func (c *Client) HasCustodianPermissions(ctx context.Context, granteeUserID, grantorUserID string) (has bool, err error) {
+	perms, err := c.GetUserPermissions(ctx, granteeUserID, grantorUserID)
+	if err != nil {
+		return false, err
+	}
+	_, ok := perms[permission.Custodian]
+	return ok, nil
+}
+
+func (c *Client) HasWritePermissions(ctx context.Context, granteeUserID, grantorUserID string) (has bool, err error) {
```
toddkazakov: This is not used, please remove it

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,121 @@
...
+
+type keycloakUserAccessor struct {
+	cfg                      *KeycloakConfig
+	adminToken               *oauth2.Token
+	adminTokenRefreshExpires time.Time
+	keycloakClient           *keycloakClient
+}
+
+func NewKeycloakUserAccessor(config *KeycloakConfig) *keycloakUserAccessor {
+	newKeycloakClient(config)
```
toddkazakov:
```suggestion
```

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if user == nil || user.Profile == nil {
+		return nil, userLib.ErrUserProfileNotFound
+	}
+	return user.Profile.ToLegacyProfile(pointer.ToStringArray(user.Roles)), nil
+}
+
+func (m *keycloakUserAccessor) FindUserProfileV2(ctx context.Context, id string) (*userLib.UserProfile, error) {
```
toddkazakov:
```suggestion
func (m *keycloakUserAccessor) FindUserProfile(ctx context.Context, id string) (*userLib.UserProfile, error) {
```

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if keycloakUser == nil {
+		return nil, userLib.ErrUserNotFound
+	}
+	return newUserFromKeycloakUser(keycloakUser), nil
+}
+
+func (m *keycloakUserAccessor) FindUserProfile(ctx context.Context, id string) (*userLib.LegacyUserProfile, error) {
```
toddkazakov:
```suggestion
func (m *keycloakUserAccessor) FindLegacyUserProfile(ctx context.Context, id string) (*userLib.LegacyUserProfile, error) {
```

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,121 @@
...
+		return users, err
+	}
+
+	for _, user := range keycloakUsers {
+		users = append(users, newUserFromKeycloakUser(user))
+	}
+	return users, nil
+}
+
+func (m *keycloakUserAccessor) UpdateUserProfile(ctx context.Context, userID string, p *userLib.LegacyUserProfile) error {
```
toddkazakov:
```suggestion
func (m *keycloakUserAccessor) UpdateLegacyUserProfile(ctx context.Context, userID string, p *userLib.LegacyUserProfile) error {
```

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return err
+	}
+	if !userLib.HasClinicOrClinicianRole(roles) && p.Clinic != nil {
+		p.Clinic = nil
+	}
+	return m.keycloakClient.UpdateUserProfile(ctx, userID, p.ToUserProfile())
+}
+
+func (m *keycloakUserAccessor) UpdateUserProfileV2(ctx context.Context, userID string, p *userLib.UserProfile) error {
```
toddkazakov:
```suggestion
func (m *keycloakUserAccessor) UpdateUserProfile(ctx context.Context, userID string, p *userLib.UserProfile) error {
```

user/user_accessor.go
```diff
@@ -0,0 +1,90 @@
...
+	// ErrUserProfileMigrationInProgress means a specific user profile is
+	// currently being migrated so the client should ideally wait and
+	// retry their operation again since the migration for a single user
+	// should be no longer than a few seconds.
+	ErrUserProfileMigrationInProgress = errors.New("user migration is in progress")
+)
+
+type LegacyProfileAccessor interface {
+	FindUserProfile(ctx context.Context, userID string) (*LegacyUserProfile, error)
+	UpdateUserProfile(ctx context.Context, userID string, p *LegacyUserProfile) error
```
toddkazakov:
```suggestion
	UpdateLegacyUserProfile(ctx context.Context, userID string, p *LegacyUserProfile) error
```

user/user_accessor.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// ErrUserProfileMigrationInProgress means a specific user profile is
+	// currently being migrated so the client should ideally wait and
+	// retry their operation again since the migration for a single user
+	// should be no longer than a few seconds.
+	ErrUserProfileMigrationInProgress = errors.New("user migration is in progress")
+)
+
+type LegacyProfileAccessor interface {
+	FindUserProfile(ctx context.Context, userID string) (*LegacyUserProfile, error)
```
toddkazakov:
```suggestion
	FindLegacyUserProfile(ctx context.Context, userID string) (*LegacyUserProfile, error)
```

user/user_accessor.go
```diff
@@ -0,0 +1,90 @@
...
+	UpdateUserProfile(ctx context.Context, userID string, p *LegacyUserProfile) error
+	DeleteUserProfile(ctx context.Context, userID string) error
```
toddkazakov: Deleting a user profile was not implemented in seagull and should be removed.

user/user_accessor.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+type ProfileAccessor interface {
+	LegacyProfileAccessor
+	UpdateUserProfileV2(ctx context.Context, userID string, p *UserProfile) error
```
toddkazakov:
```suggestion
	UpdateUserProfile(ctx context.Context, userID string, p *UserProfile) error
```

user/user_accessor.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+//go:generate mockgen -build_flags=--mod=mod -destination=./user_mock.go -package=user . ProfileAccessor,UserAccessor
+
+var (
+	ShorelineManagedRoles = map[string]struct{}{"patient": {}, "clinic": {}, "clinician": {}, "custodial_account": {}}
+
+	ErrUserNotFound        = errors.New("user not found")
+	ErrUserProfileNotFound = errors.New("profile not found")
+	ErrUserConflict        = errors.New("user already exists")
```
toddkazakov: Not used

user/user_accessor.go
```diff
@@ -0,0 +1,90 @@
...
+	ErrEmailConflict       = errors.New("email already exists")
```
toddkazakov: Not used

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,121 @@
...
+	return m.keycloakClient.UpdateUserProfile(ctx, userID, p)
+}
+
+func (m *keycloakUserAccessor) DeleteUserProfile(ctx context.Context, userId string) error {
```
toddkazakov: This should be removed - seagull returns 501 when trying to delete a profile

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"golang.org/x/oauth2"
+
+	"github.com/tidepool-org/platform/pointer"
+	userLib "github.com/tidepool-org/platform/user"
+)
+
+type keycloakUserAccessor struct {
+	cfg                      *KeycloakConfig
+	adminToken               *oauth2.Token
```
toddkazakov: This can be removed as it's not used

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,121 @@
...
+	adminTokenRefreshExpires time.Time
```
toddkazakov: This can be removed as it's not used

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,121 @@
...
+	"time"
+
+	"golang.org/x/oauth2"
+
+	"github.com/tidepool-org/platform/pointer"
+	userLib "github.com/tidepool-org/platform/user"
+)
+
+type keycloakUserAccessor struct {
+	cfg                      *KeycloakConfig
```
toddkazakov: This can be removed as it's not used

user/user.go
```diff
@@ -52,11 +55,28 @@ type Client interface {
...
+	Enabled              bool         `json:"-" bson:"-"`
+	CreatedTime          string       `json:"createdTime,omitempty" bson:"createdTime,omitempty"`
+	CreatedUserID        string       `json:"createdUserId,omitempty" bson:"createdUserId,omitempty"`
+	ModifiedTime         string       `json:"modifiedTime,omitempty" bson:"modifiedTime,omitempty"`
+	ModifiedUserID       string       `json:"modifiedUserId,omitempty" bson:"modifiedUserId,omitempty"`
+	DeletedTime          string       `json:"deletedTime,omitempty" bson:"deletedTime,omitempty"`
+	DeletedUserID        string       `json:"deletedUserId,omitempty" bson:"deletedUserId,omitempty"`
+	Profile              *UserProfile `json:"profile,omitempty" bson:"-"`
+	PasswordExists       *bool        `json:"passwordExists,omitempty" bson:"-"`
+	// The following 2 properties are only returned for the route that returns users that have shared their data w/ another user
```
toddkazakov: Is suggest removing those two fields from the User model. Feel free to create a new view model that embeds the User struct in the API layer (as this is the only place where those two fields are being set)

user/user.go
```diff
@@ -52,11 +55,28 @@ type Client interface {
 }
 
 type User struct {
-	UserID        *string   `json:"userid,omitempty" bson:"userid,omitempty"`
-	Username      *string   `json:"username,omitempty" bson:"username,omitempty"`
-	EmailVerified *bool     `json:"emailVerified,omitempty" bson:"emailVerified,omitempty"`
-	TermsAccepted *string   `json:"termsAccepted,omitempty" bson:"termsAccepted,omitempty"`
-	Roles         *[]string `json:"roles,omitempty" bson:"roles,omitempty"`
+	UserID               *string      `json:"userid,omitempty" bson:"userid,omitempty"`
```
toddkazakov: I see a lot of fields here were copied from shoreline and  are not used. Can we clean this up?

user/user.go
```diff
@@ -100,16 +116,62 @@ func (u *User) IsPatient() bool {
...
+
+func (u *User) IsEnabled() bool {
+	if u.IsMigrated {
+		return u.Enabled
+	}
+	return u.PwHash != "" && !u.IsDeleted()
+}
+
+func (u *User) IsDeleted() bool {
+	// mdb only?
```
toddkazakov: Yes, can we remove this?

user/user.go
```diff
@@ -100,16 +116,62 @@ func (u *User) IsPatient() bool {
...
+func (u *User) AreTermsAccepted() bool {
+	if u.TermsAccepted == nil {
+		return false
+	}
+	_, err := TimestampToUnixString(*u.TermsAccepted)
+	return err == nil
+}
+
+func (u *User) IsEnabled() bool {
+	if u.IsMigrated {
```
toddkazakov: All users are now in Keycloak, this can be simplified

user/user.go
```diff
@@ -100,16 +116,62 @@ func (u *User) IsPatient() bool {
...
+	return ""
+}
+
+// IsClinic returns true if the user is legacy clinic Account
+func (u *User) IsClinic() bool {
+	return u.HasRole(RoleClinic)
+}
+
+func (u *User) IsCustodialAccount() bool {
+	return u.HasRole(RoleCustodialAccount)
```
toddkazakov: Are there custodial users in Keycloak that don't have this role?

user/keycloak/client.go
```diff
@@ -0,0 +1,615 @@
...
+	Realm                 string `envconfig:"TIDEPOOL_KEYCLOAK_REALM" required:"true"`
+	AdminUsername         string `envconfig:"TIDEPOOL_KEYCLOAK_ADMIN_USERNAME" required:"true"`
+	AdminPassword         string `envconfig:"TIDEPOOL_KEYCLOAK_ADMIN_PASSWORD" required:"true"`
+}
+
+func (c *KeycloakConfig) FromEnv() error {
+	return envconfig.Process("", c)
+}
+
+// keycloakUser is an intermediate user representation from FullModel to gocloak's model - though is this actually needed? Can it be removed entirely?
```
toddkazakov: Let's remove it

user/keycloak/client.go
```diff
@@ -0,0 +1,615 @@
...
+	// is computed based on the presence of a password hash in the user struct. This flag is used by
+	// other services (e.g. hydrophone) to determine whether the user is custodial or not.
+	if !user.IsCustodialAccount() {
+		user.PwHash = "true"
+	}
+
+	return user
+}
+
+func userToKeycloakUser(u *userLib.User) *keycloakUser {
```
toddkazakov: Is this used?

auth/service/api/v1/profile.go
```diff
@@ -0,0 +1,246 @@
...
+		return
+	}
+	if err := r.ProfileAccessor().UpdateUserProfileV2(ctx, userID, profile); err != nil {
+		r.handleUserOrProfileErr(responder, err)
+		return
+	}
+	responder.Data(http.StatusOK, profile)
+}
+
+func (r *Router) DeleteProfile(res rest.ResponseWriter, req *rest.Request) {
```
toddkazakov: This can be removed - seagull doesn't allow deleting user profiles

auth/store/mongo/legacy_seagull_profile_repository.go
```diff
@@ -0,0 +1,129 @@
...
+		return err
+	}
+	// Handle case where a migration was started in between the start of this function and the update
+	if updatedDoc.IsMigrating() {
+		return user.ErrUserProfileMigrationInProgress
+	}
+	return nil
+}
+
+func (p *LegacySeagullProfileRepository) DeleteUserProfile(ctx context.Context, userID string) error {
```
toddkazakov: This can be removed - seagull doesn't allow deleting user profiles

user/keycloak/client.go
```diff
@@ -0,0 +1,615 @@
...
+		return err
+	}
+	if user == nil {
+		return userLib.ErrUserNotFound
+	}
+	user.Attributes.Profile = p
+	return c.UpdateUser(ctx, user)
+}
+
+func (c *keycloakClient) DeleteUserProfile(ctx context.Context, id string) error {
```
toddkazakov: This can be removed - seagull doesn't allow deleting user profiles

user/legacy_raw_seagull_profile.go
```diff
@@ -0,0 +1,150 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/tidepool-org/platform/pointer"
+)
+
+var (
+	ErrSeagullFieldNotFound  = errors.New("seagull field not found within value object string")
```
toddkazakov: Is this used? If not, let's delete it

user/profile.go
```diff
@@ -0,0 +1,561 @@
...
+			Birthday:       up.Birthday,
+			DiagnosisDate:  up.DiagnosisDate,
+			DiagnosisType:  up.DiagnosisType,
+			TargetDevices:  up.TargetDevices,
+			TargetTimezone: up.TargetTimezone,
+			About:          up.About,
+			MRN:            up.MRN,
+			BiologicalSex:  up.BiologicalSex,
+		}
+		if up.Email != "" && !IsUnclaimedCustodialEmail(up.Email) {
```
toddkazakov: Let's remove those email from the profile if they are not used.

user/profile.go
```diff
@@ -0,0 +1,561 @@
...
+		legacyProfile.Patient.IsOtherPerson = true
+		// Handle case where Custodian user (contains fake child) and one of the FullName's is empty.
+		legacyProfile.FullName = cmp.Or(up.Custodian.FullName, up.FullName)
+		legacyProfile.Patient.FullName = pointer.FromString(cmp.Or(up.FullName, up.Custodian.FullName))
+	}
+	return legacyProfile
+}
+
+// ClearPatientInfo makes a copy of up, clearing out certain patient information - this is called usually due to lack of permissions to the patient information
+func (up *UserProfile) ClearPatientInfo() *UserProfile {
```
toddkazakov: Should the profile implement the standard `Sanitize` similar to other models?

store/structured/mongo/config.go
```diff
@@ -26,6 +26,12 @@ func LoadConfig() (*Config, error) {
 	return cfg, err
 }
 
+func LoadConfigPrefix(prefix string) (*Config, error) {
```
toddkazakov: Is this used? If not, let's delete it.

permission/permission.go
```diff
@@ -15,9 +25,17 @@ const (
...
 
+//go:generate mockgen -build_flags=--mod=mod -destination=./client_mock.go -package=permission . Client
 type Client interface {
 	GetUserPermissions(ctx context.Context, requestUserID string, targetUserID string) (Permissions, error)
 	UpdateUserPermissions(ctx context.Context, requestUserID string, targetUserID string, permissions Permissions) error
+	// GroupsForUser returns permissions that have been shared with granteeUserID. It is keyed by the user that has shared something with granteeUserID. It includes the user themself.
+	GroupsForUser(ctx context.Context, granteeUserID string) (Permissions, error)
+	// UsersInGroup returns permissions that the user with id sharerID has shared with others, keyed by user id. It includes the user themself.
+	UsersInGroup(ctx context.Context, sharerID string) (Permissions, error)
+	HasMembershipRelationship(ctx context.Context, granteeUserID, grantorUserID string) (has bool, err error)
```
toddkazakov: What do you think about renaming this to `UsersHaveSharingRelationship`? The implementation checks bi-directionally so `grantee` and `grantor` are incorrectly used

permission/permission.go
```diff
@@ -15,9 +25,17 @@ const (
 	Write     = "upload"
 )
 
+//go:generate mockgen -build_flags=--mod=mod -destination=./client_mock.go -package=permission . Client
 type Client interface {
 	GetUserPermissions(ctx context.Context, requestUserID string, targetUserID string) (Permissions, error)
 	UpdateUserPermissions(ctx context.Context, requestUserID string, targetUserID string, permissions Permissions) error
+	// GroupsForUser returns permissions that have been shared with granteeUserID. It is keyed by the user that has shared something with granteeUserID. It includes the user themself.
+	GroupsForUser(ctx context.Context, granteeUserID string) (Permissions, error)
```
toddkazakov: What do you think about renaming this `PermissionsGrantedToUser`?

permission/permission.go
```diff
@@ -15,9 +25,17 @@ const (
...
+	// UsersInGroup returns permissions that the user with id sharerID has shared with others, keyed by user id. It includes the user themself.
+	UsersInGroup(ctx context.Context, sharerID string) (Permissions, error)
```
toddkazakov: What do you think about renaming this `PermissionsGrantedByUser`?

toddkazakov (changes requested):
The biggest issue are the looser permissions in `/v1/users/:userId/users` and the lack of sanitization in `/v1/users/:userId/profile` and `/v1/users/legacy/:userId/profile`. In addition to that, I think the code can be further cleaned up.

user/user.go
lostlevels: Copy paste duplicate.

user/profile.go
lostlevels: Was originally intended to return if any profile fields were in the input attributes.

user/profile.go
lostlevels: This was for fake child migration. There is a keycloak extension to handle it.

user/keycloak/user_accessor.go
```diff
@@ -0,0 +1,90 @@
+package keycloak
+
+import (
+	"context"
+
+	"github.com/tidepool-org/platform/pointer"
+	userlib "github.com/tidepool-org/platform/user"
```
toddkazakov: The import alias is not consistent with the platform import alias conventions. I'm not sure the import alias is even needed.

user/user.go
```diff
@@ -52,11 +59,31 @@ type Client interface {
...
-	EmailVerified *bool     `json:"emailVerified,omitempty" bson:"emailVerified,omitempty"`
-	TermsAccepted *string   `json:"termsAccepted,omitempty" bson:"termsAccepted,omitempty"`
-	Roles         *[]string `json:"roles,omitempty" bson:"roles,omitempty"`
+	UserID         *string             `json:"userid,omitempty" bson:"userid,omitempty"`
+	Username       *string             `json:"username,omitempty" bson:"username,omitempty"`
+	EmailVerified  *bool               `json:"emailVerified,omitempty" bson:"emailVerified,omitempty"`
+	TermsAccepted  *string             `json:"termsAccepted,omitempty" bson:"termsAccepted,omitempty"`
+	Roles          *[]string           `json:"roles,omitempty" bson:"roles,omitempty"`
+	Emails         []string            `json:"emails,omitempty" bson:"emails,omitempty"`
+	PwHash         string              `json:"-" bson:"pwhash,omitempty"`
```
toddkazakov: This field is not used, not serialized to json and only has a bson tag but is never written to mongo. I suggest removing it.

user/user.go
```diff
@@ -52,11 +59,31 @@ type Client interface {
...
+	Hash           string              `json:"-" bson:"userhash,omitempty"`
```
toddkazakov: This field is not used, not serialized to json and only has a bson tag but is never written to mongo. I suggest removing it.

user/user.go
```diff
@@ -52,11 +59,31 @@ type Client interface {
 }
 
 type User struct {
```
toddkazakov, line 62: Does this struct need bson tags if it is guaranteed it will never touch mongo?

user/user.go
```diff
@@ -52,11 +59,31 @@ type Client interface {
...
+	Username       *string             `json:"username,omitempty" bson:"username,omitempty"`
+	EmailVerified  *bool               `json:"emailVerified,omitempty" bson:"emailVerified,omitempty"`
+	TermsAccepted  *string             `json:"termsAccepted,omitempty" bson:"termsAccepted,omitempty"`
+	Roles          *[]string           `json:"roles,omitempty" bson:"roles,omitempty"`
+	Emails         []string            `json:"emails,omitempty" bson:"emails,omitempty"`
+	PwHash         string              `json:"-" bson:"pwhash,omitempty"`
+	Hash           string              `json:"-" bson:"userhash,omitempty"`
+	Enabled        bool                `json:"-" bson:"-"`
+	Profile        *Profile            `json:"profile,omitempty" bson:"-"`
+	PasswordExists *bool               `json:"passwordExists,omitempty" bson:"-"`
```
toddkazakov: This field is only referenced in `Sanitize` - can be removed

user/user.go
```diff
@@ -52,11 +59,31 @@ type Client interface {
...
-	Username      *string   `json:"username,omitempty" bson:"username,omitempty"`
-	EmailVerified *bool     `json:"emailVerified,omitempty" bson:"emailVerified,omitempty"`
-	TermsAccepted *string   `json:"termsAccepted,omitempty" bson:"termsAccepted,omitempty"`
-	Roles         *[]string `json:"roles,omitempty" bson:"roles,omitempty"`
+	UserID         *string             `json:"userid,omitempty" bson:"userid,omitempty"`
+	Username       *string             `json:"username,omitempty" bson:"username,omitempty"`
+	EmailVerified  *bool               `json:"emailVerified,omitempty" bson:"emailVerified,omitempty"`
+	TermsAccepted  *string             `json:"termsAccepted,omitempty" bson:"termsAccepted,omitempty"`
+	Roles          *[]string           `json:"roles,omitempty" bson:"roles,omitempty"`
+	Emails         []string            `json:"emails,omitempty" bson:"emails,omitempty"`
```
toddkazakov: This field is always empty - can be removed

user/user_accessor.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// UserAccessor is the interface that can retrieve users.
+// It is the equivalent of shoreline's shoreline's Storage
+// interface, but for now will only retrieve user
+// information.
+type UserAccessor interface {
+	ProfileAccessor
+	RoleGetter
+	FindUser(ctx context.Context, user *User) (*User, error)
```
toddkazakov: Seems unused - can be removed

user/user.go
```diff
@@ -100,16 +123,70 @@ func (u *User) IsPatient() bool {
...
+	for _, u := range us {
+		if err := u.Sanitize(details); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// IsClinic returns true if the user is legacy clinic Account
+func (u *User) IsClinic() bool {
```
toddkazakov: This is not used - can be removed

user/user.go
```diff
@@ -100,16 +123,70 @@ func (u *User) IsPatient() bool {
...
+	return u.HasRole(RoleClinic)
+}
+
+func (u *User) IsCustodialAccount() bool {
```
toddkazakov: This is only used to populate `PwHash` which is not used and should be removed. This function can be removed as well.

user/user.go
```diff
@@ -100,16 +123,70 @@ func (u *User) IsPatient() bool {
...
+	return u.HasRole(RoleCustodialAccount)
+}
+
+// IsClinician returns true if the user is a clinician
+func (u *User) IsClinician() bool {
```
toddkazakov: This is not used - can be removed

user/user.go
```diff
@@ -100,16 +123,70 @@ func (u *User) IsPatient() bool {
...
+	return u.HasRole(RoleClinician)
+}
+
+func (u *User) AreTermsAccepted() bool {
```
toddkazakov: This is not used - can be removed

user/user_accessor.go
```diff
@@ -0,0 +1,87 @@
...
+	FindUserById(ctx context.Context, id string) (*User, error)
+	FindUsersWithIds(ctx context.Context, ids []string) ([]*User, error)
```
toddkazakov, line 47: Seems unused - can be removed

user/user_accessor.go
```diff
@@ -0,0 +1,87 @@
...
+
+// UserAccessor is the interface that can retrieve users.
+// It is the equivalent of shoreline's shoreline's Storage
+// interface, but for now will only retrieve user
+// information.
+type UserAccessor interface {
+	ProfileAccessor
+	RoleGetter
+	FindUser(ctx context.Context, user *User) (*User, error)
+	FindUserById(ctx context.Context, id string) (*User, error)
```
toddkazakov:
```suggestion
	Get(ctx context.Context, id string) (*User, error)
```

user/user_accessor.go
```diff
@@ -0,0 +1,87 @@
...
+	FindUsersWithIds(ctx context.Context, ids []string) ([]*User, error)
+}
+
+type TokenIntrospectionResult struct {
```
toddkazakov: Does this belong to the top-level user package? Can it be moved to `user/keycloak`?

user/user_accessor.go
```diff
@@ -0,0 +1,87 @@
...
+	Active           bool        `json:"active"`
+	Subject          string      `json:"sub"`
+	EmailVerified    bool        `json:"email_verified"`
+	ExpiresAt        int64       `json:"eat"`
+	RealmAccess      RealmAccess `json:"realm_access"`
+	IdentityProvider string      `json:"identityProvider"`
+}
+
+type AccessTokenCustomClaims struct {
```
toddkazakov: Does this belong to the top-level user package? Can it be moved to `user/keycloak`?

user/user_accessor.go
```diff
@@ -0,0 +1,87 @@
...
+	jwx.Claims
+	IdentityProvider string `json:"identity_provider,omitempty"`
+}
+
+type RealmAccess struct {
```
toddkazakov: Does this belong to the top-level user package? Can it be moved to `user/keycloak`?

user/user.go
```diff
@@ -100,16 +123,70 @@ func (u *User) IsPatient() bool {
 	return false
 }
 
+func IsUnclaimedCustodialEmail(email string) bool {
```
toddkazakov, line 123: This function is unused, but I believe it should be called in keycloak client to remove the email from the user object if it is an unclaimed custodial email. Wdyt?

user/user.go
```diff
@@ -142,4 +219,7 @@ func ValidateID(value string) error {
 	return nil
 }
 
-var idExpression = regexp.MustCompile(`^([0-9a-f]{10}|[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12})$`)
+// IsValidUserID return true if the string is in a human readable uuid hex 8-4-4-4-12 format or legacy alphanumeric 10 characters
+func IsValidUserID(id string) bool {
```
toddkazakov, line 205: This is duplicate of `auth.IsValidUserID`. Can we somehow consolidate those?

user/user.go
```diff
@@ -24,15 +26,20 @@ const (
...
+		RoleBrokered:         struct{}{},
+		RoleCarePartner:      struct{}{},
+		RoleClinic:           struct{}{},
+		RoleClinician:        struct{}{},
+		RoleCustodialAccount: struct{}{},
+		RoleDemo:             struct{}{},
+		RolePatient:          struct{}{},
+	}
+
+	idExpression           = regexp.MustCompile(`^([0-9a-f]{10}|[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12})$`)
```
toddkazakov: This is a duplicate of the regexp in the `auth` package. Can we somehow consolidate those?

user/keycloak/client.go
```diff
@@ -177,6 +177,7 @@ func (c *keycloakClient) UpdateUser(ctx context.Context, user *userlib.User) err
 	}
 
 	attrs := map[string][]string{}
+	maps.Copy(attrs, user.Attributes)
```
toddkazakov: I know this was added as a result of my previous review where I flagged an issue that removes non-profile attributes from the keycloak user, but this introduces a new bug. Profile fields cannot be removed from the user profile because only non-empty attributes are added to the map in `user.Profile.ToAttributes()`.

user/user_accessor.go
toddkazakov: Actually, you can just embed the `user.Client` interface

user/timeutil.go
```diff
@@ -0,0 +1,30 @@
+package user
```
toddkazakov: Can this file be moved to the keycloak package?

user/profile.go
toddkazakov: Instead of using a boolean attribute can we just add the custodial role to the account instead?

user/store/mongo/legacy_seagull_profile_repository.go
```diff
@@ -0,0 +1,114 @@
+package mongo
```
toddkazakov, line 1: This file seems misplaced - should it be in the `user` package?

toddkazakov (changes requested):
There's a new bug when merging the existing user attributes with the updated profile attributes, which makes it impossible to remove an attribute from the user profile.

I feel that the code should be cleaned up further - there's a lot of unused cruft that has been copied over from shoreline.

user/user.go
lostlevels: Not needed, removed.

user/timeutil.go
lostlevels: Yes.

user/user_accessor.go
lostlevels: It appears unused, removing.

user/user_accessor.go
lostlevels: It appears unused, removing.

user/user.go
lostlevels: Yup

user/user.go
lostlevels: Added to the `User.Sanitize` and `TrustUser.Sanitize` and tests.

auth/service/api/v1/router_test.go
```diff
@@ -45,6 +70,983 @@ var _ = Describe("Router", func() {
...
+							Roles:         &unclaimedRoles,
+							Profile:       &unclaimedCustodialProfile,
+						}
+						sanitizedUnclaimedCustodialDetails = &user.User{
+							UserID:        pointer.FromString(unclaimedCustodialUserID),
+							Username:      nil,
+							EmailVerified: pointer.FromBool(false),
+							Roles:         &unclaimedRoles,
+							Profile:       &unclaimedCustodialProfile,
+						}
```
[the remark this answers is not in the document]
lostlevels: Ignoring: This is a ginkgo-ism lifecycle pattern for better or worse. Each variable in BeforeEach is a copy for each spec.

store/structured/mongo/config.go
```diff
@@ -72,7 +72,11 @@ func (c *Config) AsConnectionString() string {
 }
 
 func (c *Config) Load() error {
-	return envconfig.Process("", c)
+	return c.LoadPrefix("")
+}
+
+func (c *Config) LoadPrefix(prefix string) error {
+	return envconfig.Process(prefix, c)
 }
```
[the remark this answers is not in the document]
lostlevels: This is incorrect as the `required` tag means envconfig will error out if the prefixed version is not set.

user/keycloak/client.go
```diff
@@ -0,0 +1,520 @@
...
+	}
+
+	if err := c.keycloak.LogoutAllSessions(ctx, token.AccessToken, c.cfg.Realm, id); err != nil {
+		if aErr, ok := err.(*gocloak.APIError); ok && aErr.Code == http.StatusNotFound {
+			return nil
+		}
+	}
+
+	return err
+}
```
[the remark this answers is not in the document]
lostlevels: Resolved by deleting the function as it's not used.

user/store/mongo/legacy_seagull_profile_repository.go
```diff
@@ -0,0 +1,114 @@
...
+	var updatedDoc user.LegacySeagullDocument
+	err = p.FindOneAndUpdate(ctx, uselector, update, uopts).Decode(&updatedDoc)
+	if err != nil {
+		return err
+	}
+	// Handle case where a migration was started in between the start of this function and the update
+	if updatedDoc.IsMigrating() {
+		return user.ErrUserProfileMigrationInProgress
+	}
+	return nil
```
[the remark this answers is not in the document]
lostlevels: I'm working under the assumption that the probability of an update during a migration of an individual profile is low.

user/legacy_raw_seagull_profile.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		valueObj = map[string]any{}
+	}
+	valueObj["profile"] = profile
+	bytes, err := json.Marshal(valueObj)
+	if err != nil {
+		return "", err
+	}
+	return string(bytes), nil
+}
```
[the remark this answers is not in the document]
lostlevels: This is not quite right. If the JSON string in `value` is invalid JSON, clients that read the profile would never get any "partial" settings either way since their value is in a bad state. Creating a new object for the profile is fine since an invalid `value` is effectively a valueless profile.

lostlevels:
@toddkazakov Fixes added, though I may do another pass and do some refactoring after.

user/keycloak/client.go
```diff
@@ -0,0 +1,498 @@
...
+	if err != nil {
+		return
+	}
+	t := time.Unix(i, 0)
+	timestamp = t.Format(timestampFormat)
+	return
+}
+
+func timestampStringToUnixTime(timestamp string) (unixTime string, err error) {
+	tm, err := time.Parse(timestamp, timestampFormat)
```
toddkazakov, line 492:
```suggestion
	tm, err := time.Parse(timestampFormat, timestamp)
```

auth/service/api/v1/profile.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	details := request.GetAuthDetails(ctx)
+	if details == nil {
+		return nil, request.ErrorUnauthenticated()
+	}
+	if details.IsService() || details.UserID() == userID {
+		return profile, nil
+	}
+	trustorPerms, err := r.PermissionsClient().GetUserPermissions(ctx, details.UserID(), userID)
+	if err != nil {
```
toddkazakov, line 68: A 404 returned by gatekeeper will result in a 500 error here because the client transforms a 404 in an unauthorized error. We should return the sanitized profile instead.

auth/service/api/v1/permission.go
```diff
@@ -0,0 +1,84 @@
...
+			if details == nil {
+				request.MustNewResponder(res, req).Error(http.StatusUnauthorized, request.ErrorUnauthenticated())
+				return
+			}
+			if details.IsService() || details.UserID() == targetUserID {
+				handlerFunc(res, req)
+				return
+			}
+			hasPerms, err := r.PermissionsClient().HasCustodianPermissions(ctx, details.UserID(), targetUserID)
+			if err != nil {
```
toddkazakov, line 36: `HasCustodianPermissions` will return a non-nil error if gatekeeper returns 404. I believe the unauthorized error should be handled - here or inside `HasCustodianPermissions`

auth/service/service/service.go
```diff
@@ -795,6 +841,46 @@ func (s *Service) initializeUserEventsHandler() error {
...
+		return errors.New("empty user accessor passed to initializeUserProfileAccessor")
+	}
+	cfg := storeStructuredMongo.NewConfig()
+	// Note the "SEAGULL" prefix, this is so that the regular env vars
+	// for mongo access such as TIDEPOOL_STORE_SCHEME are
+	// SEAGULL_TIDEPOOL_STORE_SCHEME so as to not conflict with existing
+	// TIDEPOOL_STORE_SCHEME values. This is done instead of using a
+	// seagull client as seagull will eventually be removed so no sense
+	// in keeping it around.
+	if err := cfg.LoadPrefix("SEAGULL"); err != nil {
```
toddkazakov, line 869: envconfig falls back to the unprefixed variable when the prefixed one is missing, so this doesn't do what it is expected to do in case the variable is missing

user/user.go
```diff
@@ -26,15 +28,20 @@ const (
...
+		RoleBrokered:         struct{}{},
+		RoleCarePartner:      struct{}{},
+		RoleClinic:           struct{}{},
+		RoleClinician:        struct{}{},
+		RoleCustodialAccount: struct{}{},
+		RoleDemo:             struct{}{},
+		RolePatient:          struct{}{},
+	}
+
+	IdExpression           = regexp.MustCompile(`^([0-9a-f]{10}|[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12})$`)
```
toddkazakov, line 42: Claude flagged that this new regexp now allows uppercase user ids. Is there a good reason to relax the validation?

user/user_accessor.go
toddkazakov: @lostlevels Is there a reason to keep this around?

toddkazakov (changes requested):

user/user.go
lostlevels: No reason.

auth/service/api/v1/permission.go
lostlevels: Handled in `HasCustodianPermissions` now.

user/user_accessor.go
lostlevels: No

PR #700, opened 2024-03-01
inject a provenance subdocument to device data

ewollesen:
BACK-2504

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ format-write-changed:
 imports: goimports
 	@echo "goimports -d -e -local 'github.com/tidepool-org/platform'"
 	@cd $(ROOT_DIRECTORY) && \
-		O=`find . -not -path './vendor/*' -name '*.go' -type f -exec goimports -d -e -local 'github.com/tidepool-org/platform' {} \; 2>&1` && \
+		O=`find . -not -path './vendor/*' -not -path '**/test/mock.go' -name '*.go' -type f -exec goimports -d -e -local 'github.com/tidepool-org/platform' {} \; 2>&1` && \
 		[ -z "$${O}" ] || (echo "$${O}" && exit 1)
 
 imports-write: goimports
--- a/data/service/api/v1/datasets_data_create.go
+++ b/data/service/api/v1/datasets_data_create.go
@@ -1,16 +1,21 @@
 package v1
 
 import (
+	"context"
+	"net"
 	"net/http"
 	"strconv"
+	"strings"
 
-	"github.com/tidepool-org/platform/data/summary"
+	"github.com/ant0ine/go-json-rest/rest"
+	"github.com/golang-jwt/jwt/v4"
 
-	"github.com/tidepool-org/platform/data/summary/types"
-
+	"github.com/tidepool-org/platform/auth"
 	"github.com/tidepool-org/platform/data"
 	dataNormalizer "github.com/tidepool-org/platform/data/normalizer"
 	dataService "github.com/tidepool-org/platform/data/service"
+	"github.com/tidepool-org/platform/data/summary"
+	"github.com/tidepool-org/platform/data/summary/types"
 	dataTypesFactory "github.com/tidepool-org/platform/data/types/factory"
 	"github.com/tidepool-org/platform/log"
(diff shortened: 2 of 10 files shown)


data/service/api/v1/datasets_data_create.go
```diff
@@ -95,10 +94,18 @@ func DataSetsDataCreate(dataServiceContext dataService.Context) {
 		return
 	}
 
+	var provenance *data.Provenance
+	if claims, err := dataServiceContext.AuthClient().ServerSessionTokenClaims(ctx); err != nil {
```
toddkazakov, line 98: This doesn't look correct - we should get the client of the access token that was passed in the `x-tidepool-session-token` header.

ewollesen: Yeah, you caught this for review before it was ready, sorry... But all the same, you're right, that's the wrong token from which to grab the client id. 😄 

I'll move it to a draft.

Makefile
```diff
@@ -112,7 +112,7 @@ format-write-changed:
 imports: goimports
 	@echo "goimports -d -e -local 'github.com/tidepool-org/platform'"
 	@cd $(ROOT_DIRECTORY) && \
-		O=`find . -not -path './vendor/*' -name '*.go' -type f -exec goimports -d -e -local 'github.com/tidepool-org/platform' {} \; 2>&1` && \
+		O=`find . -not -path './vendor/*' -not -path '**/test/mock.go' -name '*.go' -type f -exec goimports -d -e -local 'github.com/tidepool-org/platform' {} \; 2>&1` && \
```
ewollesen, line 115: To add a little detail, not only are they generated in a way that goimports doesn't like, but then Travis fails the build as a result.

data/service/api/v1/datasets_data_create_provenance_test.go
```diff
@@ -0,0 +1,145 @@
+package v1
```
toddkazakov, line 1: Package should be `v1_test` for consistency

toddkazakov (changes requested):
Looks good overall. I believe we need to set the provenance to data sets (upload records) as well.

data/service/api/v1/datasets_data_create_provenance_test.go
ewollesen: If I do that, then I can't call `selectXFF` or 'collectProvenanceInfo`, as they're not exported, and there's no reason for them to be.

I'm all for testing the exported APIs, but personally I have no problem with going beyond and testing non-exported functions too.

I suppose I could scaffold a bunch of junk to test selectXFF or collectProvenanceInfo via the exported API, but what does that get me? My selectXFF tests get a lot more complicated and brittle, and if selectXFF changes, I'll still have to fix all the tests.

So I'd rather keep the package v1, but if consistency is more important, reply back and I'll do the scaffolding route... Unless you have a better idea?

data/service/api/v1/datasets_data_create.go
```diff
@@ -124,3 +132,83 @@ func DataSetsDataCreate(dataServiceContext dataService.Context) {
...
+
+// collectProvenanceInfo from a request and its auth details.
+//
+// All work is done as a best effort right now.
+func collectProvenanceInfo(ctx context.Context, req *rest.Request, details request.AuthDetails) *data.Provenance {
+	lgr := log.LoggerFromContext(ctx)
+	provenance := &data.Provenance{}
+
+	token := ""
+	if _, ok := req.Header[auth.TidepoolSessionTokenHeaderKey]; ok {
```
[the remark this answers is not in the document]
ewollesen: Fixed

ewollesen:
> Looks good overall. I believe we need to set the provenance to data sets (upload records) as well.

OK, will do.

data/service/api/v1/datasets_data_create_provenance_test.go
toddkazakov: This exact same issue tripped me [here](https://github.com/tidepool-org/clinic/commit/5e1381a214be06dc12963906c2dcbef3ab82f22f). I don't see any harm in exporting those functions.

toddkazakov (approved):

PR #701, opened 2024-03-06
[BACK-2830][BACK-2840] Add RealtimeRecords to summaries and reporting API

Roukoswarf:


data/store/mongo/mongo_data.go
```diff
@@ -40,20 +40,24 @@ func (d *DataRepository) ListUserDataSets(ctx context.Context, userID string, fi
...
-	// Can read only from the new deviceDataSets collection via DataSetRepository when migration completed.
-	dataSet, err := d.DataSetRepository.GetDataSet(ctx, id)
+	return d.DataSetRepository.GetDataSet(ctx, id)
+}
+
+func (d *DataRepository) GetDataSetByID(ctx context.Context, dataSetID string) (*upload.Upload, error) {
+	return d.DataSetRepository.GetDataSetByID(ctx, dataSetID)
+}
+
+func (d *DataRepository) IsDataSetAutomated(ctx context.Context, dataSetID string) (bool, error) {
```
toddkazakov, line 50: I don't think we should introduce a new term `automated` to avoid confusion. We are referring to those datasets as `continuous`.

data/store/store.go
```diff
@@ -32,6 +30,7 @@ type DataSetRepository interface {
 
 	GetDataSetsForUserByID(ctx context.Context, userID string, filter *Filter, pagination *page.Pagination) ([]*upload.Upload, error)
 	GetDataSetByID(ctx context.Context, dataSetID string) (*upload.Upload, error)
+	IsDataSetAutomated(ctx context.Context, dataSetID string) (bool, error)
```
toddkazakov, line 33: I'm not entirely sure this required. You can use `GetDataSetById` and add a method `IsContinuous()` on the struct instead of adding this to the repository.

data/summary/types/cgm.go
```diff
@@ -35,6 +35,9 @@ type CGMBucketData struct {
 	TotalGlucose float64 `json:"totalGlucose" bson:"totalGlucose"`
 	TotalMinutes int     `json:"totalMinutes" bson:"totalMinutes"`
 	TotalRecords int     `json:"totalRecords" bson:"totalRecords"`
+
+	RealtimeRecords int `json:"realtimeRecords" bson:"realtimeRecords"`
```
toddkazakov, line 39: I think you should leave a comment here describing what's a realtime record and what's a deferred record.

data/summary/types/summary.go
```diff
@@ -73,16 +80,11 @@ type DeviceDataCursor interface {
...
-	FirstData time.Time
-	LastData  time.Time
-
-	EarliestModified time.Time
-
-	LastUpload time.Time
-
-	LastUpdated     time.Time
-	NextLastUpdated time.Time
+type DeviceDataManager interface {
```
toddkazakov, line 83: Consider renaming this - "manage" implies mutating the data, but this interface only queries the data store. Is this a `DataFetcher` for summary data?

data/summary/types/summary.go
```diff
@@ -161,7 +168,7 @@ type StatsPt[T Stats] interface {
 	Init()
 	GetBucketsLen() int
 	GetBucketDate(int) time.Time
-	Update(context.Context, DeviceDataCursor) error
+	Update(context.Context, DeviceDataCursor, DeviceDataManager) error
```
toddkazakov, line 171: Is there a way this can be simplified to only use the device data manager instead of a cursor and a manager?

data/summary/types/summary.go
```diff
@@ -346,7 +353,7 @@ func removeExcessBuckets[A BucketDataPt[T], T BucketData](buckets *[]*Bucket[A,
 	*buckets = (*buckets)[excess:]
 }
 
-func AddData[A BucketDataPt[T], T BucketData, R RecordTypes, D RecordTypesPt[R]](buckets *[]*Bucket[A, T], userData []D) error {
+func AddData[A BucketDataPt[T], T BucketData, R RecordTypes, D RecordTypesPt[R]](buckets *[]*Bucket[A, T], userData []D, realtimeUpload map[string]bool) error {
```
toddkazakov, line 356: Consider adding a type alias with a `IsRealtimeUpload(uploadId)` method to improve readability.

data/summary/store/store.go
```diff
@@ -80,6 +82,144 @@ func (r *TypelessRepo) DeleteSummary(ctx context.Context, userId string) error {
...
+					i -= indexA + 1
+				}
+			}
+		}
+	}
+
+	return count
+}
+
+func (r *TypelessRepo) GetRealtimePatients(ctx context.Context, userIds []string, startTime time.Time, endTime time.Time) (map[string]int, error) {
```
toddkazakov, line 145: Consider renaming this to `GetPatientsWithRealtimeData`

data/summary/store/store.go
```diff
@@ -80,6 +82,144 @@ func (r *TypelessRepo) DeleteSummary(ctx context.Context, userId string) error {
 	return nil
 }
 
+// isRealtimeUser processes two slices of Item and returns an int count of days with realtime records.
+// this currently doesn't handle N slices, only 1-2, might need adjustment for more types.
+func isRealtimeUser(a, b []*types.Bucket[*types.CGMBucketData, types.CGMBucketData]) int {
```
toddkazakov, line 87: The name and the signature are off. If a function starts with `is` it should return a boolean. Consider renaming this to `GetNumberOfDaysWithRealtimeData`.

toddkazakov: It is also not clear why you need to pass two slices.

data/service/api/v1/summary.go
```diff
@@ -79,6 +85,109 @@ func GetSummary[T types.Stats, A types.StatsPt[T]](dataServiceContext dataServic
...
+	d.StartTime = parser.Time("startDate", time.RFC3339)
+	d.EndTime = parser.Time("endDate", time.RFC3339)
+}
+
+func (d *RealtimePatientsFilter) Validate(validator structure.Validator) {
+	validator.Time("startDate", d.StartTime).NotZero()
+	validator.Time("endDate", d.EndTime).NotZero()
+}
+
+func GetRealtimePatients(dataServiceContext dataService.Context) {
```
toddkazakov, line 129: Consider renaming this to `GetPatientsWithRealtimeData`.

clinics/service.go
```diff
@@ -155,3 +156,19 @@ func (d *defaultClient) getPatient(ctx context.Context, clinicID, patientID stri
 	}
 	return response.JSON200, nil
 }
+
+func (d *defaultClient) GetPatients(ctx context.Context, clinicId string) ([]clinic.Patient, error) {
+	params := &clinic.ListPatientsParams{
+		Limit: pointer.FromAny(1001),
```
toddkazakov, line 162: The limit should be passed as an argument

data/service/api/v1/summary.go
```diff
@@ -79,6 +85,109 @@ func GetSummary[T types.Stats, A types.StatsPt[T]](dataServiceContext dataServic
...
+	EndDate       time.Time `json:"endDate"`
+}
+
+type RealtimePatientResponse struct {
+	Id               string    `json:"id"`
+	FullName         string    `json:"fullName"`
+	BirthDate        time.Time `json:"birthDate"`
+	MRN              *string   `json:"mrn"`
+	RealtimeDays     int       `json:"realtimeDays"`
+	IsSufficientData bool      `json:"isSufficientData"`
```
toddkazakov, line 107: Should this be `hasSufficientData`?

data/service/api/v1/summary.go
```diff
@@ -79,6 +85,109 @@ func GetSummary[T types.Stats, A types.StatsPt[T]](dataServiceContext dataServic
...
+			BirthDate:        patients[i].BirthDate.Time,
+			MRN:              patients[i].Mrn,
+			RealtimeDays:     userIdsRealtimeDays[*patients[i].Id],
+			IsSufficientData: userIdsRealtimeDays[*patients[i].Id] >= store.RealtimeUserThreshold,
+		}
+	}
+
+	response := RealtimePatientsResponse{
+		Config: RealtimePatientConfigResponse{
+			SchemaVersion: 1,
```
toddkazakov, line 179: This should be a const and probably should be removed if it's not used.

data/service/api/v1/summary.go
```diff
@@ -79,6 +85,109 @@ func GetSummary[T types.Stats, A types.StatsPt[T]](dataServiceContext dataServic
...
+			Code:          "CPT-99454",
```
toddkazakov, line 180: This should be a `const`.

data/summary/summary.go
```diff
@@ -89,10 +102,18 @@ func (c *GlucoseSummarizer[T, A]) GetMigratableUserIDs(ctx context.Context, pagi
 	return c.summaries.GetMigratableUserIDs(ctx, pagination)
 }
 
+func (c *Manager) GetRealtimePatients(ctx context.Context, userIds []string, startTime time.Time, endTime time.Time) (map[string]int, error) {
```
toddkazakov, line 105: The pointer receiver should be changed to `m` and the manager constructor and functions should be moved in this file next to each other.

data/summary/summary.go
```diff
@@ -21,12 +21,15 @@ const (
 
 type SummarizerRegistry struct {
 	summarizers map[string]any
+	Manager     *Manager
```
toddkazakov, line 24: I don't think the manager should be in the registry

data/summary/summary.go
```diff
@@ -89,10 +102,18 @@ func (c *GlucoseSummarizer[T, A]) GetMigratableUserIDs(ctx context.Context, pagi
 	return c.summaries.GetMigratableUserIDs(ctx, pagination)
 }
 
+func (c *Manager) GetRealtimePatients(ctx context.Context, userIds []string, startTime time.Time, endTime time.Time) (map[string]int, error) {
+	return c.summaries.GetRealtimePatients(ctx, userIds, startTime, endTime)
+}
+
+func (c *Manager) DeleteSummaries(ctx context.Context, userId string) error {
```
toddkazakov, line 109: Is this method used?

data/summary/summary.go
```diff
@@ -55,21 +58,31 @@ var _ Summarizer[types.CGMStats, *types.CGMStats] = &GlucoseSummarizer[types.CGM
...
+func NewCGMSummarizer(collection *storeStructuredMongo.Repository, dataRepo dataStore.DataRepository) Summarizer[types.CGMStats, *types.CGMStats] {
 	return &GlucoseSummarizer[types.CGMStats, *types.CGMStats]{
-		deviceData: deviceData,
-		summaries:  store.New[types.CGMStats, *types.CGMStats](collection),
+		userData:  dataRepo,
+		summaries: store.New[types.CGMStats, *types.CGMStats](collection),
+	}
+}
+
+func NewManager(collection *storeStructuredMongo.Repository) *Manager {
```
toddkazakov, line 83: "Manager" is very generic. Consider moving all of the business logic from the API layer to this struct and renaming this to `ClinicRealtimeDataReportGenerator`.

toddkazakov (changes requested):

data/store/mongo/mongo_data.go
Roukoswarf: The definition of "automated" contained many more conditions originally which was under debate, so it used to be distinct, but now I agree this is silly.

I have added `IsContinuous()` to `Upload` type.

data/store/store.go
Roukoswarf: Added `IsContinuous()` to `Upload` type.

data/summary/types/cgm.go
Roukoswarf: Good idea, added.

data/summary/types/summary.go
Roukoswarf: Thats a better name, changed.

data/summary/store/store.go
Roukoswarf: yeah, this is an artifact of me not reading the expected return json properly and thinking I only needed a boolean, i changed the function without changing the name, ill make it match.

as for the two slices, this is an optimization/simplification for now, as either bgm or cgm buckets can trigger the day marking, so doing one slice at a time would require returning a list of days, then de-duplicating. As a single pass, we can skip the rest of the buckets for the day for both types together, simplifying it to a counter.

In the future, which is scheduled for the Insulin Stats work, this will be moved to the DB with a single map reduce query or similar, as they will be in a single collection, making this much more straight forward.

data/service/api/v1/summary.go
Roukoswarf: was copied from the response json provided by frontend in the ticket, but it is a bit weird sounding, ill change it.

data/service/api/v1/summary.go
Roukoswarf: also copied from frontend example, but yes, i dont expect it to be needed, ill remove for now.

data/summary/summary.go
Roukoswarf: No, but the function below it exists, i thought it a hole to not have it like all other functions.
But I don't want to expose it on the API yet today.

data/summary/types/summary.go
Roukoswarf: That makes sense, plus allows for handling the missing uploadId case cleanly, which is an assumption the function makes about its input, we can make it sparse with this.

data/summary/summary.go
Roukoswarf: the manager is more accurately `TypelessSummarizer`, and ive renamed it as that, but GetSummarizer/addSummarizer are generics with specific requirements so I didnt see a clean way without another property.

ideally it would be added to the summarizers like everything else, but it doesnt have a GetSummary/UpdateSummary, since its all calculated live.

toddkazakov (changes requested):
I think this PR needs serious redesign. This implementation has many leaky abstractions because of the decision to overload existing CGM and BGM summaries with realtime buckets. This has led to complicated code without clear separation of concerns. 

This is a high-level summary of the design problems I found with the current implementation:
- The API layer is overloaded with the responsibility to call the clinic service and execute logic for each patient of the clinic, etc. This clearly is business logic so it should be moved elsewhere and encapsulated in a new service. 
- The business layer now has a new concept of a typeless summarizer which is introduced to bypass the typed business layer (direct consequence of storing realtime buckets for a single user in multiple summary types). The idea behind summaries generic was to be able to easily add new summary types by only implementing the parts where summaries differ.
- CGM and BGM stats calculator have conditional logic based on the whether the data passed comes from continuous or normal dataset. This logic is triggered based on a single boolean variable. This is a code smell - it's the responsibility of the caller to pass the correct data to be summarized.
- The persistence layer now merges data from two summaries (and as you pointed out uses incorrectly typed slices for bgm data). Persistence layer should be much simpler - e.g. fetch data for a given type and period.

I have also left few inline comments.

---  

All of those issues can be resolved by introducing a new type of summary -`RealtimeDataSummary`. The responsibility of this summary would be keep track of the hourly buckets with realtime data. The aggregated stats can be empty if you don't need them. You can set this new summary as outdated only if a user uploads cgm or bgm data in a continuous dataset (so you don't waste compute cycles for non-continuous data)

Similar to how you have a `CGMSummarizer` and `BGMSummarizer` you should add a new `RealtimeDataSummarizer`. The new summarizer should call a new `ContinuousDataFetcher` which will only return data from continuous datasets (e.g. by filtering out the data based on the upload id). 

You should introduce a new service (e.g. `RealtimeDataReporter`) which will have all of the business logic for generating such a report. It will:
- Get the list of patients for a given clinic
- Throw an error if the clinic has more than 1000 patients
- Fetch the `RealtimeDataSummary` for each user
- Compile a report for all users with realtime data
- Return this data to API layer so it can be marshalled as json (or csv if needed)

The new Request/Response structs should be moved higher up similar to other platform services and should be named accordingly.

data/service/api/v1/summary.go
```diff
@@ -31,6 +37,8 @@ func SummaryRoutes() []dataService.Route {
 
 		dataService.Get("/v1/summaries/migratable/cgm", GetMigratableUserIDs[types.CGMStats, *types.CGMStats], api.RequireAuth),
 		dataService.Get("/v1/summaries/migratable/bgm", GetMigratableUserIDs[types.BGMStats, *types.BGMStats], api.RequireAuth),
+
+		dataService.Get("/v1/summaries/realtime/:clinicId", GetPatientsWithRealtimeData, api.RequireAuth),
```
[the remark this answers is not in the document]
Roukoswarf: Is this desired schema? It would be mixing routes where /clinic* no longer goes specifically to the clinic service, I'm not sure whats better, I understand the intent though.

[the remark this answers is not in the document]
toddkazakov: I think so, yes. It better represents the relationship between the clinic and the generated report. [This](https://github.com/tidepool-org/development/blob/master/charts/tidepool/charts/hydrophone/templates/4-routetable.yaml#L165-L173) is an example where this pattern is already used.

data/summary/summary.go
Roukoswarf: continuous type was split from bgm/cgm, this is no longer relevant.

data/summary/summary.go
Roukoswarf: `Reporter` class functionality now replaces this

data/service/api/v1/summary.go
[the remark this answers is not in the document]
Roukoswarf: done

clinics/service.go
```diff
@@ -155,3 +156,23 @@ func (d *defaultClient) getPatient(ctx context.Context, clinicID, patientID stri
...
 	return response.JSON200, nil
 }
+
+func (d *defaultClient) GetPatients(ctx context.Context, clinicId string, userToken string) ([]clinic.Patient, error) {
+	params := &clinic.ListPatientsParams{
+		Limit: pointer.FromAny(1001),
+	}
+
+	response, err := d.httpClient.ListPatientsWithResponse(ctx, clinicId, params, func(ctx context.Context, req *http.Request) error {
+		req.Header.Add(auth.TidepoolSessionTokenHeaderKey, userToken)
```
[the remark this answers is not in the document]
Roukoswarf: done.

data/summary/summary.go
```diff
@@ -20,84 +19,105 @@ const (
 )
 
 type SummarizerRegistry struct {
-	summarizers map[string]any
+	summarizers        map[string]any
+	TypelessSummarizer *TypelessSummarizer
```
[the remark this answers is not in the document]
Roukoswarf: refactored out.

data/service/api/v1/summary.go
```diff
@@ -79,6 +87,108 @@ func GetSummary[T types.Stats, A types.StatsPt[T]](dataServiceContext dataServic
...
+	startTime := time.Now().UTC().AddDate(0, 0, -60)
+	endTime := time.Now().UTC()
+
+	details := request.GetAuthDetails(ctx)
+	if !details.IsService() {
+		dataServiceContext.RespondWithError(service.ErrorUnauthorized())
+		return
+	}
+
+	patients, err := dataServiceContext.ClinicsClient().GetPatients(ctx, clinicId, details.Token())
```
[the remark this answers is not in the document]
Roukoswarf: I think I have an okay fix for this, but I would appreciate any additional comments, as its not the prettiest due to the passing of the clinic client and token down further than I would like to accomplish this.

data/data.go
```diff
@@ -132,3 +133,15 @@ func ErrorValueStringAsIDNotValid(value string) error {
 }
 
 var idExpression = regexp.MustCompile("^[0-9a-z]{32}$") // TODO: Want just "[0-9a-f]{32}" (Jellyfish uses [0-9a-z])
+
+type UserLastUpdated struct {
```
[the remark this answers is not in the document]
Roukoswarf: this was relocated out of the summary package because of a new circular import regarding this struct, I have renamed it and added comments.

data/summary/types/continuous.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	s.Buckets = s.Buckets[:offset]
+
+	if len(s.Buckets) > 0 {
+		return s.Buckets[len(s.Buckets)-1].LastRecordTime
+	}
+	return
+}
+
+func (s *ContinuousStats) Update(ctx context.Context, cursor DeviceDataCursor, dataRepo DeviceDataFetcher) error {
```
toddkazakov, line 91: The cursor here should only return data relevant for calculating the summary. You can further simplify this by creating a new struct which encapsulates the logic for filtering out data from non-continuous datasets.

```go
type DeviceDataCursor[T Datum] interface {
  HasNext(context.Context) bool
  GetNextData() []T, error
  RemainingBatchLength() int
}

type ContinuousDeviceDataCursor[T Datum] struct {
  cursor DeviceDataCursor
  dataRepo DataRepository
}

func (c *ContinuousDeviceDataCursor) GetNextData() ([]T, error) {
  // Check if each datum is continuous, filter out if not
}

func (c *ContinuousDeviceDataCursor) IsUploadContinuous(uploadId string) bool {
  // Cache it here
}
```

data/summary/types/summary.go
```diff
@@ -21,9 +29,10 @@ import (
 )
 
 const (
-	SummaryTypeCGM = "cgm"
-	SummaryTypeBGM = "bgm"
-	SchemaVersion  = 3
+	SummaryTypeCGM        = "cgm"
+	SummaryTypeBGM        = "bgm"
+	SummaryTypeContinuous = "continuous"
+	SchemaVersion         = 4
```
toddkazakov, line 35: Do we need to bump up the version?

data/store/mongo/mongo_datum.go
```diff
@@ -601,7 +557,7 @@ func (d *DatumRepository) GetDataRange(ctx context.Context, userId string, typ s
 		return nil, errors.New("userId is empty")
 	}
 
-	if typ == "" {
+	if len(typ) < 1 {
```
toddkazakov, line 560:
```suggestion
	if len(typ) == 0 {
```

data/store/mongo/mongo_datum.go
```diff
@@ -739,18 +711,23 @@ func (d *DatumRepository) populateLastUpload(ctx context.Context, userId string,
...
 		"_userId": userId,
-		"type":    typ,
+		"_active": bson.M{"$in": bson.A{true, false}},
 		"time": bson.M{
 			"$gte": status.FirstData,
 			"$lte": status.LastData,
 		},
 	}
 
+	if len(typ) > 1 {
```
toddkazakov, line 725: The special case here should be `len(typ) == 1`, because `$in` can handle empty arrays. Currently, there is a bug when `typ` is `nil` or `empty`. In addition to adding a guard for the length I think you should flip the logic.

data/store/mongo/mongo_datum.go
```diff
@@ -698,17 +664,23 @@ func (d *DatumRepository) getTimeRange(ctx context.Context, userId string, typ s
...
-		"_active": bson.M{"$ne": -1111},
-		"type":    typ,
+		"_active": bson.M{"$in": bson.A{true, false}},
 		"time": bson.M{
 			"$gte": status.FirstData,
 			"$lte": status.LastData,
 		},
 	}
+
+	if len(typ) > 1 {
```
toddkazakov, line 678: The special case here should be `len(typ) == 1`, because `$in` can handle empty arrays. Currently, there is a bug when `typ` is `nil` or `empty`. In addition to adding a guard for the length I think you should flip the logic.

data/store/mongo/mongo_datum.go
```diff
@@ -667,13 +628,18 @@ func (d *DatumRepository) getTimeRange(ctx context.Context, userId string, typ s
...
 		"_active": true,
 		"_userId": userId,
-		"type":    typ,
 		"time": bson.M{
 			"$gte": pastCutoff,
 			"$lte": futureCutoff,
 		},
 	}
 
+	if len(typ) > 1 {
```
toddkazakov, line 637: The special case here should be `len(typ) == 1`, because `$in` can handle empty arrays. Currently, there is a bug when `typ` is `nil` or `empty`. In addition to adding a guard for the length I think you should flip the logic.

data/store/mongo/mongo_datum.go
```diff
@@ -633,13 +589,18 @@ func (d *DatumRepository) GetDataRange(ctx context.Context, userId string, typ s
...
 		"_active": true,
 		"_userId": userId,
-		"type":    typ,
 		"time": bson.M{
 			"$gt":  status.FirstData,
 			"$lte": status.LastData,
 		},
 	}
 
+	if len(typ) > 1 {
```
toddkazakov, line 598: The special case here should be `len(typ) == 1`, because `$in` can handle empty arrays. You should flip the logic even if there is guard earlier in the function. This code was copy/pasted to other functions which do not check if `typ` is non-empty - this wouldn't have been a problem if special case here was handled correctly.

data/store/test/data_repository.go
```diff
@@ -126,6 +124,16 @@ type GetDataSetOutput struct {
 	Error   error
 }
 
+type IsDataSetAutomatedInput struct {
```
toddkazakov, line 127: All of the `IsDataSetAutomated` functionality can be removed, right?

task/summary/backfillrunner.go
```diff
@@ -181,5 +181,12 @@ func (t *BackfillTaskRunner) Run(ctx context.Context) error {
 	}
 	t.logger.Info(fmt.Sprintf("Backfilled %d BGM summaries", count))
 
+	t.logger.Debug("Starting User Continuous Summary Creation")
```
toddkazakov, line 184: This can be a loop over `[]string{"bgm", "cgm", "continuous"}`

toddkazakov (changes requested):
Looks much better! The last major remaining change would be to add a cursor which filters our non-continuous data. I left a suggestion how to do it with generics, but I didn't actually check if it compiles correctly.

task/summary/backfillrunner.go
Roukoswarf: makes sense, done.

data/store/test/data_repository.go
Roukoswarf: yes, good catch.

data/store/mongo/mongo_datum.go
Roukoswarf: replaced all cases of len < 1 with == 0 throughout the file.

data/store/mongo/mongo_datum.go
Roukoswarf: done

data/store/mongo/mongo_datum.go
Roukoswarf: done

data/store/mongo/mongo_datum.go
Roukoswarf: done

data/store/mongo/mongo_datum.go
Roukoswarf: done

toddkazakov (approved):

PR #702, opened 2024-03-13
changes alerts endpoints to match API docs

ewollesen:
This change was requested during review of the matching documentation update,

BACK-2500

### DIFF
--- a/data/service/api/v1/alerts.go
+++ b/data/service/api/v1/alerts.go
@@ -19,9 +19,9 @@ import (
 
 func AlertsRoutes() []service.Route {
 	return []service.Route{
-		service.Get("/v1/alerts/:userID/:followedUserID", GetAlert, api.RequireAuth),
-		service.Post("/v1/alerts/:userID/:followedUserID", UpsertAlert, api.RequireAuth),
-		service.Delete("/v1/alerts/:userID/:followedUserID", DeleteAlert, api.RequireAuth),
+		service.Get("/v1/users/:followedUserID/followers/:userID/alerts", GetAlert, api.RequireAuth),
+		service.Post("/v1/users/:followedUserID/followers/:userID/alerts", UpsertAlert, api.RequireAuth),
+		service.Delete("/v1/users/:followedUserID/followers/:userID/alerts", DeleteAlert, api.RequireAuth),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


data/service/api/v1/alerts.go
```diff
@@ -36,20 +38,19 @@ func DeleteAlert(dCtx service.Context) {
 		return
 	}
 
-	if err := checkUserIDConsistency(authDetails, r.PathParam("userID")); err != nil {
+	path := getUserIDsFromPath(r)
```
darinkrauss, line 41: [nit] Consider renaming `path` to `pathUserIDs`. I was confused by "getUserIDsFromPath" returning the "path".

data/service/api/v1/alerts.go
```diff
@@ -119,23 +120,21 @@ func UpsertAlert(dCtx service.Context) {
...
+	cfg := &alerts.Config{UserID: path.UserID, FollowedUserID: path.FollowedUserID, Alerts: *a}
 	if err := repo.Upsert(ctx, cfg); err != nil {
 		dCtx.RespondWithError(platform.ErrorInternalServerFailure())
 		lgr.WithError(err).Error("upserting alerts config")
 		return
 	}
 }
 
-var ErrUnauthorized = fmt.Errorf("unauthorized")
+var ErrUnauthorized = platformerrors.New("unauthorized")
```
darinkrauss, line 137: Please just create the error where it is needed. Using a "global" variable breaks the cause/caller "stack" for the error itself. (That is, it shows this location and not where the error actually originated.)

darinkrauss (changes requested):
I urge you to reconsider the use of "follower" versus "followed" within the same API. I understand your intention, but there is only a single character difference between the words (and very easy to miss) and it is just confusing switching between the two.

data/service/api/v1/alerts.go
ewollesen: Here I thought I was doing right by not using fmt... 

How do you recommend I alter tests to check for that error? I'll try to see how it's done in other places, but I've typically used package level errors, as they're what I can actually test for. It's what's done all over the Go stdlib. I understand we have a different errors package though, so continue to have patience with me. :)

darinkrauss: Its not ideal, but you can use `MatchError` for something like:
```
Expect(err).To(MatchError("unauthorized"))
```

ewollesen: Done

darinkrauss (approved):
LGTM!

PR #704, opened 2024-03-21
adds ability to send push notifications to Apple devices

ewollesen:
This is just a start, there's nothing using this yet, but it felt like a complete change, and one worth pushing on its own.

The intent is to allow sending push notifications, while providing minimal
friction to switching push notification providers. A future requirement to
send notifications to Android is expected, but in the meantime, the front-end
team has settled on APNs for their notifications. It's simple enough to
support multiple notification providers, and this interface can easily be
removed in the future if we decide on a single provider and don't feel that
the abstraction brings further value.
    
The change brings in a new dependency, github.com/sideshow/apns2, however it's
a small library, with very few dependencies of its own[1], primarily
those packages I would likely use in implementing the functionality myself.
    
[1]: Aside from the kingpin library, whose functionality we aren't using
anyway, as it's only used for a standalone binary that's part of the package,
it's not used by the library itself.
    
BACK-2507

### DIFF
--- a/alerts/send.go
+++ b/alerts/send.go
@@ -1,0 +1,94 @@
+package alerts
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/sideshow/apns2"
+	"github.com/sideshow/apns2/payload"
+	"github.com/sideshow/apns2/token"
+
+	"github.com/tidepool-org/platform/devicetokens"
+	"github.com/tidepool-org/platform/errors"
+)
+
+// Pusher abstracts the underlying push notification mechanism.
+type Pusher interface {
+	// Push a notification to a device.
+	//
+	// deviceToken should uniquely identify a device, while Notification
+	// abstracts the provider-specific notification payload.
+	Push(ctx context.Context, deviceToken devicetokens.DeviceToken, notification *Notification) error
+}
+
+// Notification models a provider-independent push notification.
+type Notification struct {
+	Title   string
+	Message string
+}
+
+// APNSPusher implements push notifications via Apple APNs.
+type APNSPusher struct {
+	serviceToken *token.Token
+	bundleID     string
+}
+
+// NewAPNSPusher creates a Pusher for sending device notifications via Apple's
+// APNs.
+//
+// The signingKey is the raw token signing key received from Apple (.p8 file
+// containing PEM-encoded private key), along with its respective team id, key
+// id, and application bundle id.
+//
+// https://developer.apple.com/documentation/usernotifications/sending-notification-requests-to-apns
+func NewAPNSPusher(signingKey []byte, keyID, teamID, bundleID string) (*APNSPusher, error) {
+	authKey, err := token.AuthKeyFromBytes(signingKey)
+	if err != nil {
+		return nil, err
+	}
+	token := &token.Token{
+		AuthKey: authKey,
+		KeyID:   keyID,
+		TeamID:  teamID,
+	}
+	return &APNSPusher{
+		bundleID:     bundleID,
+		serviceToken: token,
+	}, nil
+}
+
+func (p *APNSPusher) Push(ctx context.Context, deviceToken devicetokens.DeviceToken, note *Notification) error {
+	if deviceToken.Apple == nil {
+		return errors.New("Unable to push notification: APNSPusher can only use Apple device tokens but the Apple token is nil")
+	}
+
+	// TODO: look at the clientmanager package in apns2
+	client := apns2.NewTokenClient(p.serviceToken)
+	if deviceToken.Apple.Environment == "production" {
+		client = client.Production()
+	} else {
+		client = client.Development()
+	}
+
+	appleNote := p.buildAppleNotification(deviceToken.Apple, note)
+	resp, err := client.PushWithContext(ctx, appleNote)
+	if err != nil {
+		return errors.Wrap(err, "Unable to push notification")
+	}
+	if resp.StatusCode != http.StatusOK {
+		return errors.Newf("Unable to push notification: APNs returned non-200 status: %d, %s", resp.StatusCode, resp.Reason)
+	}
+
+	return nil
+}
+
+func (p *APNSPusher) buildAppleNotification(deviceToken *devicetokens.AppleDeviceToken, note *Notification) *apns2.Notification {
+	payload := payload.NewPayload().
+		Alert(note.Message).
+		AlertBody(note.Message)
+	return &apns2.Notification{
+		DeviceToken: string(deviceToken.Token),
+		Payload:     payload,
+		Topic:       p.bundleID,
+	}
+}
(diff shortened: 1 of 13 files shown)


alerts/send.go
```diff
@@ -0,0 +1,94 @@
...
+type Pusher interface {
+	// Push a notification to a device.
+	//
+	// deviceToken should uniquely identify a device, while Notification
+	// abstracts the provider-specific notification payload.
+	Push(ctx context.Context, deviceToken devicetokens.DeviceToken, notification *Notification) error
+}
+
+// Notification models a provider-independent push notification.
+type Notification struct {
```
darinkrauss, line 25: Do we have plans for supporting OS-specific push notification data? Are there product requirements regarding the different OS experiences? Have the different OS experiences been considered from a product perspective? Non-technical folks may not understand how push notifications differ between iOS and Android. For example, sounds, numbered badges, message threading, localization, visibility, priority, etc.

A good place to get some ideas is Firebase Cloud Messaging (FCM). FCM supports both iOS and Android push notifications (as well as web). Take a look at https://pkg.go.dev/firebase.google.com/go/messaging#Message. It has a few common fields, but also very specific iOS and Android fields.

I suppose we could bury this OS-specific configuration within the implementation and from the API/interface side only allow Title and Message. I suppose it depends upon whether this solution will be our general PN system or if it will be very specific to the Care Partner app alerts. If more general, then I suspect we'll need to expose more OS-specific options.

alerts/send.go
```diff
@@ -0,0 +1,94 @@
...
+	// TODO: look at the clientmanager package in apns2
+	client := apns2.NewTokenClient(p.serviceToken)
+	if deviceToken.Apple.Environment == "production" {
+		client = client.Production()
+	} else {
+		client = client.Development()
+	}
+
+	appleNote := p.buildAppleNotification(deviceToken.Apple, note)
+	resp, err := client.PushWithContext(ctx, appleNote)
```
darinkrauss, line 74: Retry if error?

darinkrauss: Or is that handled at a higher layer?

alerts/send.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	if resp.StatusCode != http.StatusOK {
+		return errors.Newf("Unable to push notification: APNs returned non-200 status: %d, %s", resp.StatusCode, resp.Reason)
+	}
+
+	return nil
+}
+
+func (p *APNSPusher) buildAppleNotification(deviceToken *devicetokens.AppleDeviceToken, note *Notification) *apns2.Notification {
+	payload := payload.NewPayload().
```
darinkrauss, line 86: If we are resending a push notification for a specific alert (that has not been acknowledged), it would be useful to thread the PN.

alerts/send.go
```diff
@@ -0,0 +1,94 @@
...
+	}, nil
+}
+
+func (p *APNSPusher) Push(ctx context.Context, deviceToken devicetokens.DeviceToken, note *Notification) error {
+	if deviceToken.Apple == nil {
+		return errors.New("Unable to push notification: APNSPusher can only use Apple device tokens but the Apple token is nil")
+	}
+
+	// TODO: look at the clientmanager package in apns2
+	client := apns2.NewTokenClient(p.serviceToken)
```
darinkrauss, line 66: We'll want to create this client once and keep it around. From the APNS2 documentation:

> As per the Apple APNs Provider API, you should keep a handle on this client so that you can keep your connections with APNs open across multiple notifications; don’t repeatedly open and close connections. APNs treats rapid connection and disconnection as a denial-of-service attack.

I think this implies a higher-level "manager" object that is created at the beginning of time for the service and maintains the client.

darinkrauss: And, to support multiple simultaneous outgoing PNs, you'll want to look into the APNS2 client manager (as your TODO comment states above).

alerts/send.go
darinkrauss: Also, we'll probably want to specify a CollapseID so that if multiple PNs are sent they are collapsed into one. (This is orthogonal to the ThreadID.) This can happen if there are network issues while sending to APNS (it is accepted, but we get an error/timeout).

alerts/send.go
```diff
@@ -0,0 +1,94 @@
...
+	appleNote := p.buildAppleNotification(deviceToken.Apple, note)
+	resp, err := client.PushWithContext(ctx, appleNote)
+	if err != nil {
+		return errors.Wrap(err, "Unable to push notification")
+	}
+	if resp.StatusCode != http.StatusOK {
+		return errors.Newf("Unable to push notification: APNs returned non-200 status: %d, %s", resp.StatusCode, resp.Reason)
+	}
+
+	return nil
```
darinkrauss, line 82: We may want to log the resulting `ApnsID` and general information about the push notification for tracking sake. We might even need to capture this as part of the push notification record so that we can confirm we at least got it into the APNS system in case of AER analysis.

darinkrauss (commented):
Some comments to consider.

alerts/send.go
ewollesen: The short answer is that the mobile team had already decided to use APNS, and are prepared with that workflow.

I've no doubt we'll be supporting FCM at some point, so I made sure to try not to paint myself into any corners. But in the meantime, the spec called for platform to accept APNS tokens and use them to send push notifications.

I read enough about how APNS and FCM both work to be confident that we can easily support both with a minimum of fuss. We'd need to put some thin abstraction layer in anyway, so that we can test the code.

alerts/send.go
ewollesen: Yep, a good call out. I'm aware. This was just the shortest/easiest path to something that worked. :)

I'll be sure to revise this before we're ready to get live.

alerts/send.go
ewollesen: Coming Soon (tm)

alerts/send.go
ewollesen: Sure

alerts/send.go
ewollesen: All good points, this is just minimum implementation right now.

alerts/send.go
ewollesen: As for generality vs specificity... I'm not aware of anything other than the care partner app that will be sending push notifications right now. Therefore my vote is to keep it as simple as possible until we have some idea of what more we'd actually need.

alerts/send.go
ewollesen: > I think this implies a higher-level "manager" object that is created at the beginning of time for the service and maintains the client.

The manager just tracks a list of clients, and can remove old ones (it's not clear why a client being old is a problem, they auto-reconnect). More importantly for us, it only works with certificate clients, while we're using token clients. In other words, the client manager won't work for our use case.

I don't think we need more than a single client for now, but if we decide we need multiple later, we can use the sync.Pool or something along those lines.

PR #707, opened 2024-04-12
don't look for device data provenance without a JWT

ewollesen:
If a request uses a service secret, then there can be no extraction of provenance information, as that's only carried in a JWT, so don't try.

While failed attempts weren't hurting anything, they were adding noise to the logs.


lostlevels (approved):

ewollesen:
https://tidepool.atlassian.net/browse/BACK-2995

data/service/api/v1/datasets_data_create.go
```diff
@@ -106,7 +105,9 @@ func DataSetsDataCreate(dataServiceContext dataService.Context) {
 	for _, datum := range datumArray {
 		datum.SetUserID(dataSet.UserID)
 		datum.SetDataSetID(dataSet.UploadID)
-		datum.SetProvenance(CollectProvenanceInfo(ctx, req, details))
+		if authDetails.HasToken() {
+			datum.SetProvenance(CollectProvenanceInfo(ctx, req, authDetails))
+		}
```
darinkrauss, line 110: Might I suggest pushing all logic of what has and what does not have provenance into the `CollectProvenanceInfo` function?

```suggestion
		if provenance := CollectProvenanceInfo(ctx, req, authDetails) {
			datum.SetProvenance(provenance)
		}
```

darinkrauss: Then move the check for `HasToken` into `CollectProvenanceInfo`.

data/service/api/v1/datasets_data_create.go
```diff
@@ -136,28 +137,26 @@ func DataSetsDataCreate(dataServiceContext dataService.Context) {
...
-		if token != "" {
-			if strings.EqualFold(token[:len("bearer ")], "bearer ") {
-				token = token[len("bearer "):]
-			}
+	token := authDetails.Token()
+	if token != "" && strings.Contains(strings.ToLower(token), "bearer") {
+		if strings.EqualFold(token[:len("bearer ")], "bearer ") {
+			token = token[len("bearer "):]
 		}
 	}
```
darinkrauss, line 149: Perhaps simpler?
```suggestion
    if strings.BeginsWith(strings.ToLower(token), "bearer ") {
		token = token[len("bearer "):]
	}
```

data/service/api/v1/datasets_data_create.go
```diff
@@ -136,28 +137,26 @@ func DataSetsDataCreate(dataServiceContext dataService.Context) {
 // CollectProvenanceInfo from a request and its auth details.
 //
 // All work is done as a best effort right now.
-func CollectProvenanceInfo(ctx context.Context, req *rest.Request, details request.AuthDetails) *data.Provenance {
+func CollectProvenanceInfo(ctx context.Context, req *rest.Request, authDetails request.AuthDetails) *data.Provenance {
 	lgr := log.LoggerFromContext(ctx)
```
darinkrauss, line 141: From above, you'd add:
```suggestion
    if !authDetails.HasToken() {
        return nil
    }

	lgr := log.LoggerFromContext(ctx)
```

data/service/api/v1/datasets_data_create_provenance_test.go
```diff
@@ -61,15 +61,16 @@ func newTestReqAndDetails(clientID, userID, sourceIP string) (*rest.Request, req
 	if sourceIP != "" {
 		remoteAddr = sourceIP + ":1234"
 	}
+	token := newTestToken(clientID)
```
darinkrauss, line 64: How about a test covering this change? That is, no provenance is collected if the auth details do not have a token?

darinkrauss (changes requested):
A few suggestions to consider, but I think you at least need a test covering the reason for this change (auth details without a token).

data/service/api/v1/datasets_data_create.go
ewollesen: > Might I suggest pushing all logic of what has and what does not have provenance into the `CollectProvenanceInfo` function?

Why? I don't understand how this is an improvement. You've not explained any reason for the change.

data/service/api/v1/datasets_data_create.go
ewollesen: Yes, thanks! I'll just double check the logic works out okay, then pull this in.

data/service/api/v1/datasets_data_create.go
darinkrauss: Sure, I suppose I should have explained in more detail. 

It appears, at least to me, that there is a prerequisite required *before* calling `CollectProvenanceInfo`, namely ensuring that `authDetails.HasToken` is true. So, the function should either document that prerequisite for future use (by someone other than yourself) or it should handle the prerequisite itself. The latter seems the easiest and fully encapsulates the concept of "get provenance info from auth details".

ewollesen: Thanks for clarifying. In going over your suggestion, I realized as well something that you wrote in your explanation:

> that there is a prerequisite required before calling CollectProvenanceInfo,

That was not the case previously. It wasn't my intention to make it a prerequisite, but you're right that my first edition of this PR did have that effect. I don't think that's desirable. I think it's better to collect whatever provenance information we can, from whatever's available. That's why I have tests to handle the case where any of the various values in the auth details are empty. The goal is to collect whatever information is there, and not abort the first time something that was expected is missing (though we should log that). Thanks for pointing that out. I'm going to change up my approach as a result.

Restating: The objective is to be callable with or without a token, but to not be a noisy logger unless the token isn't present *when it's expected to be present*. More concretely, if the request is not authenticated with a service secret, then we expect there to be a token. If there's not, that's something we should log. We shouldn't just return nil however. Doing so would break a bunch of tests, and just doesn't do any good.

data/service/api/v1/datasets_data_create_provenance_test.go
ewollesen: Added.

data/service/api/v1/datasets_data_create.go
ewollesen: Yes, thanks.

darinkrauss (approved):

ewollesen:
https://tidepool.atlassian.net/browse/BACK-2995  (added after PR created, but mentioned in commit message)

PR #713, opened 2024-04-25
[BACK-2963] Standard Deviation and other new CGM stats

Roukoswarf:


data/summary/types/cgm.go
```diff
@@ -150,6 +159,12 @@ type CGMPeriod struct {
 	HasTimeInAnyHighRecords   bool `json:"hasTimeInAnyHighRecords" bson:"hasTimeInAnyHighRecords"`
 	TimeInAnyHighRecords      *int `json:"timeInAnyHighRecords" bson:"timeInAnyHighRecords"`
 	TimeInAnyHighRecordsDelta *int `json:"timeInAnyHighRecordsDelta" bson:"timeInAnyHighRecordsDelta"`
+
+	SD float64 `json:"SD" bson:"SD"`
```
toddkazakov, line 163: Consider not using an acronyms here

data/summary/types/cgm.go
```diff
@@ -36,6 +37,14 @@ type CGMBucketData struct {
 	TotalGlucose float64 `json:"totalGlucose" bson:"totalGlucose"`
 	TotalMinutes int     `json:"totalMinutes" bson:"totalMinutes"`
 	TotalRecords int     `json:"totalRecords" bson:"totalRecords"`
+
+	TotalVariance float64 `json:"totalVariance" bson:"totalVariance"`
+}
+
+type CGMTotalStats struct {
```
ewollesen, line 44: note to self: what is this for?

data/summary/types/cgm.go
```diff
@@ -231,6 +246,26 @@ func (s *CGMStats) Update(ctx context.Context, cursor fetcher.DeviceDataCursor)
 	return nil
 }
 
+func (B *CGMBucketData) CalculateVariance(value float64, duration float64) float64 {
+	var mean float64 = 0
+	if B.TotalMinutes > 0 {
+		mean = B.TotalGlucose / float64(B.TotalMinutes)
+	}
+
+	weight := float64(B.TotalMinutes) + duration
```
ewollesen, line 255: This appears to be using the "weighted incremental algorithm" as detailed on Wikipedia at https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Weighted_incremental_algorithm

If so, consider calling that out in a comment. It helps reviewers double check that you've nailed the algorithm if we know what you were aiming for.

data/summary/types/cgm.go
```diff
@@ -231,6 +246,26 @@ func (s *CGMStats) Update(ctx context.Context, cursor fetcher.DeviceDataCursor)
...
+	newMean := mean + (duration/weight)*(value-mean)
+	return B.TotalVariance + duration*(value-mean)*(value-newMean)
+}
+
+func (B *CGMBucketData) CombineVariance(newBucket *CGMBucketData) float64 {
```
ewollesen, line 260: A note that this uses https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm

would be helpful for future reviewers, when verifying.

data/summary/types/cgm.go
```diff
@@ -283,19 +321,33 @@ func (s *CGMStats) CalculateSummary() {
...
+	totalStats := &CGMTotalStats{}
+	totalOffsetStats := &CGMTotalStats{}
+	dayCounted := false
+	offsetDayCounted := false
 
 	for i := 0; i < len(s.Buckets); i++ {
 		currentIndex := len(s.Buckets) - 1 - i
 
+		// new day, reset day counting flag
+		if i%24 == 0 {
```
ewollesen, line 333: 24 is a magic number here.

While it's pretty obvious that it's the number of hours in a day, what's less obvious is that it's being used as a duration of each element in a bucket. 

That last bit is the important one, in the (unlikely) event that the duration should change, but even without that change, it helps any reader understand why 24 is appearing in the code.

ewollesen, line 333: I'm not as up-to-date with the Buckets as I probably should be, but this code would seem to assume there are no "holes" in the data. That is, that each element of a Bucket covers an hour, and that there not gaps or blanks... There's an element per hour for the total span of the buckets. 

Is that true to the design? I poked a little, but didn't find where blanks or empty values were inserted.

data/summary/types/cgm_test.go
```diff
@@ -25,6 +25,8 @@ import (
 
 	"github.com/tidepool-org/platform/data/summary/types"
 
+	"gonum.org/v1/gonum/stat"
```
ewollesen, line 28: nit: the order of these imports could be cleaned up

data/summary/types/cgm_test.go
```diff
@@ -94,6 +96,36 @@ func NewDataSetCGMDataRanges(startTime time.Time, hours float64, ranges DataRang
...
+func NewDataSetCGMVariance(startTime time.Time, hours int, perHour int, SD float64) ([]data.Datum, float64) {
+	requiredRecords := hours * perHour
+	typ := pointer.FromString("cbg")
+
+	var dataSetData = make([]data.Datum, requiredRecords)
+	var uploadId = test.RandomSetID()
+	var deviceId = "SummaryTestDevice"
+
+	// dedicated rand, to be sure the sequence is kept just in case.
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
ewollesen, line 108: nit: Recording/printing this seed in the event of a failure could be useful... But I guess it could always be done later too, when there's an error that's found.

data/summary/types/cgm_test.go
```diff
@@ -94,6 +96,36 @@ func NewDataSetCGMDataRanges(startTime time.Time, hours float64, ranges DataRang
...
+
+	var values []float64
+
+	// generate requiredRecords of data
+	for count := 0; count < requiredRecords; count += perHour {
+		for inHour := 0; inHour < perHour; inHour++ {
+			datumTime := startTime.Add(time.Duration(-(count + inHour + 1)) * time.Hour / time.Duration(perHour))
+
+			datum := NewGlucose(typ, pointer.FromString(units), &datumTime, &deviceId, &uploadId)
```
ewollesen, line 117: super nit: you could pull this out of the loops for a small speed bump. :)

ewollesen (changes requested):
Mostly nits. I do think that the SD/CV should be expanded though.

data/summary/types/cgm.go
```diff
@@ -231,6 +246,26 @@ func (s *CGMStats) Update(ctx context.Context, cursor fetcher.DeviceDataCursor)
 	return nil
 }
 
+func (B *CGMBucketData) CalculateVariance(value float64, duration float64) float64 {
```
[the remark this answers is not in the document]
Roukoswarf: No particular reason, it stood out better, Buckets starts with a capital B, I've started using multi-letter receivers in other locations, for example:
```
func (gs *GlucoseSummarizer[A, T])
```
Which I feel is nicer.

`B` from the revision history is quite old, part of the original generics refactor, I just went through renaming things from having all the same, indistinguishable `s` for `stats` which was very very confusing.

data/summary/types/cgm.go
Roukoswarf: no holes, but holes are coming soon with the buckets rework im working on, so much has to change to handle both the buckets moving to their own collection, having variable duration (compressing older data points), and having holes matching where there is no data, it's probably not worth putting too much thought into this section, as the entire bucket handling is being rewritten.

data/summary/types/cgm_test.go
Roukoswarf: I was expecting when this function was originally coded that it would generate much closer to requested StDev, but it results in quite far off requested values even with its own rand, it would require a completely different approach to generate a specific deviation, so this should just use the default rand, which is already printed at the start of the test.

I've removed this line.

data/summary/types/cgm_test.go
Roukoswarf: fixed across all summary files.

data/summary/types/cgm.go
Roukoswarf: because this will soon change in many ways, I am voting to keep the magic number as it matches the implementation (no holes, always 24 hourly buckets, no variable duration), and will very soon be replaced by a more dynamic implementation which will outsource much of the time processing to db queries, removing the need for hourly offsets entirely.

data/summary/types/cgm.go
Roukoswarf: Wikipedia can change at any time, archive.org link? copy of the text block inside the comment? How would you preserve this in the comment?

data/summary/types/cgm_test.go
Roukoswarf: Maybe its unclear as github highlights a section and im reading this wrong, but how would we avoid NewGlucose in the loop? Reuse with a copy might be faster, but also a bit less clear, the unit test already runs near instantly.

Roukoswarf: I did move all the repeated pointer functions out of the loop, perhaps thats what was meant.

data/summary/types/cgm.go
ewollesen: Anything on the web can change at any time (think about archive.org's recentl legal cases and what impact they might have). The point is just to try and be helpful. It's much easier to verify this code does what you think it does when you know exactly what it's trying to do. :)

IMO linking to wikipedia is fine for this. Wikipedia has a history function, and things like this aren't too likely to change. It's not a hotly contested area of current research.

Wikipedia references this article: https://dl.acm.org/doi/10.1145/359146.359153 you could link that instead/too. It's from CACM which recently became free to anyone.

data/summary/types/cgm.go
ewollesen: See previous response

data/summary/types/cgm.go
ewollesen: I'll trust your judgment then. :)

ewollesen: Thank you for the clarification.

data/summary/types/cgm_test.go
ewollesen: Oh yeah, sorry, it was just the `pointer.FromString(units)` that I was trying to highlight. I probably fat-fingered the selection when I was commenting and didn't realize.

data/summary/types/bgm_test.go
```diff
@@ -49,8 +44,8 @@ func NewDataSetBGMDataAvg(deviceId string, startTime time.Time, hours float64, r
 		for i, glucoseValue := range glucoseValues {
 			datumTime := startTime.Add(time.Duration(-(count + i + 1)) * time.Minute * 10)
 
-			datum := NewGlucose(typ, pointer.FromString(units), &datumTime, &deviceId, &uploadId)
-			datum.Value = pointer.FromFloat64(glucoseValue)
+			datum := NewGlucose(typ, unit, &datumTime, deviceIdPtr, uploadId)
```
ewollesen, line 47:
```suggestion
			datum := NewGlucose(typ, &units, &datumTime, &deviceId, uploadId)
```
Heh, I know you just changed this, and it's fine, but I see a bit better what's happening here. Could you just use the `&` operator here and avoid `pointer.From*` entirely for units and deviceId? Then you could remove the variable assignments above on lines 35 & 36?

Doesn't matter, take it or leave it.

data/summary/types/cgm_test.go
```diff
@@ -86,8 +81,8 @@ func NewDataSetCGMDataRanges(startTime time.Time, hours float64, ranges DataRang
 		for i, bracket := range glucoseBrackets {
 			datumTime := startTime.Add(time.Duration(-(count+i+1))*time.Minute*5 - gapCompensation)
 
-			datum := NewGlucose(typ, pointer.FromString(units), &datumTime, &deviceId, &uploadId)
-			datum.Value = pointer.FromFloat64(bracket[0] + (bracket[1]-bracket[0])*rand.Float64())
+			datum := NewGlucose(typ, unit, &datumTime, &deviceId, &uploadId)
+			datum.Value = pointer.FromAny(bracket[0] + (bracket[1]-bracket[0])*rand.Float64())
```
ewollesen, line 85: This is pretty clearly a float64, no? Why not use pointer.FromFloat64? I like types, they help keep my code safe and alert me to problems early on.

data/summary/types/cgm.go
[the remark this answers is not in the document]
ewollesen: OK then. I'll just have to get used to it I guess. :)

I've no problem with two-letter receivers either, or even words, when they make sense. At the end of the day, my brain will adjust.

data/summary/types/bgm_test.go
Roukoswarf: I converted units from a const to a var, so it can be addressed, and removed all other unneeded pointer.FromAny calls in the data generation.

data/summary/types/cgm_test.go
Roukoswarf: Types are still validated, this is a generic function and does not return a typeless interface, closing as discussed.

data/summary/types/cgm.go
Roukoswarf: Linked to the relevant wikipedia section in a comment.

data/summary/types/cgm.go
Roukoswarf: Linked to the relevant wikipedia section in a comment.

data/summary/types/cgm.go
Roukoswarf: Expanded variable names to be full names `StandardDeviation` and `CoefficientOfVariation`

data/summary/types/cgm.go
```diff
@@ -244,6 +244,7 @@ func (s *CGMStats) Update(ctx context.Context, cursor fetcher.DeviceDataCursor)
 	return nil
 }
 
+// CalculateVariance Implemented using https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Weighted_incremental_algorithm
```
ewollesen, line 247: TY

data/summary/types/cgm_test.go
```diff
@@ -1488,10 +1482,14 @@ var _ = Describe("CGM Summary", func() {
 
 				for _, period := range periodKeys {
 					By(fmt.Sprintf("checking period %s", period))
-					Expect(userCGMSummary.Stats.Periods[period].StandardDeviation).To(BeNumerically("~", targetSD, 1.5))
+					// the range for this value is quite large, as the result is still RNG, we check the precise result in the final period
+					Expect(userCGMSummary.Stats.Periods[period].StandardDeviation).To(BeNumerically("~", targetSD, 4))
```
ewollesen, line 1486: We don't have anything that uses these intermediate values, do we?

Since they vary so much (which isn't that surprising), and if we don't use the intermediate values, it's probably ok not to test them.

ewollesen (approved):
Great work Alex!

data/summary/types/cgm_test.go
Roukoswarf: I'm mostly just checking that they are populated with something believable, but yes, since the generation hits 30d at once, 1d especially is likely to be very far off of the 30d value due to the sample size and rng.

I might make the generation return all 4 expected values, if i can find an elegant way to do calculate that without too much work, or ill just check that the values exist instead, as adding rng to tests passing is not funny for the guy who has to hit re-run 3 times on bad luck.

data/summary/types/bgm_test.go
```diff
@@ -322,53 +316,69 @@ var _ = Describe("BGM Summary", func() {
...
 				for i := len(userBGMSummary.Stats.Buckets) - 1; i >= 0; i-- {
 					f := fmt.Sprintf("hour %d", i+1)
 					By(f)
-					Expect(userBGMSummary.Stats.Buckets[i].Data.TotalRecords).To(Equal(5))
+					Expect(userBGMSummary.Stats.Buckets[i].Data.TotalRecords).To(Equal(6))
 
-					lastRecordTime = datumTime.Add(-time.Hour*time.Duration(len(userBGMSummary.Stats.Buckets)-i-1) - time.Minute*12)
-					Expect(userBGMSummary.Stats.Buckets[i].LastRecordTime).To(Equal(lastRecordTime))
+					//lastRecordTime = datumTime.Add(-time.Hour*time.Duration(len(userBGMSummary.Stats.Buckets)-i-1) - time.Minute*10)
+					// TODO
```
ewollesen, line 331: Is this still to be done?

If not please remove, if so, please... do?

If you end up back in here, consider the `var lastRecordTime` on line 292 as well.

ewollesen (changes requested):
Just one question around a TODO comment and some commented out code.

data/summary/types/bgm_test.go
Roukoswarf: Uncommented, was a fully working check I only temporarily needed to bypass, good catch.

ewollesen (approved):

PR #714, opened 2024-04-25
adds a plain-text logger for development or testing

ewollesen:
JSON logging is great for production, but when running things locally, it's very helpful to have something more readable.

This devlog injects itself when the POD_NAMESPACE env var is "default" (used by the development repo's local k8s cluster) or when the TIDEPOOL_USE_DEVLOG env var is set (to any non-blank value).

It has a few features:

  1. shortens timestamp
  2. leaves out process information
  3. shows basic caller information only for error messages (filename:line)
  4. abbreviates log level output ("error" => "EE", "debug" => "DD", etc)
  5. handles other fields by appending them to the main log message

### DIFF
--- a/application/provider.go
+++ b/application/provider.go
@@ -13,6 +13,7 @@ import (
 	configEnv "github.com/tidepool-org/platform/config/env"
 	"github.com/tidepool-org/platform/errors"
 	"github.com/tidepool-org/platform/log"
+	"github.com/tidepool-org/platform/log/devlog"
 	logJson "github.com/tidepool-org/platform/log/json"
 	"github.com/tidepool-org/platform/sync"
 	"github.com/tidepool-org/platform/version"
@@ -101,7 +102,16 @@ func NewProvider(prefix string, scopes ...string) (*ProviderImpl, error) {
 
 	level := configReporter.WithScopes("logger").GetWithDefault("level", "warn")
 
-	logger, err := logJson.NewLogger(writer, log.DefaultLevelRanks(), log.Level(level))
+	var logger log.Logger
+	var loggerPackage = configReporter.WithScopes("logger").GetWithDefault("package", "json")
+	switch loggerPackage {
+	case "json":
+		logger, err = logJson.NewLogger(writer, log.DefaultLevelRanks(), log.Level(level))
+	case "devlog":
+		logger, err = devlog.New(writer, log.DefaultLevelRanks(), log.Level(level))
+	default:
+		err = errors.Newf("unrecognized logger package: %s", loggerPackage)
+	}
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to create logger")
 	}
--- a/log/devlog/devlog.go
+++ b/log/devlog/devlog.go
@@ -1,0 +1,118 @@
+// Package devlog provides a plain-text logger suitable for reading from the
+// console.
+//
+// For production, the team prefers JSON logging, but plain-text can be useful
+// for local development and tests.
+//
+// So while this package shouldn't be used for production, it can be enabled
+// for your development use by setting TIDEPOOL_LOGGER_PACKGE=devlog as an
+// environment variable. If you're using the development repo's helm chart,
+// there's a global.loggerPackage override available to do the same.
+
+package devlog
+
+import (
+	"fmt"
+	"io"
+	stdlog "log"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/tidepool-org/platform/errors"
+	"github.com/tidepool-org/platform/log"
+)
+
+// New provides a plain-text log.Logger suitable for development or testing.
+func New(writer io.Writer, levelRanks log.LevelRanks, level log.Level) (log.Logger, error) {
+	if writer == nil {
+		return nil, errors.New("writer is missing")
+	}
+	logLogger := stdlog.New(writer, "", 0)
+	return log.NewLogger(&serializer{logLogger}, levelRanks, level)
+}
+
+func NewWithDefaults(writer io.Writer) (log.Logger, error) {
+	return New(writer, log.DefaultLevelRanks(), log.DebugLevel)
+}
+
+type serializer struct {
+	*stdlog.Logger
+}
+
+func (s *serializer) Serialize(fields log.Fields) error {
+	var pairs = make([]string, 0, len(fields))
+	var msg, msgTime, msgLevel, msgCaller string
+	var showCaller = false
+
+	msgLevel = abbreviateLevel("")
+	msgTime = time.Now().Format(time.Stamp)
+
+	for key, value := range fields {
+		switch key {
+		case "caller":
+			msgCaller = formatCaller(value)
+		case "level":
+			level := sOrPlusV(value)
+			msgLevel = abbreviateLevel(level)
+			if level == string(log.ErrorLevel) {
+				showCaller = true
+			}
+		case "message":
+			msg = sOrPlusV(value)
+		case "process":
+			// process isn't useful outside of sumo
+		case "time":
+			// time is terser and constant width for console logging
+			tv, err := time.Parse(time.RFC3339Nano, sOrPlusV(value))
+			if err != nil {
+				msgTime = sOrPlusV(value)
+				continue
+			}
+			msgTime = tv.Format(time.Stamp)
+		default:
+			pairs = append(pairs, key+"="+sOrPlusV(value))
+		}
+	}
+	if showCaller {
+		pairs = append(pairs, msgCaller)
+	}
+	sort.Strings(pairs)
+	rest := ""
+	if len(pairs) > 0 {
+		rest = ": " + strings.Join(pairs, " ")
+	}
+	s.Logger.Printf(msgTime + " " + msgLevel + " " + msg + rest)
+	return nil
+}
+
+func formatCaller(value any) string {
+	cc, ok := value.(*errors.Caller)
+	if !ok {
+		return "caller=" + sOrPlusV(value)
+	}
+	return fmt.Sprintf("caller=%s:%d", cc.File, cc.Line)
+}
+
+func sOrPlusV(thing any) (s string) {
+	if stringer, ok := thing.(fmt.Stringer); ok {
+		return stringer.String()
+	} else if s, ok := thing.(string); ok {
+		return s
+	}
+	return fmt.Sprintf("%+v", thing)
+}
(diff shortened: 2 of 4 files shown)


ewollesen:
Here's a quick / dumb example of its output:
```
Apr 25 17:09:28 EE testing the error level: caller=application/provider.go:115 some field=junk
Apr 25 17:09:28 II Logger level is debug
Apr 25 17:09:28 DD Initializing secret
Apr 25 17:09:28 DD Loading auth client config
Apr 25 17:09:28 DD Creating auth client
Apr 25 17:09:28 DD Starting auth client
```

application/provider.go
```diff
@@ -105,6 +106,12 @@ func NewProvider(prefix string, scopes ...string) (*ProviderImpl, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to create logger")
 	}
+	if os.Getenv("POD_NAMESPACE") == "default" || os.Getenv("TIDEPOOL_USE_DEVLOG") != "" {
```
darinkrauss, line 109: A few things here:
- I would highly prefer we not force this on all local developers (namely me). Either remove this code from the PR (and keep in a local stash for your use) or just use an environment variable to enable/disable. That is, remove the `POD_NAMESPACE` check at a minimum.
- If you're going to use an environment variable, consider using the `configReporter` to fetch it. That is the generally the standard for configuration in platform. Yea, I know, consistency.
- If you're going to use an environment variable, please use if/else to create one or the other logger. That is, don't create the JSON logger and then replace it with the plain text logger. Just create one or the other.

log/devlog/devlog.go
```diff
@@ -0,0 +1,127 @@
+// Package devlog provides a plain-text logger suitable for reading from the
```
darinkrauss, line 1: I would call this "plaintext" or "text" or "plain" (not "devlog"). That is, the package name and file names.

log/devlog/devlog.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"fmt"
+	"io"
+	"log"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/tidepool-org/platform/errors"
+	plog "github.com/tidepool-org/platform/log"
```
darinkrauss, line 23: `plog` to `platformLog`? Or, as we've done in other code, use `stdLog` for the standard library log. (Similar to `stdError`.) 

IMHO, except in certain, very common cases (i.e. "ID"), abbreviations tend to lead to confusion. What makes sense to one person as an abbreviation may mean something else to another.

log/devlog/devlog.go
```diff
@@ -0,0 +1,127 @@
...
+func formatCaller(value any) string {
+	cc, ok := value.(*errors.Caller)
+	if !ok {
+		return "caller=" + sOrPlusV(value)
+	}
+	return fmt.Sprintf("caller=%s:%d", cc.File, cc.Line)
+}
+
+func sOrPlusV(thing any) (s string) {
+	//defer func() { s = strings.TrimSpace(s) }()
```
darinkrauss, line 103: Remove?

log/devlog/devlog.go
```diff
@@ -0,0 +1,127 @@
...
+	for key, value := range fields {
+
+		switch key {
+		case "caller":
+			msgCaller = formatCaller(value)
+			continue
+		case "level":
+			level := sOrPlusV(value)
+			msgLevel = abbreviateLevel(level)
+			if level == "error" {
```
darinkrauss, line 60: Consider using `platformLog.ErrorLevel` constant.

log/devlog/devlog.go
```diff
@@ -0,0 +1,127 @@
...
+			continue
+		case "time":
+			// time is terser and constant width for console logging
+			tv, err := time.Parse(time.RFC3339Nano, sOrPlusV(value))
+			if err != nil {
+				msgTime = sOrPlusV(value)
+				continue
+			}
+			msgTime = tv.Format(time.Stamp)
+			continue
```
darinkrauss, line 78: Rather than all of the `continue` statements in each `case`, why not just put the `pairs = append(pairs, key+"="+sOrPlusV(value))` in a `default` (and then remove the unnecessary `continue`).

log/devlog/devlog.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func abbreviateLevel(fullLevel string) string {
+	if a, ok := abbr[fullLevel]; ok {
+		return a
+	}
+	if fullLevel == "" {
+		return "??"
+	}
+	return strings.ToUpper(fullLevel[0:2])
```
darinkrauss, line 126: I'd just use "??" if not one of the known levels. (There is the edge case where the fullLevel only has one character.)

darinkrauss (changes requested):
Some changes needed and some to consider.

log/devlog/devlog.go
ewollesen: slog and log already both provide plain text loggers. This is a specific customization (that happens to use plain text instead of JSON), so I think the existing name is better.

darinkrauss: Sure, I'll agree to disagree.

application/provider.go
ewollesen: Doh, sorry, it was my intention to make this opt-in, not opt-out. I was exploring ways of getting it enabled and pushed this in a bit of a rush (I was hacking on it during the all-hands :D)

I'll definitely get it to be opt-in only... I'll have to explore more options.

The TIDEPOOL_USE_DEVLOG is great, but is more intended for folks like @Roukoswarf that run platform directly. Those of us using development/k8s can't easily inject env vars into the development environment (that I'm aware of).

That said, I don't want to add an env var to the helm chart that's only use is for local development... that doesn't feel great. 

So I think my next step is to poke around in development and see if there's some kind of easy to inject local stuff easily... Opt-in of course.

ewollesen: If you know of a way to do that (what I mentioned above about adding local config bits to development), I'm all ears.

darinkrauss: Won't something like `kubectl set env deployment/data TIDEPOOL_USE_DEVLOG=true` work?

ewollesen: It does... though I was hoping to be able to set something once and have it be "always on".

If I add a setting to the helm chart, I can achieve that. What would you think about adding this to the helm chart?:
```
 - name: TIDEPOOL_LOGGER_PACKAGE
   value: {{ .Values.global.loggerPackage | default "" }}
```
This would allow us to set a value in development/local/Tiltconfig.yaml:
```
global:
  loggerPackage: "devlog"
```
If the value of TIDEPOOL_LOGGER_PACKAGE is anything other than "devlog", it would use the default (json) logger.

100% opt-in. Safe from accidentally being set in non-local dev environments, and still allows those that want to (myself among them) to "set it once and forget it".

log/devlog/devlog.go
ewollesen: Sure, doesn't matter much anyway, and not sure how it'd come about anyway.

Good catch on the edge case.

application/provider.go
ewollesen: I've put up a PR: https://github.com/tidepool-org/development/pull/282

darinkrauss: Okay, I can live with that.

application/provider.go
```diff
@@ -105,6 +106,12 @@ func NewProvider(prefix string, scopes ...string) (*ProviderImpl, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to create logger")
 	}
+	if os.Getenv("TIDEPOOL_LOGGER_PACKAGE") == "devlog" {
```
darinkrauss, line 109: Please simplify to something like:
```
    var logger log.Logger
    if os.Getenv("TIDEPOOL_LOGGER_PACKAGE") == "devlog" {
        logger, err = devlog.New(writer, log.DefaultLevelRanks(), log.Level(level))
    } else {
        logger, err = logJson.NewLogger(writer, log.DefaultLevelRanks(), log.Level(level))
    }
    if err != nil {
	    return nil, errors.Wrap(err, "unable to create logger")
    }
```
so it only creates one logger.

darinkrauss (dismissed):
Approving, but I'd still like changes for the two remaining unresolved comments.

application/provider.go
ewollesen: I've switched to a switch (alliteration!) I wrote it the way I originally did to try to convey that the jsonLog was the default, and devlog was an auxiliary choice. Your suggestion (a very natural way to write the code) didn't capture that intention well. Which is why I didn't just do that in the first place.

With a switch it's clear that the default is json, so I'm happy—hopefully it's acceptable to you as well.

log/devlog/devlog.go
```diff
@@ -0,0 +1,118 @@
+// Package devlog provides a plain-text logger suitable for reading from the
+// console.
+//
+// For production, the team prefers JSON logging, but plain-text can be useful
+// for local development and tests.
+//
+// So while this package shouldn't be used for production, it can be enabled
+// for your development use by setting TIDEPOOL_USE_DEVLOG=1 as an environment
+// variable. It will also be enabled if POD_NAMESPACE=default (this is set via
+// the development repo when using it's kind k8s cluster).
```
darinkrauss, line 10: Update the comment to match the latest `TIDEPOOL_LOGGER_PACKAGE` environment variable?

darinkrauss (changes requested):
A couple of comments. The only one needing change is updating the comment in `devlog.go`. The other, about the explicit `case "json"`, is just a suggestion.

application/provider.go
```diff
@@ -101,7 +102,13 @@ func NewProvider(prefix string, scopes ...string) (*ProviderImpl, error) {
 
 	level := configReporter.WithScopes("logger").GetWithDefault("level", "warn")
 
-	logger, err := logJson.NewLogger(writer, log.DefaultLevelRanks(), log.Level(level))
+	var logger log.Logger
+	switch configReporter.WithScopes("logger").GetWithDefault("package", "") {
```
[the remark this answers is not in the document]
ewollesen: That feels a little heavy-handed for something that doesn't affect the proper functioning of the program, don't you think? I think as a user I'd prefer it just used a sane default in the face of bad input, at least for this input. But that said, I'll make the change.

darinkrauss (approved):
LGTM!

ewollesen:
BACK-3014 created after the fact

PR #715, opened 2024-05-06
adds List and Get methods to alerts client

ewollesen:
The Get endpoint already exists on the service, so only the List endpoint needed to be added there.

BACK-2554


ewollesen:
To use this in QA, it must be paired with https://github.com/tidepool-org/hydrophone/pull/145

data/service/service/standard.go
```diff
@@ -419,3 +439,96 @@ func (s *Standard) initializeUserEventsHandler() error {
...
+	// lines everything up as expected.
+	topicPrefix := strings.ReplaceAll(commonConfig.KafkaTopicPrefix, "-", ".")
+	prefixedTopics := make([]string, 0, len(config.KafkaAlertsTopics))
+	for _, topic := range config.KafkaAlertsTopics {
+		prefixedTopics = append(prefixedTopics, topicPrefix+topic)
+	}
+
+	alerts := s.dataStore.NewAlertsRepository()
+	dataRepo := s.dataStore.NewDataRepository()
+	s.Logger().WithField("permissionClient", s.permissionClient).Debug("yo!")
```
toddkazakov, line 505: yo,yo,yo!

toddkazakov (commented):
Looks good overall, but the retry mechanism which is implemented here doesn't satisfy the latency requirements. The current implementation is ok for internal usage, but it's not production ready. This could be handled in a separate PR if this makes the development and QA process easier.

data/events/alerts.go
```diff
@@ -0,0 +1,354 @@
...
+
+	// Notes could be pushed into a Kafka topic to have a more durable retry,
+	// but that can be added later.
+	for _, note := range notes {
+		lgr := lgr.WithField("recipientUserID", note.RecipientUserID)
+		tokens, err := c.DeviceTokens.GetDeviceTokens(ctx, note.RecipientUserID)
+		if err != nil {
+			lgr.WithError(err).Info("Unable to retrieve device tokens")
+		}
+		if len(tokens) == 0 {
```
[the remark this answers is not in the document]
ewollesen: I think there is a need. Each note can have a different recipient. Each recipient has their own device tokens.

This is intended to handle the case where multiple care partners need to be sent the same notification. We generate one note per recipient. Each recipient can likewise have multiple tokens.

Alternatively, GetDeviceTokens could be optimized to retrieve the device tokens for multiple users, but that feels strange to me. I have a gut feeling that it wouldn't make the code any easier to understand, but I could rework in that direction.

data/events/alerts.go
```diff
@@ -0,0 +1,354 @@
...
+	if err := unmarshalMessageValue(msg.Value, cfg); err != nil {
+		return err
+	}
+	lgr := c.logger(ctx)
+	lgr.WithField("cfg", cfg).Info("consuming an alerts config message")
+
+	ctxLog := c.logger(ctx).WithField("followedUserID", cfg.FollowedUserID)
+	ctx = log.NewContextWithLogger(ctx, ctxLog)
+
+	notes, err := c.Evaluator.Evaluate(ctx, cfg.FollowedUserID)
```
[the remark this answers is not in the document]
ewollesen: In that case, there's a repetition going on, because the package's name is "alerts", so you end up with "alerts.AlertNotification". Would you be ok with "Notification" for the type, and a change for variable names from "note" => "notification"?

If you're editing a file within the alerts package, it should be clear that a Notification is a notification of an alert, no?

data/events/events.go
```diff
@@ -58,3 +65,135 @@ func (u *userDeletionEventsHandler) HandleDeleteUserEvent(payload ev.DeleteUserE
...
+}
+
+func (r *SaramaRunner) Initialize() error {
+	group, err := sarama.NewConsumerGroup(r.Config.Brokers, r.Config.GroupID, r.Config.Sarama)
+	if err != nil {
+		return errors.Wrap(err, "Unable to build sarama consumer group")
+	}
+	handler := asyncevents.NewSaramaConsumerGroupHandler(&asyncevents.NTimesRetryingConsumer{
+		Consumer: r.Config.MessageConsumer,
+		Delay:    CappedExponentialBinaryDelay(AlertsEventRetryDelayMaximum),
```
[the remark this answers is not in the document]
ewollesen: This will require some more in-depth thought on my part... Will do.

data/events/alerts.go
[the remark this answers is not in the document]
toddkazakov: Oh, yes, sorry. Ignore my initial comment.

data/events/alerts.go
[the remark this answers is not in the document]
toddkazakov: > Would you be ok with "Notification" for the type, and a change for variable names from "note" => "notification"?

Sure, that sounds reasonable

auth/store/test/device_token_repository.go
```diff
@@ -21,6 +23,16 @@ func (r *DeviceTokenRepository) Expectations() {
 	r.DeviceTokenAccessor.Expectations()
 }
 
+func (r *DeviceTokenRepository) GetAllByUserID(ctx context.Context, userID string) ([]*devicetokens.Document, error) {
```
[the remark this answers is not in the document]
ewollesen: Fixed.

data/events/events.go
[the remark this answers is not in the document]
ewollesen: Yeah, I think you're right, let's get this review merged, and I'll work on getting a multiple topic solution set up. Given the flexibility we have now, it shouldn't be too bad.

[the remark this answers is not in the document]
ewollesen: I have the multi-tier retry in the eric-alerts-multi-topic-retry branch.

toddkazakov (dismissed):

toddkazakov (dismissed):

ewollesen:
@toddkazakov I just removed two config env vars. I believe we talked about that before, but it slipped my mind until I was reviewing the helm chart changes today, where they came up again.

So the re-review here is just around the config parsing, in the most recent commit of the PR, nothing else is changed.

data/events/events.go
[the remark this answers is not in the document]
ewollesen: This should be implemented in this branch now.

alerts/client.go
```diff
@@ -16,22 +16,20 @@ import (
 
 // Client for managing alerts configs.
 type Client struct {
-	client        PlatformClient
-	logger        platformlog.Logger
-	tokenProvider auth.ServerSessionTokenProvider
```
ewollesen, line 21: Darin upstreamed some changes that remove the necessity of a separate token provider.

alerts/config.go
```diff
@@ -142,20 +296,91 @@ func (a NoCommunicationAlert) Validate(validator structure.Validator) {
 	validator.Duration("delay", &dur).InRange(0, 6*time.Hour)
 }
 
+// Evaluate if CGM data is being received by Tidepool.
+//
+// Assumes data is pre-sorted by Time in descending order.
+func (a NoCommunicationAlert) Evaluate(ctx context.Context, data []*glucose.Glucose) *Notification {
```
ewollesen, line 302: All of NoCommunicationAlert handling will be re-worked in a later PR. It can be disregarded for now.

ewollesen:
Ready for review once again.

data/events/recorder.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"context"
+
+	"github.com/tidepool-org/platform/alerts"
+	"github.com/tidepool-org/platform/errors"
+	"github.com/tidepool-org/platform/log"
+	lognull "github.com/tidepool-org/platform/log/null"
+)
+
+type Recorder struct {
```
toddkazakov, line 12: This is vague. Can we use a more descriptive name?

data/service/api/v1/alerts.go
```diff
@@ -24,6 +24,8 @@ func AlertsRoutes() []service.Route {
 		service.Get("/v1/users/:userId/followers/:followerUserId/alerts", GetAlert, api.RequireAuth),
 		service.Post("/v1/users/:userId/followers/:followerUserId/alerts", UpsertAlert, api.RequireAuth),
 		service.Delete("/v1/users/:userId/followers/:followerUserId/alerts", DeleteAlert, api.RequireAuth),
+		service.Get("/v1/users/:userId/followers/alerts", ListAlerts, api.RequireServer),
+		service.Get("/v1/users/without_communication", GetUsersWithoutCommunication, api.RequireServer),
```
toddkazakov, line 28: This API endpoint doesn't return users but `LastCommunication` and should be renamed to better reflect the resource it returns.

data/store/mongo/mongo_recorder.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (r *recorderRepo) filter(lastComm alerts.LastCommunication) map[string]any {
+	return map[string]any{
+		"userId":    lastComm.UserID,
+		"dataSetId": lastComm.DataSetID,
+	}
+}
+
+func (d *recorderRepo) UsersWithoutCommunication(ctx context.Context) ([]alerts.LastCommunication, error) {
+	start := time.Now().Add(-5 * time.Minute)
```
toddkazakov, line 60: Shouldn't this value be passed as a parameter?

data/store/mongo/mongo_recorder.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (r *recorderRepo) filter(lastComm alerts.LastCommunication) map[string]any {
+	return map[string]any{
+		"userId":    lastComm.UserID,
+		"dataSetId": lastComm.DataSetID,
+	}
+}
+
+func (d *recorderRepo) UsersWithoutCommunication(ctx context.Context) ([]alerts.LastCommunication, error) {
```
toddkazakov, line 59: Doesn't return "users", consider renaming this.

data/store/mongo/mongo_recorder.go
```diff
@@ -0,0 +1,74 @@
...
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+
+	"github.com/tidepool-org/platform/alerts"
+	"github.com/tidepool-org/platform/errors"
+	structuredmongo "github.com/tidepool-org/platform/store/structured/mongo"
+)
+
+// recorderRepo implements RecorderRepository, writing data to a MongoDB collection.
```
toddkazakov, line 17: Rename this to something more descriptive

alerts/config.go
```diff
@@ -31,57 +37,254 @@ type Config struct {
...
+
+	if c.Alerts.NoCommunication == nil || !c.Alerts.NoCommunication.Enabled {
+		return nil, false
+	}
+
+	lgr := c.LoggerWithFields(log.LoggerFromContext(ctx))
+	ctx = log.NewContextWithLogger(ctx, lgr)
+	nc := c.Alerts.NoCommunication.Evaluate(ctx, last)
+	needsUpsert := c.Activity.NoCommunication.Update(nc.OutOfRange)
+	// TODO check re-eval? I don't think so
```
toddkazakov, line 205: Open a ticket if this needs to be addressed and remove the TODO comment

alerts/config.go
```diff
@@ -6,20 +6,26 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"os"
+	"slices"
 	"time"
 
 	"github.com/tidepool-org/platform/data"
-	"github.com/tidepool-org/platform/data/blood/glucose"
+	nontypesglucose "github.com/tidepool-org/platform/data/blood/glucose"
```
toddkazakov, line 14: Rename this (and other occurrences) to `dataBloodGlucose` for consistency. I don't think `nontypesglucose` has any benefit over the existing alias used throughout the repository.

alerts/config.go
```diff
@@ -214,31 +641,57 @@ type Threshold ValueWithUnits
...
+type Notification struct {
+	// Message communicates the alert to the recipient.
+	Message         string
+	RecipientUserID string
+	FollowedUserID  string
+	Sent            func(time.Time)
+}
+
+// RecordsRepository encapsulates queries of the records collection for use with alerts.
+type RecordsRepository interface {
```
toddkazakov, line 684: I think it's important to capture what type of "communications" this repository stores and that those are used solely for alerting purposes. This applies to the struct, interface and collection. Unfortunately, I can't come up with a good suggestion myself.

toddkazakov (changes requested):

data/store/mongo/mongo_recorder.go
ewollesen: Could you share your reasoning? I don't see any other value that makes sense.

Any value less than 5 minutes doesn't qualify to send an alert, which is why this value is used. I'll make that more obvious by defining a named constant.

data/service/service/standard.go
```diff
@@ -419,3 +442,98 @@ func (s *Standard) initializeUserEventsHandler() error {
...
+		Brokers: commonConfig.KafkaBrokers,
+		GroupID: "alerts",
+		Topics:  prefixedTopics,
+		Sarama:  commonConfig.SaramaConfig,
+		MessageConsumer: &dataEvents.AlertsEventsConsumer{
+			Consumer: ec,
+		},
+	}
+
+	retryDelays := []time.Duration{0, 1 * time.Second}
```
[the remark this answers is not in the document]
ewollesen: Done.

See also https://github.com/tidepool-org/development/pull/314

alerts/config.go
ewollesen: I've taken a stab at a better name. Let me know what you think.

alerts/tasks.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+}
+
+func (r *CarePartnerRunner) scheduleNextRun(tsk *task.Task, lastStart time.Time) {
+	// Ideally, we would start the next run 1 second after this run...
+	nextDesiredRun := lastStart.Add(time.Second)
+	now := time.Now()
+	if nextDesiredRun.Before(now) {
+		r.logger.Info("care partner is bumping nextDesiredRun")
+		// nextDesiredRun, when added to time.Now in tsk.RepeatAvailableAfter, must
```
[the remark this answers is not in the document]
ewollesen: @darinkrauss any thoughts or advice on this before I spend too much time working on it?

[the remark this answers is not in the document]
darinkrauss: @ewollesen I *think* that the change you'd need to make is, in `task/queue/queue.go`:
```
func (q *queue) computeState(tsk *task.Task) {
	switch tsk.State {
	case task.TaskStatePending:
		if tsk.AvailableTime == nil || time.Now().After(*tsk.AvailableTime) {
			tsk.AppendError(errors.New("pending task requires future available time"))
			tsk.SetFailed()
		}
```
to:
```
func (q *queue) computeState(tsk *task.Task) {
	switch tsk.State {
	case task.TaskStatePending:
        if tsk.AvailableTime == nil {
            tsk.AvailableTime = time.Now()
        } else if time.Now().After(*tsk.AvailableTime) {
			tsk.AppendError(errors.New("pending task requires future available time"))
			tsk.SetFailed()
		}
```
I think it is okay, but there would need to be a deeper look into Dexcom task to see if it would be an issue. I don't think so, but :shrug:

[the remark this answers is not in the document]
ewollesen: @darinkrauss @toddkazakov made the change, and added a method to make use of it.

I don't see any problems with the dexcom task, but you'd be the expert there for sure Darin. As far as I could see, the dexcom task always either have an available time, or are in an failed state (and so available time doesn't come into play).

task/queue/queue.go
```diff
@@ -374,7 +374,9 @@ func (q *queue) completeTask(ctx context.Context, tsk *task.Task) {
 func (q *queue) computeState(tsk *task.Task) {
 	switch tsk.State {
 	case task.TaskStatePending:
-		if tsk.AvailableTime == nil || time.Now().After(*tsk.AvailableTime) {
+		if tsk.AvailableTime == nil {
+			tsk.AvailableTime = pointer.FromAny(time.Now())
+		} else if time.Now().After(*tsk.AvailableTime) {
 			tsk.AppendError(errors.New("pending task requires future available time"))
 			tsk.SetFailed()
 		}
```
darinkrauss, line 382: Actually, after thinking a bit more on this, let's change this to:
```suggestion
		if now := time.Now(); tsk.AvailableTime == nil || tsk.AvailableTime.Before(now) {
			tsk.AvailableTime = pointer.FromAny(now)
		}
```

darinkrauss: This should fully resolve any edge cases (due to timing, etc.) plus allow for `nil` to just mean "ASAP".

data/events/alerts.go
```diff
@@ -0,0 +1,233 @@
+package events
```
darinkrauss, line 1: Is this part of the formal (public) Tidepool data model or is it just an internal-only (never to be exposed to the user)? If public, then I'd like to review and comment, but it'll need to wait a bit until I have more time.

task/service/service/service.go
```diff
@@ -359,6 +388,75 @@ func (s *Service) initializeTaskQueue() error {
 	return nil
 }
 
+func (s *Service) initializeAlertsClient() error {
+	s.Logger().Debug("initializing alerts client")
+
+	platformConfig := platform.NewConfig()
+	platformConfig.UserAgent = s.UserAgent()
+	reporter := s.ConfigReporter().WithScopes("data", "client")
```
darinkrauss, line 396: Why is this using "data" client, shouldn't this be "alert(s)" client?

task/service/service/service.go
```diff
@@ -359,6 +388,75 @@ func (s *Service) initializeTaskQueue() error {
...
+	}
+	s.alertsClient = alerts.NewClient(platformClient, s.Logger())
+
+	return nil
+}
+
+func (s *Service) initializePusher() error {
+	var err error
+
+	apns2Config := &struct {
```
darinkrauss, line 416: Typically we define the config struct alongside the object is is configuring (APNSPusher, in this case). It allows greatly isolation of knowledge. Also, my personal preference is that we keep using ConfigReporter rather than adding in yet another configuration mechanism.

toddkazakov: I would like us to move away from config reporter. With many seemingly similar components which use generic names (services, clients, service clients, standard clients, etc) it's absolutely impossible to determine where an env variable is used and how the configuration is loaded.

darinkrauss: I agree, that makes complete sense to me. I agree that ConfigReporter isn't completely obvious and moving to a more direct mechanism for configuration is a good idea.

I think, however, we should have a specific, agreed-upon plan for doing so. I'd rather not have multiple alternatives to ConfigReporter in the code, but one agreed upon solution going forward. Plus, I think we may be able to update the various configs with specific struct fields and apply a surgical change to ConfigReporter that would prevent a large change to remove ConfigReporter (at least in the short-term).

ewollesen: It gets worse. While envconfig is a nice package, it doesn't work the way we need it to. The problem comes when you start embedding structs and needing to alter the "prefix" of the environment vars that get looked up based on that embedding. 

For example [auth.Config](https://github.com/tidepool-org/platform/blob/master/auth/client/client.go#L16-L19) embeds [platform.Config](https://github.com/tidepool-org/platform/blob/master/platform/config.go#L10-L14) which in turn embeds [client.Config](https://github.com/tidepool-org/platform/blob/master/client/config.go#L12-L26).

That client.Config has a field called Address. The auth client wants the value found at TIDEPOOL_AUTH_CLIENT_ADDRESS, however the data client wants the value at TIDEPOOL_DATA_SERVICE_ADDRESS. But there's no way to communicate that from the auth client's auth.Config or the data client's platform.Config (data client doesn't have it's own data.Config).

So if we were to try to use only envconfig, we're now faced with a prefix problem. That is, it isn't possible with envconfig to annotate an embedded struct such that the value in the annotation is appended to the existing prefix. Even if that _were_ possible, we wouldn't know what that prefix should be.

To make this concrete, here are the same three structs, stripped down a little and re-annotated (bluntly) for use with envconfig:

```go
// client.Config
type Config struct {
    Address string `envconfig:"ADDRESS"` // There's no way to know what prefix to use yet,
                                         // this struct is only used after being embedded in
                                         // another struct. That is, we want this to be
                                         // TIDEPOOL_AUTH_CLIENT_ADDRESS. If that prefix
                                         // could be passed down, we might be onto
                                         // something.
    UserAgent string `envconfig:"USER_AGENT"` // Assuming we could pass the prefix down to
                                              // Address above, we _don't_ want a prefix
                                              // here! That said, we've never changed the
                                              // user agent that I'm aware of, so maybe this
                                              // could be removed as a config option? But
                                              // that's not the point. The point is: while
                                              // some embedded struct fields want to inherit
                                              // a prefix, others don't. So not only would
                                              // we have to enhance envconfig with a way to
                                              // inherit prefixes, we'd have to add a tag to
                                              // bypass prefix inheritance.
}

// platform.Config
type Config struct {
    *client.Config // There's no way to know what the prefix should be here, because
                   // platform.Config is always used from within another client's Config
                   // (like auth.Config below), so even if we could add a prefix here (which
                   // we can't), we wouldn't know what it should be! Perhaps we wouldn't
                   // need/want a prefix here, but rather we'd just want to pass any given
                   // prefix along.
    ServiceSecret string `envconfig:"SERVICE_SECRET"` // This value too, needs a prefix, but
                                                      // platform doesn't know what that
                                                      // value should be. It's only the
                                                      // struct that embeds this one that
                                                      // knows what this value should be
                                                      // annotated with.
}

// auth.Config
type Config struct {
    *platform.Config `envconfig:"AUTH_CLIENT"` // Having this appended to the prefix would
                                               // be a start, but still not enough, because
                                               // we'd need the prefix to also be passed
                                               // down to the platform.Config, and from
                                               // there again down to the Address field
                                               // that's defined in client.Config.
    *ExternalConfig `envconfig:"AUTH_CLIENT"`
}
```

So... what's the solution? I'm always struck by how little actual code is in each of the various Config's. So my suggestion is to get rid of client.Config and platform.Config. Take the fields from them and copy them into the auth.Config, and make a data.Config that has them, and then use envconfig without any struct embedding. It would mean a little bit of duplication, but we gain simpler configuration. The only methods on client.Config and platform.Config relate to either validation or implementing a bridge interface between envconfig and ConfigReporter. The validations will be easy to copy over, and the bridge interface will be removed.

The duplication is unfortunate, but I think acceptable. It's not an area of high churn and it's all code that's easy to understand.

toddkazakov: > While envconfig is a nice package, it doesn't work the way we need it to. The problem comes when you start embedding structs and needing to alter the "prefix" of the environment vars that get looked up based on that embedding.

I would argue that this is something that we don't want and is an anti-pattern. You're familiar with the code above, but what you described is not obvious by just reading the code above. Environment variables are "flat" - they don't represent a tree structures - trying to make them fit one is a non-starter. Our config structs should be flat - if there is behaviour that is common to those config structs, then this behaviour should be put behind an interface.

I absolutely agree that duplication is perfectly fine as it simplifies the code a whole lot. Consider the following example that I think is much easier to understand.

```golang
type UserAgentConfig struct {
   UserAgent string `envconfig:"TIDEPOOL_SERVICE_USER_AGENT"`
}

type TaskServiceCredentials struct {
  ClientID string     `envconfig:"TIDEPOOL_TASK_SERVICE_CLIENT_ID"`
  ClientSecret string `envconfig:"TIDEPOOL_TASK_SERVICE_CLIENT_SECRET"`
}

// Implement Credentials interfaces
func (b TaskServiceCredentials) GetClientID() string {
  return b.ClientID
}

func (b TaskServiceCredentials) GetClientSecret() string {
  return b.ClientSecret
}

type BlobServiceCredentials struct {
  ClientID string     `envconfig:"TIDEPOOL_BLOB_SERVICE_CLIENT_ID"`
  ClientSecret string `envconfig:"TIDEPOOL_BLOB_SERVICE_CLIENT_SECRET"`
}

// Implement Credentials interfaces
func (b BlobServiceCredentials) GetClientID() string {
  return b.ClientID
}

func (b BlobServiceCredentials) GetClientSecret() string {
  return b.ClientSecret
}

type OAuthCredentials interface {
  GetClientID() string
  ClientClientSecret() string
}

type AuthenticationClientConfig struct {
  Address string `envconfig:"TIDEPOOL_AUTH_SERVICE_ADDRESS"`
}

type DataClientConfig struct {
  Address string `envconfig:"TIDEPOOL_DATA_SERVICE_ADDRESS"`
}

func InitTaskService() error {
  dataClientConfig := DataClientConfig{}
  if err := envconfig.Process("", &dataClientConfig); err != nil {
    return err
  }

  authClientConfig := AuthenticationClientConfig{}
  if err := envconfig.Process("", &authClientConfig); err != nil {
    return err
  }

  creds := TaskServiceCredentials{}
  if err := envconfig.Process("", &creds); err != nil {
    return err
  }

  ...
  dataClient := NewDataClient(dataClientConfig, authClientConfig, userAgentConfig, creds)
  ...
}

func NewDataClient(dataClientConfig DataClientConfig, authClientConfig AuthenticationClientConfig, userAgentConfig UserAgentConfig, credentials OAuthCredentials) DataClient {
  // auth client for obtaining and refreshing service tokens
  authClient := NewExternalAuthClient(authClientConfig, credentials)
  ...
}

func NewExternalAuthClient(authClientConfig AuthenticationClientConfig, credentials Credentials) {
  // configure the OAuth client here
}
```

Let's discuss this in a meeting.

ewollesen: > I absolutely agree that duplication is perfectly fine as it simplifies the code a whole lot. Consider the following example that I think is much easier to understand.

I'm in full agreement. Sorry if that wasn't clear.

ewollesen: ... but is this something we really want to tackle as part of this ticket? Because I don't think it should be. I'd be glad to work on these changes, in a follow-on ticket.

I've been tasked with "wrapping up" this area of the code to get it stable for storage. Adding widespread config changes seems out of scope.

task/service/service/service.go
ewollesen: The alerts client uses endpoints that are exposed by the data service, just like the data client. I saw that I could re-use this value and not have to make additional changes to other repos. Perhaps it would have been better if this had been the data service config, because then any client that needed to talk to the data service could use it with this kind of confusion.

Duplicating the value is a thing we can do, but that too might cause some confusion then there'd be two different clients that use the same values and both would need to be updated in the event of a change there, and it could be confusing as to why they're not in the same client in the first place. But it was decided a good while ago that we didn't want a separate alerts service.

I see downsides either way, so made a judgment call. Would a comment be of use here?

data/events/alerts.go
ewollesen: > Is this part of the formal (public) Tidepool data model or is it just an internal-only (never to be exposed to the user)? If public, then I'd like to review and comment, but it'll need to wait a bit until I have more time.

Users would never need to interact with this directly, no. The events here are the mechanism by which we trigger and send alerts to care partners.

This code is being "shelved" due to our recent partner's contract cancellation.

darinkrauss: Does "shelved" mean merge into `master`? If so, we need to be sure this code isn't intentionally or unintentionally used.

task/service/service/service.go
darinkrauss: I agree this is out-of-scope for this ticket, particularly if it is being "shelved" (though not sure what that means). 

With that said, I do think we should have a discussion. I do like the use of interfaces for configuration. I'd like to consider https://github.com/caarlos0/env as an alternative. I used it in a past project, it is seeing regular support, and does handle prefixes. 

I think part of the problem with our configuration isn't so much that the prefixes are a problem, but rather that the configuration is all over the place. What I'd like to see (and what has worked very well for me in the past) is a single global config for the service that has all of the configuration required for the service. It can use sub-structs with env prefixes, but it is very easy to see (and validate/test) that all required configuration is available immediately upon service startup.

In any case, a meeting at some point would be good.

toddkazakov: > I'd like to consider https://github.com/caarlos0/env as an alternative. 

This one would work too. I proposed envconfig, because it's already used in most of the services, including platform. 

> I think part of the problem with our configuration isn't so much that the prefixes are a problem, but rather that the configuration is all over the place.

I have lost numerous hours trying to piece together whether a value that's being configured in the helm chart is interpreted correctly not only where it's loaded, but where it's used too - both in development and production environments when debugging issues. Since configuration is passed from kubernetes to the services as env variables, it should easy to find the configuration values both in the helm chart and in the code. The only way to achieve this that I see is to use full environment variable names so they are searchable in both repositories.

> What I'd like to see (and what has worked very well for me in the past) is a single global config for the service that has all of the configuration required for the service.

I see the value in having the configuration in a single place, but having it split in modules has never been an issue for me. The two problems that deal with frequently are:
- I need to add a configuration value in the helm chart after I made a code change. What is the name of the environment variable that I should in the helm chart so it's loaded correctly (going from code to config)
- The service I am working on is experiencing an issue and it acts as if the environment is not configured correctly. The environment variables in the container look correct. I need to find the exact place in the code where a given environment variable is being used (going from config to code).

alerts/tasks.go
```diff
@@ -0,0 +1,228 @@
...
+func (r *CarePartnerRunner) GetRunnerType() string {
+	return task.CarePartnerType
+}
+
+func (r *CarePartnerRunner) GetRunnerTimeout() time.Duration {
+	return 30 * time.Second
+}
+
+func (r *CarePartnerRunner) GetRunnerDeadline() time.Time {
+	return time.Now().Add(30 * time.Second)
```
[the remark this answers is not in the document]
toddkazakov: Still not fixed. Left a new comment.

data/store/mongo/mongo_recorder.go
toddkazakov: `UsersWithoutCommunication` doesn't actually reflect what the implementation does - it only returns users which didn't have communication in the last 5 minutes. `UsersWithoutCommunication(300 * time.Second)` on the other hand reflects this. I see that you renamed this to `OverdueCommunications` which is better.

alerts/config.go
toddkazakov: Looks good

data/events/alerts.go
toddkazakov: @darinkrauss the alert configuration objects in `alerts/config.go` are part of the public API (not the data model, specifically). AFAIK, there aren't a new data types introduced as part of this work.

data/service/service/standard.go
```diff
@@ -419,3 +442,125 @@ func (s *Standard) initializeUserEventsHandler() error {
...
+	// intended to be the one place that the sarama Logger is initialized,
+	// before any of the properties that need it are run.
+	sarama.Logger = log.NewSarama(s.Logger())
+	return nil
+}
+
+func (s *Standard) initializePusher() error {
+	var err error
+
+	apns2Config := &struct {
```
toddkazakov, line 457: I think it makes sense to define this struct where the config values are used as Darin suggested.

toddkazakov (changes requested):

task/service/service/service.go
ewollesen: > I have lost numerous hours trying to piece together whether a value that's being configured in the helm chart is interpreted correctly not only where it's loaded, but where it's used too

and

> but having it split in modules has never been an issue for me

Seem to be in conflict with each other. I'm firmly in the camp of the first quoted segment, and fully agree with Darin's desire to have all config loaded as close to "main" as possible, then a simple config struct passed around from there as/if necessary. At least having all env vars loaded there.

data/store/mongo/mongo_recorder.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (r *lastCommunicationsRepo) filter(lastComm alerts.LastCommunication) map[string]any {
+	return map[string]any{
+		"userId":    lastComm.UserID,
+		"dataSetId": lastComm.DataSetID,
+	}
+}
+
+func (d *lastCommunicationsRepo) OverdueCommunications(ctx context.Context) ([]alerts.LastCommunication, error) {
+	start := time.Now().Add(-alerts.MinimumNoCommunicationDelay)
```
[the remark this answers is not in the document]
ewollesen: I think alerts makes more sense, as that's where it's used when validating an incoming alerts.Config. But I can alias it or move it here if it's a big deal to you.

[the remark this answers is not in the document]
ewollesen: Huh, yeah... This *should* be used in alerts to validate the minimum delay setting on the alerts.Config. But it appears not to be at the moment. I'll fix that.

data/service/service/standard.go
ewollesen: This struct isn't used anywhere else. You know that, because it's defined inline. There is nowhere else for it to be defined that makes any sense, as this is the one and only place it's used.

It's used only to get these env vars. I can switch to os.Getenv if you prefer. But defining the struct somewhere else is only going to make things more confusing.

task/service/service/service.go
toddkazakov: > Seem to be in conflict with each other?

Only in the context of platform. Take the clinic service for example - https://github.com/search?q=repo%3Atidepool-org%2Fclinic%20envconfig&type=code. Configuration is loaded in each module, however it's pretty clear the source of the configuration values.

> I'm firmly in the camp of the first quoted segment, and fully agree with Darin's desire to have all config loaded as close to "main" as possible, then a simple config struct passed around from there as/if necessary.

I am not convinced this is such a good idea, because it increases coupling and reminds me of the [God Object](https://en.wikipedia.org/wiki/God_object) anti-pattern. It's something I can live with as longs as I can `grep` environment variable names.

ewollesen:
> But it's not a big deal, so I'll go ahead and make the change.

No I won't. We discussed this via Slack and decided it wasn't necessary. The name change is enough.

task/service/service/service.go
ewollesen: Oh I definitely don't mean we should have a god object no. This is a dumb struct that does nothing but binds the necessary config together for any one given service.

data/store/mongo/mongo_recorder.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/tidepool-org/platform/alerts"
+	"github.com/tidepool-org/platform/errors"
+	structuredmongo "github.com/tidepool-org/platform/store/structured/mongo"
+)
+
+// lastCommunicationsRepo implements LastCommunicationsRepository, writing data to a
+// MongoDB collection.
+type lastCommunicationsRepo structuredmongo.Repository
+
+func (r *lastCommunicationsRepo) RecordReceivedDeviceData(ctx context.Context,
```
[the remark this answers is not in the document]
ewollesen: They do..? `r` is for repo in this case. I've generally named my receivers after the noun in the name, not the adjectives in front of it. That generally makes more sense to me, as I think of this struct as a "repo" first and foremost.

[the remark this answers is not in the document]
ewollesen: Oooh, I see there was one still using `d`, maybe from a previous name that I missed. Should be fixed now.

data/store/mongo/mongo_alerts.go
```diff
@@ -64,8 +101,60 @@ func (r *alertsRepo) EnsureIndexes() error {
...
-	return &alerts.Config{
-		UserID:         cfg.UserID,
-		FollowedUserID: cfg.FollowedUserID,
+	return bson.D{
+		{Key: "userId", Value: cfg.UserID},
+		{Key: "followedUserId", Value: cfg.FollowedUserID},
+	}
+}
+
+type alertsDataRepo structuredmongo.Repository
```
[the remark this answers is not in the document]
ewollesen: I'm not sure what would make things obvious to you, but I'll move them to the mongo.go and take a stab at it. Having a suggestion from you would help though, as I don't have any issue with the current name.

data/store/mongo/mongo_recorder.go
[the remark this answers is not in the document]
darinkrauss: Oh, how I wish Google hadn't screwed this up. It would have been so straightforward (and inline with keeping things clean and simple) to just use `self` or enforce some sort of standard (e.g. always lowercased first letter of struct name). This is one of those things that grates on my nerves. I know silly, but...

data/store/mongo/mongo_alerts.go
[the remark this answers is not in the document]
toddkazakov: Why is this a separate repository if it all it does is query the device data collection? It's not a distinct repository where alertable data is stored. My suggestion would be to move the implementation to the main device data repo.

data/service/service/standard.go
toddkazakov: This struct isn't use anywhere else, but the config values are used in the `push` package which provides an implementation of a push notification service. The push notification service can be reused - it's not a data service specific implementation. What if we wanted to use push notification in the prescription service (real example that has been discussed in the past)? We would need to define this same struct in the prescription service and duplicate the code for loading the env variables and instantiation of the pusher. This is how you end up with hundreds of lines of duplicated code that we have to maintain, e.g.:
- https://github.com/tidepool-org/platform/blob/master/blob/service/service.go
- https://github.com/tidepool-org/platform/blob/master/auth/service/service/service.go
- https://github.com/tidepool-org/platform/blob/master/data/service/service/standard.go

If the pusher is used by multiple services and you make a change to the configuration parameters or how the configuration is loaded, then you would need to make the exact same change in multiple places. My suggestion would be to define a named struct in the `push` package and instantiate it with a single call. I think we both agreed during our huddle on Wednesday that configuration should be defined where it's used but loaded as close to the service entry point as possible.

I would like to give you an example of an alternative design. Consider the [clinic client module](https://github.com/tidepool-org/platform/blob/master/auth/client/external.go#L32-L71). Both the configuration struct and initialization logic is contained in the [clinic module](https://github.com/tidepool-org/platform/blob/master/auth/client/external.go#L32-L71) (package). Using the clinic client in services which use `fx` for dependency injection is as easy as including the constructor in their [dependecy list](https://github.com/tidepool-org/platform/blob/master/prescription/application/application.go#L38). Services which manually instantiate their dependencies (e.g. the task service, need to [call a single function](https://github.com/tidepool-org/platform/blob/master/task/service/service/service.go#L263) to obtain a fully configured and ready to use clinic client.

data/store/mongo/mongo_recorder.go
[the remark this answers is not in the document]
toddkazakov: It's not a pattern that I'm used to seeing and looked strange at first sight. I'll leave it up to you to decide what's best.

data/store/mongo/mongo_alerts.go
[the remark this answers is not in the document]
ewollesen: Hmmm... Not sure. I think I did it originally to make testing easier (smaller interface), but now that I'm revisiting it, I think I can see the existing pattern and yeah, I think this can get merged back into the existing DataRepository.

Thank you for pointing it out.

data/service/service/standard.go
ewollesen: > We would need to define this same struct in the prescription service and duplicate the code for loading the env variables and instantiation of the pusher. 

Why would you duplicate the loading code? I would extract the struct out then, if I needed it in multiple places? Keep it simple / Laziness is a virtue / You aren't going to need it.

> then you would need to make the exact same change in multiple places

Only if you duplicated all of that code instead of extracting the struct out and reusing it.

I will analyze your example and attempt to mimic it for the pusher.

ewollesen: Well... I get to eat my own words.

I *did* duplicate that anonymous struct! Shame on me!

ewollesen: So yeah, I used two separate anonymous structs to load from differently named env vars. In this case, we don't want separate env vars, so I'm collapsing them down into a single struct in the alerts package. It's not in push, because the config is specific to care partner alerts. Other packages will likely want to use other values.

The downside of this is that now alerts needs to know about envconfig, which I think is unfortunate. I feel like the "main" package or maybe for platform the "service" package should be the one that knows about config values and how to load them. Oh well. If we re-arrange our config value stuff, we can look at moving things around then.

ewollesen: https://github.com/tidepool-org/development/pull/316

ewollesen: https://github.com/tidepool-org/cluster-qa2/pull/1563

I'm not deploying this until Monday, just to be safer. I don't need to ruin anyone's weekend.

ewollesen:
Rebased on latest master.

toddkazakov (approved):

toddkazakov (approved):

PR #721, opened 2024-05-07
adds an interface for sending APNs push notifications

ewollesen:
Basic steps are taken to allow for other push notification services to be easily added in the future.

BACK-2554


ewollesen:
Meant as a replacement/continuation of #704

PR #722, opened 2024-05-09
WIP: process alerts events

ewollesen:
Laying the foundation

BACK-2554


data/events/alerts.go
```diff
@@ -0,0 +1,381 @@
...
+		return log.NewContextWithLogger(ctx, c.Logger), nil
+	}
+	lgr, err := logjson.NewWithDefaults()
+	if err != nil {
+		return ctx, err
+	}
+	return log.NewContextWithLogger(ctx, lgr), nil
+}
+
+func (c *Consumer) evaluateAlerts(ctx context.Context, followedUserID, userID string) (err error) {
```
toddkazakov, line 146: Both kafka topics must be keyed by the followed user id, to avoid race conditions by ensuring that alerts for the same followed user are evaluated by the same consumer.

PR #724, opened 2024-05-28
[BACK-2989] Add Twiist provider

toddkazakov:


oauth/provider/provider.go
```diff
@@ -51,6 +51,11 @@ func NewProvider(name string, configReporter config.Reporter) (*Provider, error)
 	}
 	cfg.Scopes = SplitScopes(configReporter.GetWithDefault("scopes", ""))
 
+	authStyleInParams := configReporter.GetWithDefault("auth_style_in_params", "")
+	if authStyleInParams == "true" {
```
ewollesen, line 55: [strconv.ParseBool](https://pkg.go.dev/strconv#ParseBool) might have been a good fit here.

oauth/oauth.go
```diff
@@ -88,17 +92,24 @@ func (t *Token) Expire() {
...
 func (t *Token) RawToken() *oauth2.Token {
-	return &oauth2.Token{
+	token := &oauth2.Token{
 		AccessToken:  t.AccessToken,
 		TokenType:    t.TokenType,
 		RefreshToken: t.RefreshToken,
 		Expiry:       t.ExpirationTime,
 	}
+	if t.IdToken != "" {
+		token.WithExtra(map[string]any{
```
ewollesen, line 102: Neat. TIL.

ewollesen (commented):
My only concern is that there are 0 tests.

toddkazakov:
@ewollesen unfortunately, the newly added code is rather difficult to test in isolation - it requires a whole lot of mocking in order to test that a task is created on linking and the task is deleted when accounts are unlinked. Same holds true for the pre-existing Dexcom code which also lacks such tests. Given that there isn't much going on here, I think it's ok this to be tested manually by the QA team.

ewollesen (approved):
I can trust that you're making the right tradeoffs.

PR #725, opened 2024-05-29
[BACK-3011] Implement the GET device_logs endpoint.

lostlevels:


blob/service/api/v1/v1.go
```diff
@@ -45,7 +46,7 @@ func (r *Router) Routes() []*rest.Route {
 		rest.Delete("/v1/users/:userId/blobs", r.DeleteAll),
 
 		rest.Post("/v1/users/:userId/device_logs", r.CreateDeviceLogs),
-		rest.Get("/v1/users/:userId/device_logs", r.ListDeviceLogs),
+		rest.Get("/v1/users/:userId/device_logs", api.RequireServer(r.ListDeviceLogs)),
```
toddkazakov, line 49: Device logs should be accessible by clinics in clinical trials, so that endpoint is not exclusive to backend services.

blob/store/structured/mongo/devicelogs_repository.go
```diff
@@ -43,6 +51,63 @@ func (d *DeviceLogsRepository) EnsureIndexes() error {
...
+	query := bson.M{
+		"userId": userID,
+		"deletedTime": bson.M{
+			"$exists": false,
+		},
+	}
+	if filter.Start != nil {
+		query["startAtTime"] = bson.M{
+			"$gte": *filter.Start,
+			"$lte": *filter.Start,
```
toddkazakov, line 86:
```suggestion
```

blob/store/structured/mongo/devicelogs_repository.go
```diff
@@ -43,6 +51,63 @@ func (d *DeviceLogsRepository) EnsureIndexes() error {
...
+		}
+	}
+	if filter.End != nil {
+		query["endAtTime"] = bson.M{
+			"$gte": *filter.End,
```
toddkazakov, line 91:
```suggestion
```

blob/store/structured/mongo/devicelogs_repository.go
```diff
@@ -43,6 +51,63 @@ func (d *DeviceLogsRepository) EnsureIndexes() error {
...
+			"$lte": *filter.End,
```
toddkazakov, line 92:
```suggestion
			"$lt": *filter.End,
```

lostlevels:
Rebase

blob/blob.go
```diff
@@ -41,6 +41,7 @@ type Client interface {
 	DeleteAll(ctx context.Context, userID string) error
 	Get(ctx context.Context, id string) (*Blob, error)
 	GetContent(ctx context.Context, id string) (*Content, error)
+	GetDeviceLogsContents(ctx context.Context, userID string, filter *DeviceLogsFilter, pagination *page.Pagination) ([]*DeviceLogsContentRaw, error)
```
toddkazakov, line 44: I'm not sure this design is correct. For consistency with `GetContent`, we should be able to fetch contents of a single entry given an `id`, instead of returning the contents of multiple blobs.

blob/service/api/v1/v1_test.go
```diff
@@ -323,133 +323,133 @@ var _ = Describe("V1", func() {
...
-										},
-									}}))
-								})
-
-								parameterAssertions()
-							})
-						})
-					})
-				})
+				// Context("ListDeviceLogs", func() {
```
toddkazakov, line 327: Why is this test commented out?

toddkazakov (changes requested):

blob/blob.go
lostlevels: True but @nhamming requested yesterday a way to get the entire range of contents.

blob/service/api/v1/v1_test.go
lostlevels: Temporary so I could test it on QA.

toddkazakov (approved):

PR #727, opened 2024-05-31
quiet data provenance logs

ewollesen:
Warnings about missing user id fields were being logged in cases where we don't expect there to be a user id. Namely, when the request is using a server session token.

Waiting for #723 before merging.

### DIFF
--- a/data/service/api/v1/datasets_data_create.go
+++ b/data/service/api/v1/datasets_data_create.go
@@ -144,7 +144,7 @@ func CollectProvenanceInfo(ctx context.Context, req *rest.Request, authDetails r
 		token = token[len("bearer "):]
 	}
 
-	if token != "" {
+	if token != "" && shouldHaveJWT(authDetails) {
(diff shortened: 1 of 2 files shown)


ewollesen:
Assuming this is approved as is, would you have any objection to this be merged without QA as a "trivial change" as per our process docs?

darinkrauss (dismissed):
This looks fine. At some point we might want to clean up the entire provenance code block and separate it into a two separate paths based upon whether `IsService()` is true or not since we appear to want to capture different things in each case.

Bypassing QA would be up to Todd.

darinkrauss (approved):

ewollesen:
@toddkazakov does this count as a "trivial" change for you?

toddkazakov:
@ewollesen I think this is a question for @tjotala

tjotala (approved):
LGTM 🚢 

OK to bypass QA

PR #728, opened 2024-05-31
adds a mongodb query for alertable data

ewollesen:
Where alertable data means those data used when evaluating whether a defined alert configuration is triggered, and a notification needs to be sent.

BACK-2554

### DIFF
--- a/data/service/api/v1/users_datasets_create_test.go
+++ b/data/service/api/v1/users_datasets_create_test.go
@@ -5,10 +5,9 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/ant0ine/go-json-rest/rest"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
-	"github.com/ant0ine/go-json-rest/rest"
 
 	"github.com/tidepool-org/platform/alerts"
 	"github.com/tidepool-org/platform/auth"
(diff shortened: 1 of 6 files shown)


data/types/alertable/alertable.go
```diff
@@ -0,0 +1,31 @@
+package alertable
+
+// Package alertable extracts types used when querying data used to evaluate
+// alerts configurations.
+//
+// I tried to put these in the data/types package (as AlertableParams and
```
toddkazakov, line 6: I don't think this package should be under `data/types`. All other types in the package are device data structs.

ewollesen: > I don't think this package should be under `data/types`. All other types in the package are device data structs.

Where would you like to see it live?

ewollesen: I'll move it to data/store... is that better? Just let me know.

toddkazakov: `data/store` is probably ok

toddkazakov (approved):

ewollesen:
@tjotala how does this fit with QA? This is 100% new functionality, nothing to test for QA.

Todd (in slack) said:
> IMO, unit tests are fine. That’s new functionality, nothing is using it, no unrelated changes. As long as it builds, it should be fine, but please confirm with 
@tjotala 

I'm glad to do whatever folks think is best and fits with our processes.

tjotala (approved):
LGTM to merge as incremental work that does not affect current functionality. Will be tested along with the CPA work.

PR #731, opened 2024-06-11
adds UploadID to alerts.Config

ewollesen:
This will be used later to limit the data evaluated for care partner alerts.

BACK-2554
BACK-3047

### DIFF
--- a/alerts/config.go
+++ b/alerts/config.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"time"
 
+	"github.com/tidepool-org/platform/data"
 	"github.com/tidepool-org/platform/data/blood/glucose"
 	"github.com/tidepool-org/platform/structure"
 	"github.com/tidepool-org/platform/structure/validator"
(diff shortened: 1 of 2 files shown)


alerts/config.go
```diff
@@ -42,6 +46,7 @@ type Alerts struct {
 func (c Config) Validate(validator structure.Validator) {
 	validator.String("UserID", &c.UserID).Using(user.IDValidator)
 	validator.String("FollowedUserID", &c.FollowedUserID).Using(user.IDValidator)
+	validator.String("UploadID", &c.UploadID).Exists().Using(data.SetIDValidator)
```
darinkrauss, line 49: The first parameter should be the "JSON path" to the field so the error message has the same name with casing as the JSON field name. So:
```suggestion
	validator.String("userID", &c.UserID).Using(user.IDValidator)
	validator.String("followedUserID", &c.FollowedUserID).Using(user.IDValidator)
	validator.String("uploadID", &c.UploadID).Exists().Using(data.SetIDValidator)
```

darinkrauss (changes requested):
Looks good, with one minor change.

darinkrauss (approved):
LGTM!

PR #733, opened 2024-06-13
Include logger in structure validator, parser, and normalizer

darinkrauss:
- Include logger in structure validator, parser, and normalizer
- Update all downstream code
- Update tests to use test logger

This PR looks very large, but really it is just one change, adding a logger to the structure validator, parser, and normalizer. The important change is in the `structure` package, while everything else is just adding a logger to each validator, parser, and normalizer. Sorry, for the size of the PR, but it is really a fairly simple change.

### DIFF
--- a/alerts/config_test.go
+++ b/alerts/config_test.go
@@ -2,6 +2,7 @@ package alerts
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"strings"
 	"testing"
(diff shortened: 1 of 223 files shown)


darinkrauss:
[Realized I hadn't responded to this yet...]

There are a number of data structures with validation where the incoming data is slightly wrong and we are "fixing" that data. This allows the validation code to log warnings when this happens so we can monitor via the logs when and how frequently this is happening. In particular, this is used to monitor the ingested Dexcom data for the significant number of data "fixes" we add.

toddkazakov (approved):

darinkrauss:

PR #735, opened 2024-06-13
Improve client authentication and error handling

darinkrauss:
- Improve client authentication and error handling
- Use server session token provider rather than direct server session token with clients
- Separate platform-specific endpoint error handling from platform client
- Handle client response based upon client-specific error handler
- Update OAuth client to accept base client rather that constructing one
- Update OAuth refresh token error determination
- Fix stale server session tokens with long running tasks
- Update to latest
- Update tests

The current implementation of the various clients (based upon the root `client` package, and the `platform` client, in particular) had two significant issues:

1. Any client that relied on a server session token embedded in the context would sometimes fail if the server session token provided expired while the client was in use (failing with an "authentication token is invalid" error). This could happen occasionally with any client, but much more frequently happened with the Dexcom task clients (because doing a full fetch of a new Dexcom connection could take quite a while, giving a lot of time for the server session token to expire). The clients were all updated to use a server session token *provider* embedded in the context so *every* network request would use the *latest* server session token.

2. Whenever the `platform` client received a non-200 status code, it would *always* attempt to parse the response body for a `platform` `SerializedError` structure so that it could forward that error on. The `platform` client was used as the basis for other clients that didn't necessarily make network requests to `platform`-based services and the response body parsing would completely fail (either resulting in erroneous log messages) or, worse, if the response body included something parsable it generate a parsed error unrelated to the actual network error. The code was updated so the client can be configured to use an `ErrorParser` or not. For non-`platform`-based services there is a `NewLegacyClient` function to create a client that does not parse the response body. Furthermore, the handling of errors and capturing the response body was streamlined and fixed.

Also, some changes to the `oauth` client and the third-party OAuth library was updated.


toddkazakov (approved):

auth/client/client_test.go
```diff
@@ -32,14 +36,21 @@ var _ = Describe("Client", func() {
...
 	BeforeEach(func() {
+		serverSessionTokenProviderController = gomock.NewController(GinkgoT())
+		serverSessionTokenProvider = authTest.NewMockServerSessionTokenProvider(serverSessionTokenProviderController)
 		serverTokenSecret = authTest.NewServiceSecret()
 		serverTokenTimeout = testHttp.NewTimeout()
 		name = test.RandomStringFromRangeAndCharset(4, 16, test.CharsetAlphaNumeric)
 		logger = logTest.NewLogger()
 		Expect(logger).ToNot(BeNil())
 		serverToken = authTest.NewSessionToken()
+		serverSessionTokenProvider.EXPECT().ServerSessionToken().Return(serverToken, nil).AnyTimes()
```
jh-bate, line 47: just a query `.EXPECT()` ?? I see its generated I just didn't expect (haha) that casing

ehr/reconcile/runner_test.go
```diff
@@ -148,7 +148,6 @@ var _ = Describe("Runner", func() {
 					tasksList = append(tasksList, &t)
 				}
 
-				authClient.EXPECT().ServerSessionToken().Return("token", nil)
 				clinicsClient.EXPECT().ListEHREnabledClinics(gomock.Any()).Return(clinics, nil)
```
jh-bate, line 151: ahh ok - now I see it else where :)

jh-bate (approved):
LGTM and til about `EXPECT`

auth/client/client_test.go
darinkrauss: Yea, this is how https://github.com/golang/mock generates.

darinkrauss:

PR #736, opened 2024-06-13
Fix task service shutdown, errors, and task handling

darinkrauss:
- Use context deadline to forcibly cancel tasks at extended timeout
- Allow task queue shutdown to fully process any outstanding tasks
- Separate task queue shutdown into worker and manager groups
- Ensure critical database updates are not affected by context cancel
- Fix task service termination
- Fix pending iterator usage and error handling
- Use platform errors
- Minor logging updates
- Update test runner

The existing code that detected a long running tasks (at 2 times the runner maximum duration) would simple update the task in the database to `pending` and available. This inadvertently caused the same task to be run again while the long running task was still running. This yielded a number of errors. The code was updated to cancel the context on long running tasks, which would force long running task to actually stop cleanly.

When the current task queue implementation was signaled to shutdown it would cancel the common context for the worker routines and the manager routine. The manager routine is responsible for taking finished tasks and storing the task in the database. Upon shutdown, however, many times the manager routine would exit before one or more of the work routines and so those tasks would not be updated in the database (and, thus, left permanently in the `running` state - this was why the "unstick tasks" code was written). The code was updated to separate out two cancel functions and two wait groups (one for workers, one for manager) and allow the manager to fully handle all of the worker tasks (saving them to the database) before itself shutdown. Unless there is a crash, there should no longer be forever `running` tasks.

The task manager did not protect critical database operations (such as saving the final state of a canceled task) from a context cancel. The code was updated to protect those critical database operations from a cancel.

Also, some smaller changes to the task-related code.
 
### DIFF
--- a/task/queue/queue.go
+++ b/task/queue/queue.go
@@ -67,10 +67,27 @@ func (c *Config) Validate() error {
 }
 
 type Runner interface {
+
+	// The type of tasks that the runner supports.
 	GetRunnerType() string
+
+	// The time after which the task manager will forcefully reset the task back to pending
+	// and available. This is calculated based upon the current time and a duration significantly
+	// longer that the task duration maximum. Normally this would only be used on a task that
+	// is in the running state even though it is not running (likely due to a system crash or interruption).
 	GetRunnerDeadline() time.Time
-	GetRunnerMaximumDuration() time.Duration
-	Run(ctx context.Context, tsk *task.Task) bool
+
+	// The duration of a task where the task manager will forcefully cancel the task context to interrupt
+	// the task and force completion. This is typically a duration somewhat longer than the task
+	// duration maximum.
+	GetRunnerTimeout() time.Duration
+
+	// The typical duration maximum of the task after which a warning will be displayed.
+	GetRunnerDurationMaximum() time.Duration
+
+	// Execute the specified task within the specified context. The context will be forcefully
+	// canceled after a duration specified by GetRunnerTimeout.
+	Run(ctx context.Context, tsk *task.Task)
 }
 
 type Queue interface {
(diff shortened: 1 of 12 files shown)


task/queue/queue.go
```diff
@@ -226,8 +256,15 @@ func (q *queue) startManager(ctx context.Context) {
 			}
 
 			select {
-			case <-ctx.Done():
-				return
+			case <-ctx.Done(): // Drain and complete any interrupted tasks
```
toddkazakov, line 259: Instead of doing a non-blocking select until there are no new messages, consider closing the channel after the workers waitgroup is done and drain the channel with `for tsk := range completeTask` to make the block below easier to understand.

task/service/service/service.go
```diff
@@ -269,6 +269,13 @@ func (s *Service) initializeClinicsClient() error {
 	return nil
 }
 
+func (s *Service) terminateClinicsClient() {
```
toddkazakov, line 272: What does this accomplish? Removing the reference doesn't terminate the client. I'm having trouble understanding why this code exists. Yes, it's consistent, but adds complexity and doesn't do anything useful.

task/queue/queue.go
darinkrauss: [Realized I hadn't responded to this yet...]

Okay, this makes sense. I added it to a later PR.

jh-bate (approved):

task/service/service/service.go
toddkazakov: I understand your point, but in the long term I think it's better to do the opposite and remove the termination code in the places where it doesn't actually terminate the components. There are at least a few hundred lines (even repeated in multiple services) that are confusing, because it's not obvious why they exist.

That's a nitpick, so it's ok to leave it as it is.

darinkrauss:

PR #742, opened 2024-06-24
Fix flaky test

toddkazakov:


ewollesen (dismissed):
So it appears that the issue was that the test was only counting tasks in the "pending" state when determining if it was done.

Todd has modified the code to look at tasks that are either "pending" or "running" before deciding if the test is completed. That seems very likely to fix the problem, assuming he got the list of non-terminal states correct (and I believe he did, see [here](https://github.com/tidepool-org/platform/blob/master/task/task.go#L31-L36)).

👍🏻 

As an aside, is there a reason that the goprocess on line 139 needs to be there? Couldn't it just be the for-loop (change the returns to breaks of course)? Not asking you to change it either way, just a curiosity.

Thanks for fixing this flappy test. :)

toddkazakov:
@ewollesen you're being too nice, I fixed the over-engineered code

task/queue/multi_test.go
```diff
@@ -131,39 +130,35 @@ var _ = Describe("multi queue", func() {
 
 			nonTerminalStates := []string{task.TaskStatePending, task.TaskStateRunning}
 
-			// Wait until completion (up to 10 seconds)
+			// Wait until completion (up to 15 seconds)
```
ewollesen, line 133: Ah good, you fixed this too. :D

ewollesen (approved):
LGTM

ewollesen:
There were travis issues fixed in #743 that this will need to pass travis.

PR #741, opened 2024-06-18
WIP allow legacy deduplication

jh-bate:

PR #747, opened 2024-07-09
minimal changes for smbg LegacyIdentityFields to use RawValue

jh-bate:
Updates so that the SMBG generated legacy id matches that produced by jellyfish. 

NOTE: only need a quick eyeball as it will become part of the main PR #741


jh-bate:
superseded by #749

PR #749, opened 2024-07-11
have generic raw object for units and value

jh-bate:
use a generic object to store raw units and value for SMBG. Use the raw details to generate the legacy identity

### DIFF
--- a/data/types/blood/blood.go
+++ b/data/types/blood/blood.go
@@ -5,14 +5,16 @@ import (
 
 	"github.com/tidepool-org/platform/data/types"
 	"github.com/tidepool-org/platform/errors"
+	"github.com/tidepool-org/platform/metadata"
 	"github.com/tidepool-org/platform/structure"
 )
 
 type Blood struct {
 	types.Base `bson:",inline"`
 
-	Units *string  `json:"units,omitempty" bson:"units,omitempty"`
-	Value *float64 `json:"value,omitempty" bson:"value,omitempty"`
+	Units *string            `json:"units,omitempty" bson:"units,omitempty"`
+	Value *float64           `json:"value,omitempty" bson:"value,omitempty"`
+	Raw   *metadata.Metadata `json:"raw,omitempty" bson:"raw,omitempty"`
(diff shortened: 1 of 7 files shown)


data/types/blood/glucose/selfmonitored/selfmonitored.go
```diff
@@ -66,18 +67,26 @@ func (s *SelfMonitored) Normalize(normalizer data.Normalizer) {
...
-	s.Glucose.Normalize(normalizer)
+	if normalizer.Origin() == structure.OriginExternal {
+
+		rawUnits := pointer.CloneString(s.Units)
+		rawValue := pointer.CloneFloat64(s.Value)
+		s.SetRawUnitsAndValue(rawUnits, rawValue)
+
+		s.Units = dataBloodGlucose.NormalizeUnits(rawUnits)
+		s.Value = dataBloodGlucose.NormalizeValueForUnits(rawValue, rawUnits)
+	}
```
darinkrauss, line 78: I *think* you can get rid of the Clone here if you just reverse the normalization of units and value, so this can become:
```suggestion
	if normalizer.Origin() == structure.OriginExternal {
	    s.SetRawUnitsAndValue(s.Units, s.Value)
		s.Value = dataBloodGlucose.NormalizeValueForUnits(s.Value, s.Units)
		s.Units = dataBloodGlucose.NormalizeUnits(s.Units)
	}
```

data/types/blood/blood.go
```diff
@@ -51,3 +53,46 @@ func (b *Blood) LegacyIdentityFields() ([]string, error) {
 		types.GetLegacyTimeField(b.Time),
 	)
 }
+
+func (b *Blood) GetRawUnitsAndValue() (*string, *float64, error) {
```
darinkrauss, line 57: [nit] Consider reversing the order of units and value to match the other functions such as `NormalizeValueForUnits`.

darinkrauss (approved):
LGTM. A couple of minor comments to consider.

PR #753, opened 2024-07-24
add a fallback for Makefile's git SHA checksum calculation

ewollesen:
Under certain conditions (my custom Tilt setup), when Tilt tries to build
platform services, it ignores the .git directory, which causes calculations of
the git SHA checksums to be empty. The data service refuses to start without
some value. This change provides a fallback in the event that no git
information is available.

This shouldn't affect any of our existing build/deployment processes, which
obviously do have git values included, so if you notice anything different,
let me know.


darinkrauss (approved):
LGTM!

PR #757, opened 2024-08-14
assume an AWS role if present via env var

ewollesen:


Roukoswarf (approved):

ewollesen:
this wasn't the issue that we were facing, no need to merge

PR #759, opened 2024-08-19
[BACK-3152] Remove created user id from the prescription integrity hash

toddkazakov:


prescription/integrity.go
```diff
@@ -14,7 +14,6 @@ const (
 
 type IntegrityAttributes struct {
```
darinkrauss, line 15: [nit] Technically no need for `IntegrityAttributes` now and could just use `DataAttributes`.

darinkrauss (approved):
LGTM!

PR #760, opened 2024-08-26
Allow any valid MIME types in device log blob contents.

lostlevels:


blob/service/api/v1/v1_test.go
```diff
@@ -608,6 +631,30 @@ var _ = Describe("V1", func() {
 										"End-At":       []string{content.EndAt.Format(time.RFC3339Nano)},
 									}))
 								})
+								It("responds successfully with headers for user's own logs content of type text/plain", func() {
+									deviceLogsBlob := blobTest.RandomDeviceLogsBlob()
```
darinkrauss, line 635: I don't think this test is accurate. From what I can tell, `blobTest.RandomDeviceLogsBlob()` uses `application/json; charset=utf-8` media type, while the description of this text states `text/plain`.

darinkrauss (changes requested):
One change, I think. Let me know if not.

darinkrauss (approved):
LGTM!

PR #763, opened 2024-09-04
[BACK-2638] Move Summary buckets to collection and unify datastructure

Roukoswarf:


data/store/mongo/mongo_summary.go
```diff
@@ -41,3 +41,36 @@ func (d *SummaryRepository) EnsureIndexes() error {
...
+	//return d.CreateAllIndexes(context.Background(), []mongo.IndexModel{
+	//	{
+	//		Keys: bson.D{
+	//			{Key: "userId", Value: 1},
+	//			{Key: "type", Value: 1},
+	//		},
+	//		Options: options.Index().
+	//			SetUnique(true).
+	//			SetName("UserIDTypeUnique"),
+	//	},
```
Roukoswarf, line 59: TODO: create proper indexes and uncomment this

data/summary/test/test_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package test_test
```
toddkazakov, line 1: You're breaking the convention here - the `test` package ought to contain only test helper functions (e.g. to generate data), not actual Ginkgo Tests

data/summary/types/glucose.go
```diff
@@ -0,0 +1,453 @@
...
+
+func (st *GlucosePeriods) Update(ctx context.Context, bucketsCursor *mongo.Cursor) error {
+	// count backwards (newest first) through hourly stats, stopping at 1d, 7d, 14d, 30d,
+	// currently only supports day precision
+	nextStopPoint := 0
+	nextOffsetStopPoint := 0
+	totalStats := GlucosePeriod{}
+	totalOffsetStats := GlucosePeriod{}
+	offsetPeriods := make(GlucosePeriods)
+	bucket := &Bucket[*GlucoseBucket, GlucoseBucket]{}
```
toddkazakov, line 371: You’re defining the bucket here and it has attributes which have omitempty. This means that the attributes won’t be serialized in the mongo document. Subsequently when such bucket is deserialized the deserializer for the attribute won’t be called.

Imagine processing the following mongo documents:
```
{
  data: {
    total: {
      glucose: 5
    }
  }
},
{
  data: {
    total: {
    }
  }
}
```

and you define the bucket before its intended scope of use:
```
bucket := &Bucket[*GlucoseBucket, GlucoseBucket]{}
for bucketsCursor.Next(ctx) {
  if err := bucketsCursor.Decode(bucket); err != nil {
    return err
  }
  // You're only using the bucket here
  fmt.Println(bucket.Data.Total.Glucose)
}
```

The above code will print
```
5
5
```

Instead of the actual values that were used before persisting the structs in mongo:
```
5
0
```

data/summary/types/util.go
```diff
@@ -8,10 +8,15 @@ import (
 	"golang.org/x/exp/constraints"
 
 	glucoseDatum "github.com/tidepool-org/platform/data/types/blood/glucose"
+	"github.com/tidepool-org/platform/data/types/blood/glucose/continuous"
 )
 
 // GetDuration assumes all except freestyle is 5 minutes
 func GetDuration(dataSet *glucoseDatum.Glucose) int {
+	if dataSet.Type != continuous.Type {
```
toddkazakov, line 16:
```suggestion
	if datum.Type != continuous.Type {
```

data/summary/types/glucose.go
```diff
@@ -0,0 +1,449 @@
...
+	"github.com/tidepool-org/platform/data/blood/glucose"
+	glucoseDatum "github.com/tidepool-org/platform/data/types/blood/glucose"
+	"github.com/tidepool-org/platform/data/types/blood/glucose/continuous"
+)
+
+const MaxRecordsPerBucket = 60 // one per minute max
+
+type GlucosePeriods map[string]*GlucosePeriod
+
+type Range struct {
```
toddkazakov, line 23: Remove `omitempty` to avoid bugs

data/summary/types/glucose.go
```diff
@@ -0,0 +1,449 @@
...
+	rs.High.CalculateDelta(&current.High, &previous.High)
+	rs.VeryHigh.CalculateDelta(&current.VeryHigh, &previous.VeryHigh)
+	rs.ExtremeHigh.CalculateDelta(&current.ExtremeHigh, &previous.ExtremeHigh)
+	rs.AnyLow.CalculateDelta(&current.AnyLow, &previous.AnyLow)
+	rs.AnyHigh.CalculateDelta(&current.AnyHigh, &previous.AnyHigh)
+}
+
+type GlucoseBucket struct {
+	GlucoseRanges      `json:",inline" bson:",inline"`
+	LastRecordDuration int `json:"lastRecordDuration,omitempty" bson:"lastRecordDuration,omitempty"`
```
toddkazakov, line 206: Remove `omitempty` to avoid bugs

data/summary/types/glucose.go
```diff
@@ -0,0 +1,449 @@
...
+		return false, nil
+	}
+
+	b.GlucoseRanges.Update(record)
+	b.LastRecordDuration = GetDuration(record)
+
+	return true, nil
+}
+
+type GlucosePeriod struct {
```
toddkazakov, line 246: Remove `omitempty` to avoid bugs

data/summary/test/generators.go
```diff
@@ -0,0 +1,427 @@
+package test
```
toddkazakov, line 1: Move under `data/test`

data/summary/test/summary.go
```diff
@@ -0,0 +1,186 @@
+package test
```
toddkazakov, line 1: Move under `data/test`

toddkazakov (changes requested):
Left feedback inline - overall looks good, but there are few issues

data/client/client.go
```diff
@@ -6,7 +6,8 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/tidepool-org/platform/data/summary/types"
+	types2 "github.com/tidepool-org/platform/summary/types"
```
toddkazakov: Rename to `summaryTypes` for consistency

PR #766, opened 2024-09-17
Makefile updates; add ginkgo labels to integration tests

ewollesen:
Two minor updates

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,13 @@ TRANSFORM_GO_BUILD_CMD:=sed 's|\.\(.*\)\(/[^/]*\)/[^/]*|_bin\1\2\2 .\1\2/.|'
 
 GO_BUILD_CMD:=go build $(GO_BUILD_FLAGS) $(GO_LD_FLAGS) -o
 
+GINKGO_FLAGS += --require-suite --poll-progress-after=10s --poll-progress-interval=20s -r
+GINKGO_CI_WATCH_FLAGS += --randomize-all --succinct --fail-on-pending --cover --trace --race
+GINKGO_CI_FLAGS += $(GINKGO_CI_WATCH_FLAGS) --randomize-suites --keep-going
+
+GOTEST_PKGS ?= ./...
+GOTEST_FLAGS ?=
+
 ifdef TRAVIS_BRANCH
 ifdef TRAVIS_COMMIT
     DOCKER:=true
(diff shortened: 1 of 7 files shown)


toddkazakov (approved):

Makefile
```diff
@@ -191,29 +191,39 @@ service-restart-all:
 	@cd $(ROOT_DIRECTORY) && for SERVICE in $(shell ls -1 services) ; do $(MAKE) service-restart SERVICE="services/$${SERVICE}"; done
 	@cd $(ROOT_DIRECTORY) && for SERVICE in migrations tools; do $(MAKE) service-restart SERVICE="$${SERVICE}"; done
 
+GINKGO_FLAGS += --require-suite --poll-progress-after=10s --poll-progress-interval=20s -r # Add flags, e.g. GINKGO_FLAGS="--json-report=report.json" make test
```
darinkrauss, line 194: I'm good with this change, but please move this declaration and that of `GINKGO_CI_FLAGS`, `GOTEST_PKGS`, and `GOTEST_FLAGS` to the top of the file so they are easily visible (like all of the rest currently are). Also, the comment isn't necessary since it seems pretty obvious to me.

Makefile
```diff
@@ -191,29 +191,39 @@ service-restart-all:
...
+	@cd $(ROOT_DIRECTORY) && . ./env.test.sh && ginkgo $(GINKGO_FLAGS) $(GINKGO_CI_FLAGS) $(TEST)
 
 ci-test-until-failure: ginkgo
-	@echo "ginkgo --require-suite --poll-progress-after=10s --poll-progress-interval=20s -r --randomize-suites --randomize-all --succinct --fail-on-pending --cover --trace --race --keep-going -untilItFails $(TEST)"
-	@cd $(ROOT_DIRECTORY) && . ./env.test.sh && ginkgo --require-suite --poll-progress-after=10s --poll-progress-interval=20s -r --randomize-suites --randomize-all --succinct --fail-on-pending --cover --trace --race --keep-going -untilItFails $(TEST)
+	@echo "ginkgo $(GINKGO_FLAGS) $(GINKGO_CI_FLAGS) -untilItFails $(TEST)"
+	@cd $(ROOT_DIRECTORY) && . ./env.test.sh && ginkgo $(GINKGO_FLAGS) $(GINKGO_CI_FLAGS) -untilItFails $(TEST)
 
 ci-test-watch: ginkgo
-	@echo "ginkgo watch --require-suite --poll-progress-after=10s --poll-progress-interval=20s -r --randomize-all --succinct --fail-on-pending --cover --trace --race $(TEST)"
```
darinkrauss, line 215: Note that you added `--randomize-suites` and `--keep-going` to this target with your change.

darinkrauss (changes requested):
LGTM with one requested change (moving Makefile variables to top of file).

Makefile
ewollesen: My preference is to have them there where I can see them without scrolling to the top of the file and back. Preferences differ. I'll change them.

Makefile
ewollesen: Thank you, fixed.

darinkrauss (approved):
LGTM.

Yea, I could go either way with the location of the variables (at the top OR inline where/when needed), but having some in both areas is just confusing because you have to look in both places.

PR #767, opened 2024-09-18
increase carb ratio based on findings from prod data testing

jh-bate:
based on actual values seen from production data we have increased this threshold

### DIFF
--- a/data/store/mongo/mongo_datum.go
+++ b/data/store/mongo/mongo_datum.go
@@ -73,13 +73,16 @@ func (d *DatumRepository) EnsureIndexes() error {
 		},
 		{
 			Keys: bson.D{
+				{Key: "_userId", Value: 1},
 				{Key: "origin.id", Value: 1},
-				{Key: "type", Value: 1},
 				{Key: "deletedTime", Value: -1},
 				{Key: "_active", Value: 1},
 			},
 			Options: options.Index().
-				SetName("OriginId"),
+				SetPartialFilterExpression(bson.D{
+					{Key: "origin.id", Value: bson.D{{Key: "$exists", Value: true}}},
+				}).
+				SetName("UserIdOriginId"),
 		},
 		{
 			Keys: bson.D{
@@ -210,7 +213,6 @@ func (d *DatumRepository) ArchiveDataSetData(ctx context.Context, dataSet *uploa
 
 	selector["_userId"] = dataSet.UserID
 	selector["uploadId"] = dataSet.UploadID
-	selector["type"] = bson.M{"$ne": "upload"}
 	selector["_active"] = true
 	selector["deletedTime"] = bson.M{"$exists": false}
 	set := bson.M{
@@ -224,7 +226,7 @@ func (d *DatumRepository) ArchiveDataSetData(ctx context.Context, dataSet *uploa
 	}
 	opts := options.Update()
 	if hasOriginID {
-		opts.SetHint("OriginId")
+		opts.SetHint("UserIdOriginId")
 	}
 	changeInfo, err := d.UpdateMany(ctx, selector, d.ConstructUpdate(set, unset), opts)
 	if err != nil {
@@ -254,7 +256,6 @@ func (d *DatumRepository) DeleteDataSetData(ctx context.Context, dataSet *upload
 
 	selector["_userId"] = dataSet.UserID
 	selector["uploadId"] = dataSet.UploadID
-	selector["type"] = bson.M{"$ne": "upload"}
 	selector["deletedTime"] = bson.M{"$exists": false}
 	set := bson.M{
 		"_active":      false,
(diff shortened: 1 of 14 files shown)


data/types/bolus/combination/combination.go
```diff
@@ -61,7 +61,7 @@ func (c *Combination) Validate(validator structure.Validator) {
 	}
 
 	if c.NormalExpected != nil {
-		validator.Int("duration", c.Duration).Exists().EqualTo(DurationMinimum)
+		validator.Int("duration", c.Duration).Exists().InRange(DurationMinimum, DurationMaximum)
```
darinkrauss, line 64: Given our discussion, I think we need to completely separate the normal validation from the extended/duration validation. That is, I think we need to assume that a combo can have any valid normal along with any valid extended/duration. There should not be any assumption that it one is interrupted, then the other will be.

Thus, I think we need to change line 63-124 to:
```
	validator.Int("duration", c.Duration).Exists().InRange(DurationMinimum, DurationMaximum)
	durationExpectedValidator := validator.Int("expectedDuration", c.DurationExpected)
	if c.Duration != nil && *c.Duration >= DurationMinimum && *c.Duration <= DurationMaximum {
		durationExpectedValidator.InRange(*c.Duration, DurationMaximum)
	} else {
		durationExpectedValidator.InRange(DurationMinimum, DurationMaximum)
	}
	if c.ExtendedExpected != nil {
		durationExpectedValidator.Exists()
	} else {
		durationExpectedValidator.NotExists()
	}
	validator.Float64("extended", c.Extended).Exists().InRange(ExtendedMinimum, ExtendedMaximum)
	extendedExpectedValidator := validator.Float64("expectedExtended", c.ExtendedExpected)
	if c.Extended != nil && *c.Extended >= ExtendedMinimum && *c.Extended <= ExtendedMaximum {
		extendedExpectedValidator.InRange(*c.Extended, ExtendedMaximum)
	} else {
		extendedExpectedValidator.InRange(ExtendedMinimum, ExtendedMaximum)
	}

	validator.Float64("normal", c.Normal).Exists().InRange(NormalMinimum, NormalMaximum)
	normalExpectedValidator := validator.Float64("expectedNormal", c.NormalExpected)
	if c.Normal != nil && *c.Normal >= NormalMinimum && *c.Normal <= NormalMaximum {
		normalExpectedValidator.InRange(*c.Normal, NormalMaximum)
	} else {
		normalExpectedValidator.InRange(NormalMinimum, NormalMaximum)
	}
```

The above code was copied from both `extended.go` and `normal.go` with a couple small changes. Basically, to **remove** the validation that if the normal/extended is 0 then there *must* be a expectedNormal/expectedExtended present. That is not the case. (This same validation should likely be removed from `extended.go` and `normal.go`, also, since I can totally see a pump sending a 0 normal bolus without expectedNormal).

What do you think? Does the above code make sense? What about the subsequent changes to `extended.go` and `normal.go`?

darinkrauss: BTW, you likely can copy much of the tests from `extended_test.go` and `normal_test.go`. Assuming we make the above mentioned changes to `extended.go` and `normal.go` I'd make the changes to those test files first. Then, copy the relevant sections from those test files into `combination_test.go` to replace much of the existing related tests.

Does that make sense?

darinkrauss (changes requested):
I think we need to change the validation for `combination.go`.

data/store/mongo/mongo_datum.go
```diff
@@ -73,13 +73,16 @@ func (d *DatumRepository) EnsureIndexes() error {
 		},
 		{
 			Keys: bson.D{
+				{Key: "_userId", Value: 1},
```
darinkrauss, line 76: Did you mean to include these Mongo changes in this PR?

data/types/bolus/combination/combination_test.go
```diff
@@ -99,1069 +98,653 @@ var _ = Describe("Combination", func() {
 				),
 				Entry("sub type invalid",
 					func(datum *combination.Combination) { datum.SubType = "invalidSubType" },
-					errorsTest.WithPointerSourceAndMeta(structureValidator.ErrorValueNotEqualTo("invalidSubType", "dual/square"), "/subType", &bolus.Meta{Type: "bolus", SubType: "invalidSubType"}),
+					errorsTest.WithPointerSourceAndMeta(structureValidator.ErrorValueNotEqualTo("invalidSubType", combination.SubType), "/subType", &bolus.Meta{Type: "bolus", SubType: "invalidSubType"}),
 				),
-				Entry("sub type dual/square",
-					func(datum *combination.Combination) { datum.SubType = "dual/square" },
+				Entry("sub type normal",
```
darinkrauss, line 103: [typo] normal -> dual/square

data/store/mongo/mongo_datum.go
darinkrauss: Did this get picked up accidentally due to a merge from master?

darinkrauss (changes requested):
A questions and minor change.

data/store/mongo/mongo_datum.go
jh-bate: ahh crap - no i didn't

darinkrauss (approved):
LGTM!

PR #768, opened 2024-09-20
Return proper 401 when session token expired in POST /v1/users/:userId/device_logs

lostlevels:
Update POST /v1/users/:userId/device_logs route to use newer style permission checking to fix an issue where service was responding with a 403 instead of 401 when a user's session token was expired.


toddkazakov (dismissed):

lostlevels:
Rebased

toddkazakov (approved):

PR #769, opened 2024-09-21
Allow specifying custom cadence for EHR reports

toddkazakov:


ehr/sync/task.go
```diff
@@ -34,3 +37,32 @@ func GetClinicId(data map[string]interface{}) (string, error) {
...
+	} else {
+		tsk.SetCompleted()
+	}
+}
+
+func GetCadence(data map[string]interface{}) *time.Duration {
+	cadence := data["cadence"].(string)
+	parsed, err := time.ParseDuration(cadence)
+	if err != nil {
+		return nil
```
ewollesen, line 61: I think it would simplify the logic if instead of returning `nil`, you returned `sync.DefaultCadence`. Both `ScheduleNextExecution` and `GetReconciliationPlan` check if the response is nil, and if so, use the default value.

To go one step further, you could instead return the error (`func GetCandence(...) (time.Duration, error)`) so the caller (who presumably has more context) could decide how to handle that (whether that means using the default, or maybe logging the error before using the default) or maybe using some other value. Just returning nil (or a default) hides the error entirely. If it's really not something we care about, that's fine. If you think there's ever a time you'd want to use that error, consider returning it.

But a minor thing to be sure, and fine how it is.

ehr/sync/task.go
```diff
@@ -34,3 +37,32 @@ func GetClinicId(data map[string]interface{}) (string, error) {
...
+		tsk.RepeatAvailableAfter(DefaultCadence)
+	} else if *cadence != 0 {
+		tsk.RepeatAvailableAfter(*cadence)
+	} else {
+		tsk.SetCompleted()
+	}
+}
+
+func GetCadence(data map[string]interface{}) *time.Duration {
+	cadence := data["cadence"].(string)
```
ewollesen, line 58: Uck. I get why this is here, but it's unfortunate. :( I guess someday we could update the task service now that Generics in Go are a thing.

ehr/reconcile/planner.go
```diff
@@ -0,0 +1,87 @@
+package reconcile
+
+import (
+	"context"
+
+	duration "github.com/xhit/go-str2duration/v2"
```
ewollesen, line 6: I'm not a fan of taking a dependency for this. We support literally four values, I don't think we need a 3rd-party dependency to handle that now, or in the future.

clinics/service.go
```diff
@@ -121,6 +122,17 @@ func (d *defaultClient) ListEHREnabledClinics(ctx context.Context) ([]clinic.Cli
 	return clinics, nil
 }
 
+func (d *defaultClient) GetEHRSettings(ctx context.Context, clinicId string) (*clinic.EHRSettings, error) {
+	response, err := d.httpClient.GetEHRSettingsWithResponse(ctx, clinicId)
+	if err != nil {
+		return nil, err
+	}
+	if response.StatusCode() != http.StatusOK || response.StatusCode() != http.StatusOK {
+		return nil, fmt.Errorf("unexpected response status code %v from %v", response.StatusCode(), response.HTTPResponse.Request.URL)
```
ewollesen, line 131: Consider if this should use the existing platform errors code.

clinics/service.go
```diff
@@ -121,6 +122,17 @@ func (d *defaultClient) ListEHREnabledClinics(ctx context.Context) ([]clinic.Cli
 	return clinics, nil
 }
 
+func (d *defaultClient) GetEHRSettings(ctx context.Context, clinicId string) (*clinic.EHRSettings, error) {
+	response, err := d.httpClient.GetEHRSettingsWithResponse(ctx, clinicId)
+	if err != nil {
+		return nil, err
```
ewollesen, line 128: This might be a place to wrap with the existing platform error code to gain the benefits it brings.

ehr/reconcile/planner.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil, err
+	}
+
+	// At the end of the loop syncTasks will contain only the tasks that need to be deleted,
+	// and toCreate will contain tasks for new clinics that need to be synced.
+	for _, clinic := range clinicsList {
+		clinicId := *clinic.Id
+		settings, err := p.clinicsClient.GetEHRSettings(ctx, clinicId)
+		if err != nil {
+			return nil, err
```
ewollesen, line 43: The `clinicsClient` appears to be generated code that doesn't return platform errors. Therefore you might consider if these should use `errors.Wrap`. (I won't mention this any more for this PR, but there are other places I haven't pointed out.)

ewollesen (approved):
LGTM. No blockers, but I'll reiterate that I'm against taking a dependency for the duration parsing we're doing here.

ehr/sync/task.go
toddkazakov: > Both ScheduleNextExecution and GetReconciliationPlan check if the response is nil, and if so, use the default value.

This statement is not true. Planner doesn't use the default value, it check whether the cadence in the task is difference that the one in the settings, and if that's the case it updates the task:
```golang
// Update the task if the configured cadence has been changed
cadenceFromTask := sync.GetCadence(tsk.Data)
if cadenceFromTask == nil || *cadenceFromTask != cadenceFromSettings {
    sync.SetCadence(tsk.Data, cadenceFromSettings)
    update := task.NewTaskUpdate()
    update.Data = &tsk.Data
    toUpdate[tsk.ID] = update
}
```

This function returns exactly what the name implies. If parsing fails, then the cadence is not correctly set so it will be updated as configured. I could return the parsing error, but it will be ignored by the caller.

ehr/reconcile/planner.go
toddkazakov: I removed the dependency. I don't want to have to update 2 services if we decide to support 60 day reports one day, so I had to implement the parsing logic myself. I think this results in less readable code.

ehr/reconcile/planner.go
toddkazakov: Can you please clarify what are the benefits of using the platform `errors.Wrap`, because I'm not familiar with those.

ehr/reconcile/planner_test.go
```diff
@@ -0,0 +1,190 @@
...
+				Expect(
+					[]string{*plan.ToDelete[0].Name, *plan.ToDelete[1].Name},
+				).To(
+					ConsistOf(sync.TaskName(*firstDeleted.Id), sync.TaskName(*secondDeleted.Id)),
+				)
+			})
+
+			It("returns a task for deletion when the report settings are disabled", func() {
+				settings := clinicsTest.NewRandomEHRSettings()
+				settings.ScheduledReports.Cadence = api.N0d
```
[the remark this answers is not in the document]
toddkazakov: Changed this to `DISABLED`. I'm not a big fan of the resulting code, but at least it makes it obvious to the reader when a task is disabled.

ewollesen (approved):
Thank you for pointing out my misunderstanding. Still no blockers IMO. See @darinkrauss re platform/errors usage, if my explanation isn't satisfactory.

ewollesen:
Oh, looks like there's a go import re-ordering that needs to be fixed to make the linter happy.

https://app.travis-ci.com/github/tidepool-org/platform/builds/272549249#L442

ehr/reconcile/planner.go
toddkazakov: I've been using the standard library most recently. This may be incorrect, but it's not obvious or documented why the platform errors package should be used instead, hence my question.

ewollesen (dismissed):

ehr/reconcile/planner.go
darinkrauss: Hmm, what do folks think about me writing a "Guide to Writing Consistent Code in the Platform Repo"? 😉 

Yes, there is a reason to use the Platform `errors` package and, more explicitly, the `errors.Wrap` functionality. The `errors` package adds a lot more context to the error that chains correctly with multiple wraps. Now, while the **current** Golang `errors` package does *some* of that, the rest of Platform isn't built to handle that in any way.

So, in short, if you want errors to propagate correctly and do the "right thing" in the rest of Platform, then use the Platform `errors` package and use `errors.Wrap` when you want to add further context or explanation to the error.

I'm fine if we want to intentionally revisit this and make appropriate changes, but IMO, we should stick with how it is currently done so it works and is written as expected.

My 2 cents...

toddkazakov: Thanks for the explanation - I'll make the change, because those errors might eventually bubble up to the task error handler.

ewollesen (approved):
LGTM

PR #770, opened 2024-09-24
rename some task packages

ewollesen:
See the individual commit messages for the details. I'm open to feedback if this causes anyone stress.


darinkrauss (changes requested):
Hmmm... What actual problem is this solving? 

Also, I don't recall a documented best practice about naming test-only packages. Can you point me to that?

Also, while `platform` may not conform to other **current** best practices, I don't see the point of make changes unless it has a functional reason. Plus, these type of isolated changes (i.e. just in these few locations) just make the whole repo inconsistent and very difficult for someone to know which is the standard way to do so for this repo.

If your primary concern is that you have to explicitly name the import, then just look at the rest of the platform imports for `platform` "standard" of using explicit import names to make it very clear what is being used.

ewollesen:
> Hmmm... What actual problem is this solving?

Changes don't have to solve a problem to be an improvement. Sometimes when I see a place that I think an improvement can be made, I try to make the improvement, in the hopes that it will make things better. I recognize these are subjective, and you might not agree.

> Also, I don't recall a documented best practice about naming test-only packages. Can you point me to that?

I don't believe I claimed such a thing exists. I see the pattern in the stdlib, and I think it's a good, clear, pattern, so I duplicated it, and mentioned that in my commit messages to indicate it was the "inspiration" if you will for the naming.
 
> Also, while `platform` may not conform to other **current** best practices, I don't see the point of make changes unless it has a functional reason. Plus, these type of isolated changes (i.e. just in these few locations) just make the whole repo inconsistent and very difficult for someone to know which is the standard way to do so for this repo.

I'm just coming from a place of trying to show an example of how I think the code could be improved. You too have suggested changes to code to improve its readability. Those aren't functional reasons, but they still matter, right? 

Making this kind of change throughout the entire code base all at once will be difficult, and not a great use of time, but not improving things where we can, means stagnation. Consistency is a useful feature, but not one worth preventing progress or improvement, IMO.

> If your primary concern is that you have to explicitly name the import, then just look at the rest of the platform imports for `platform` "standard" of using explicit import names to make it very clear what is being used.

When I see explicit names, I'm more likely to be confused:

"github.com/tidepool-org/platform/service" is imported bare, or as "platform" or as "router".

"github.com/tidepool-org/platform/service/service" is imported bare, or as "serviceService".

The import "serviceTest" could refer to any of "github.com/tidepool-org/platform/auth/service/test", "github.com/tidepool-org/platform/prescription/service/test", or "github.com/tidepool-org/platform/service/test". Some of those are also imported in a bare fashion.

Those were the first few places I looked. There's not much existing consistency that I can see, and when a name is explicitly named, that name is often neither clear, nor consistent. My point being, I don't think consistency is a good argument not to make some changes.

...and I'm not suggesting that we never explicitly name an import, but I'm suggesting that the changes in this PR can reduce some amount of confusion and improve the current situation. I'm a fan of small incremental improvements. Things grow and evolve and change over time, and I think that's a good thing.

darinkrauss:
Per our offline discussion, I'm good with the idea. Please propose the issue and solution to the team for general discussion.

PR #771, opened 2024-09-24
fix for a test that fails under `go test`

ewollesen:
There has been discussion of running our test suites via `go test` instead of `ginkgo` to get the package-level parallelism that `go test` offers (and caching!), without having to re-work all of our tests to make use of ginkgo v2's parallelism. To this end, the Makefile previously had a `go-test` recipe added. However, that recipe routinely fails one test in particular.

The test in question was found to be timing-out when run via `go test`. I suspect it's because under `go test` it's run in parallel, and isn't getting as much running time as a result. In any event, giving it more time allows the test to complete successfully (ironic, given that using `go test` is overall faster, at least for me).

I'm not sure that using `go test` is a long-term win over using ginkgo's parallelism though, as we can only access the package-level parallelism of `go test`, that is, we can't call `testing.T.Parallel()` to get test-level parallelism from `go test`, because Ginkgo doesn't support it, so we're back to re-working our tests to use Ginkgo's parallelism.

But... at least now we can run the tests both ways and compare.

### DIFF
--- a/task/queue/multi_test.go
+++ b/task/queue/multi_test.go
@@ -81,7 +81,6 @@ var _ = Describe("multi queue", func() {
 
 		It("Are partitioned correctly", func() {
 			ctx := log.NewContextWithLogger(context.Background(), lgr)
-			tasks := make([]*task.Task, 0, len(types)*tasksPerType)
 			creates := make([]*task.TaskCreate, 0, len(types)*tasksPerType)
 			runners := make([]*test.CountingRunner, 0, len(types))
 			now := time.Now()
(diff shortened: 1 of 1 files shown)


task/queue/multi_test.go
```diff
@@ -109,7 +108,6 @@ var _ = Describe("multi queue", func() {
 				tsk, err := str.NewTaskRepository().CreateTask(ctx, create)
 				Expect(err).ToNot(HaveOccurred())
 				Expect(tsk).ToNot(BeNil())
-				tasks = append(tasks, tsk)
```
ewollesen, line 112: My linter indicated that `tasks` isn't being used, hence I removed it.

darinkrauss (approved):
LGTM!

PR #772, opened 2024-09-26
re-use database connections between tests

ewollesen:
This results in massive speedups for the effected package (data/store/mongo). Compare for
`go clean -cache ; make ci-test TEST=data/store/mongo`:

    Ginkgo ran 1 suite in 4m51.037201582s

  -- vs --

    Ginkgo ran 1 suite in 38.723898055s

...for a speed up of 7.5x.

And for `go clean -cache ; make test TEST=data/store/mongo`:

    Ginkgo ran 1 suite in 4m41.735427524s

  -- vs --

    Ginkgo ran 1 suite in 24.292654343s

...this time the speedup is 11.5x.

The best part? This technique can be used in other tests within our code base for more gains.

### DIFF
--- a/data/store/mongo/mongo_test.go
+++ b/data/store/mongo/mongo_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"sync"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -32,7 +33,6 @@ import (
 	logTest "github.com/tidepool-org/platform/log/test"
 	"github.com/tidepool-org/platform/page"
 	"github.com/tidepool-org/platform/pointer"
-	storeStructuredMongo "github.com/tidepool-org/platform/store/structured/mongo"
 	storeStructuredMongoTest "github.com/tidepool-org/platform/store/structured/mongo/test"
 	"github.com/tidepool-org/platform/test"
 	userTest "github.com/tidepool-org/platform/user/test"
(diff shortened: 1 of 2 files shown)


data/store/mongo/mongo_test.go
```diff
@@ -237,35 +237,42 @@ func DataSetDatumAsInterface(dataSetDatum data.Datum) interface{} {
...
+				Expect(err).To(Succeed())
+			})
+		})
 	})
 
 	Context("New", func() {
 		It("returns an error if unsuccessful", func() {
-			var err error
-			store, err = dataStoreMongo.NewStore(nil)
+			store, err := dataStoreMongo.NewStore(nil)
```
darinkrauss, line 269: [nit] This is a little confusion with `store` matching the same name as the more "global" `store`. Consider using `testStore` as you did above. Same for the test below.

darinkrauss: Or, you could pull these two tests out of the BeforeEach/AfterEach Describe. (You'd need to create a new Context with for most everything and put these outside of that.)

data/store/mongo/mongo_test.go
```diff
@@ -276,26 +283,29 @@ var _ = Describe("Mongo", Label("mongodb", "slow", "integration"), func() {
...
-				collection.Database().Drop(context.Background())
-				dataSetCollection.Database().Drop(context.Background())
-				summaryCollection.Database().Drop(context.Background())
-				alertsCollection.Database().Drop(context.Background())
-			}
+			var err error
+			ctx := context.Background()
+			all := bson.D{}
+			_, err = collection.DeleteMany(ctx, all)
+			Expect(err).To(Succeed())
```
darinkrauss, line 302: [nit] I think you can rewrite this as:
```suggestion
            Expect(collection.DeleteMany(ctx, all)).Error().ShouldNot(HaveOccurred())
```

darinkrauss: And the ones below.

store/structured/mongo/test/test.go
```diff
@@ -18,6 +19,9 @@ import (
 var database string
 var address string
 
+var cleanupHandlers []func()
+var cleanupHandlersMu sync.Mutex
```
darinkrauss, line 23: [nit] `cleanupHandlersMutex` so it is clear?

store/structured/mongo/test/test.go
```diff
@@ -30,8 +34,32 @@ var _ = ginkgo.AfterSuite(func() {
...
+// DeferredCleanup provides a way to have something run during AfterSuite.
+//
+// Ginkgo only allows a single AfterSuite to be instantiated, and it's here in this file, so
+// give other tests a chance to make use of it.
+func DeferredCleanup(f func()) {
+	cleanupHandlersMu.Lock()
+	defer cleanupHandlersMu.Unlock()
+	if cleanupHandlers == nil {
+		cleanupHandlers = []func(){}
+	}
```
darinkrauss, line 49: I believe `append` will do this for you?

darinkrauss (changes requested):
One change to remove unnecessary code, plus a few nits for your consideration.

data/store/mongo/mongo_test.go
ewollesen: Will do.

store/structured/mongo/test/test.go
ewollesen: I believe the -Mu suffix is well established as indicating a sync.Mutex.

If I grep for `'Mu\s+sync.Mutex'` I get 41 hits.

If I grep for `'Mutex\s+sync.Mutex'` I get 3 hits.

The numbers are similar when searching the Go stdlib code (40 and 8).

store/structured/mongo/test/test.go
ewollesen: > I believe `append` will do this for you?

True, but this code was removed later, so it doesn't matter this time.

data/store/mongo/mongo_test.go
ewollesen: To be clear, this change will appear in #777 not here.

ewollesen:
All comments addressed in #777

PR #773, opened 2024-09-27
use a shared MongoDB client/store in data/summary tests

ewollesen:
A quick test showed an reduction in runtime from 1m34s to 4.7s.


ewollesen:

PR #774, opened 2024-09-27
minimize database timeouts for tests

ewollesen:
These were the two tests that took the longest to run.

Each test was waiting five seconds for a timeout that we fully intended to happen. By shortening the timeout, we don't need to wait as long.

I'm not sure why we're really testing this, because I don't see how MongoDB will somehow in the future connect to hosts that don't exist.

This chops 10 seconds off from each run.

### DIFF
--- a/store/structured/mongo/mongo_test.go
+++ b/store/structured/mongo/mongo_test.go
@@ -2,6 +2,7 @@ package mongo_test
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


ewollesen:
Built on top of #773, so wait for that to merge before rebasing.

ewollesen:

PR #775, opened 2024-09-27
use a shared MongoDB client/store in blob tests

ewollesen:
A quick test showed an reduction in runtime from 46s to 4.3s.


ewollesen:
Builds on #774, so should be rebased after that is merged.

ewollesen:

PR #776, opened 2024-09-27
use a shared MongoDB client/store in prescription tests

ewollesen:
A quick test showed an reduction in runtime from 15.6s to 4.8s.

This is the last one I'll be doing for a while...

At this stage, I can run `make go-test`, and have the entire suite finish in < 30 seconds. I can live with that for now. (GOTEST_FLAGS="-count=1 -race" make go-test takes ~40 seconds on my laptop).

### DIFF
--- a/prescription/store/mongo/prescription_repository_test.go
+++ b/prescription/store/mongo/prescription_repository_test.go
@@ -5,26 +5,13 @@ import (
 	"sort"
 	"time"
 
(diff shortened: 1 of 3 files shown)


ewollesen:
This builds on #775 so wait for that to be merged before rebasing this.

ewollesen:

PR #778, opened 2024-10-01
remove correlation with expectedExtended and expectedDuration

jh-bate:
expectedExtended and expectedDuration will be seen as independent params and  no correlation between the two for validation


darinkrauss (approved):
LGTM!

PR #779, opened 2024-10-01
adds go-ci-test Makefile target

ewollesen:
This should add suitable flags to approximate what we currently have with
ginkgo / ci-test.


darinkrauss (approved):
LGTM!

PR #780, opened 2024-10-04
remove deprecated service

ewollesen:
This service was deprecated in 2018. Let's finish the migration.

The data service is the only place that used the deprecated service. I started by expanding the data service to include both a deprecated service and the new service. Then when I had all the tests at equity, I dropped out the old deprecated service.

So all the things that were tested in the deprecated service are still tested (the only thing that wasn't already tested was part of the Authenticated service).

Before this is merged, I'd probably squash it done to a single commit. I'm leaving the separate commits in case someone wants to see the step by step of how I tried to make sure nothing was missed in the migration.


data/service/service/client_test.go
ewollesen, line 1: This did nothing, so I removed it. I guess it needed to be compiled and run by ginkgo, so... getting rid of it seems like a good thing.

That and the disappointment of reading the client code, then looking for the tests only to find that the file is empty. :(

service/service/authenticated_test.go
```diff
@@ -1,8 +1,89 @@
...
+		provider.AssertOutputsEmpty()
+	})
+
+	It("returns the secret", func() {
+		Expect(svc.Secret()).To(Equal(serviceConfig["secret"]))
+	})
+
+	Context("AuthClient", func() {
+		It("returns successfully with server token", func() {
+			//runtime.Breakpoint()
```
toddkazakov, line 83: This should be removed

toddkazakov (dismissed):

service/service/authenticated_test.go
ewollesen: Will do

toddkazakov (approved):

darinkrauss (dismissed):
Yep, LGTM. 

Hmm, I wonder why I didn't do the same thing way back then. 🤔 I guess we'll never know.

BTW, 👍 on the new tests for authenticated.

ewollesen:
> Service is failing on QA5. Please resolve issues before merging.

I believe this is resolved. See latest commit on the PR.

toddkazakov (approved):

PR #777, opened 2024-09-30
Eric db reuse all in one

ewollesen:
All my db PRs in one, as requested.

PR #781, opened 2024-10-04
run both ci-test and go-ci-test via Travis.

ewollesen:
This is intended as a trial to ensure that both make ci-test and make go-ci-test find the same errors. The latter is significantly faster, so it would be preferable to use it instead, but of course we need to ensure that we're not running things incorrectly or missing out on something provided by the current ginkgo-based solution.

For more info/discussion:
  - https://tidepoolteam.slack.com/archives/CF59ADWLB/p1727967169327199
  - https://github.com/tidepool-org/platform/pull/777

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,3 +41,8 @@ addons:
 
 script:
   - make ci-generate ci-build ci-test ci-deploy ci-docker
+  # Testing to see if go-ci-test catches the same failures as ci-test. If I
+  # understand how travis works, this means that go-ci-test will run even if
+  # ci-test fails, so we can see if they're catching the same
+  # things. Likewise, if go-ci-test fails, we can compare that to ci-test.
+  - make go-ci-test
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,9 +40,9 @@ addons:
     repo_token: 91ded9b66924acbe830541ab3593daf535f05f7c6db91b5cbd2d26dcf37da0b8
 
 script:
-  - make ci-generate ci-build ci-test ci-deploy ci-docker
+  - make ci-generate ci-build ci-test ci-deploy ci-docker TIMING_CMD="time -p"
   # Testing to see if go-ci-test catches the same failures as ci-test. If I
   # understand how travis works, this means that go-ci-test will run even if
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -41,3 +41,8 @@ addons:
 
 script:
   - make ci-generate ci-build ci-test ci-deploy ci-docker
+  # Testing to see if go-ci-test catches the same failures as ci-test. If I
+  # understand how travis works, this means that go-ci-test will run even if
+  # ci-test fails, so we can see if they're catching the same
+  # things. Likewise, if go-ci-test fails, we can compare that to ci-test.
+  - make go-ci-test
```
darinkrauss, line 48: Do we want to do this inline immediately after `ci-test` on the line above? The only reason I mention this is because as it is if `go-ci-test` fails, then the build will have failed, but it will still have produced docker images.

darinkrauss (commented):
One question.

.travis.yml
ewollesen: > Do we want to do this inline immediately after `ci-test` on the line above? The only reason I mention this is because as it is if `go-ci-test` fails, then the build will have failed, but it will still have produced docker images.

Yes, I thought about that, but since we consider ci-test the "official" source of a failure, that seemed to be the preferred outcome.

Also in that case, if ci-test fails, then go-ci-test isn't run, so we won't know if it "matched".

It's all a matter of "well, what do you want to do?", ya know? This is where I landed. 

I really don't think we'll have any trouble with one failing when the other doesn't though, do you?

darinkrauss: I just think it may be confusing where the build fails, but it gets deployed anyhow (assuming the feature branch was deployed to an environment).

If the team is okay with it, then I'm good.

darinkrauss (approved):
LGTM, but please get an okay from the rest of team regarding reporting a failure with `go test` even after building (and potentially deploying) the Docker images.

ewollesen:
Got OK from the team today at huddle. Todd indicated that it was always the case, that even if the tests failed, artifacts were still pushed, so there shouldn't be any major differences from how it used to work.

PR #783, opened 2024-10-21
[BACK-3097] Final index and changes from sharded cluster in prep for sharding.

lostlevels:
There are additional optimizations that can be made (~5% gains but that will require a rebuild of some of the indexes so that can be deferred until later).

This is a cleaned up verison of `jimmy-back-2416-shadowing-service-prefix` as that had experiimental indexes, and therefor experimental names that I cant just cherry pick onto here, as we can't rename an index in MDB.

This PR also removes the disable user events handler and direct connection to shoreline to keep things cleaner.


toddkazakov (approved):

Roukoswarf (approved):
LGTM

PR #785, opened 2024-10-29
BACK-3239] Fix Dexcom Connection task issue with incorrect id

darinkrauss:
- Fix Dexcom Connection task issue with incorrect id
- https://tidepool.atlassian.net/browse/BACK-3239


jh-bate (approved):
LGTM

PR #786, opened 2024-11-01
[BACK-3245] Cleanup device logs on context canceled and respond with appropriate error code when unable to retrieve device logs contents.

lostlevels:


blob/service/api/v1/v1.go
```diff
@@ -281,11 +281,11 @@ func (r *Router) GetDeviceLogsContent(res rest.ResponseWriter, req *rest.Request
 
 	content, err := blobClient.GetDeviceLogsContent(req.Context(), *deviceLogMetadata.ID)
 	if err != nil {
-		responder.Error(http.StatusInternalServerError, err)
+		responder.CodedError(err)
 		return
 	}
 	if content == nil || content.Body == nil {
-		responder.Error(http.StatusNotFound, request.ErrorResourceNotFoundWithID(deviceLogID))
+		responder.CodedError(request.ErrorResourceNotFoundWithID(deviceLogID))
```
toddkazakov, line 288: Create and error first and then use `responder.RespondIfError()`

toddkazakov (changes requested):
See the comments I left inline

toddkazakov (approved):

PR #788, opened 2024-11-21
Makefile changes to support private repository builds

toddkazakov:
- Conditionally add `-private` suffix to repository name by passing `$PRIVATE=true` environment variable
- Allow specifying a subset of services to be built (we'll only need to build `data` and `auth` services in the private repository

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,15 @@ TIMESTAMP ?= $(shell date +%s)
 # these can vary by 1 second
 export TIMESTAMP
 
+ifeq ($(PRIVATE),true)
+  REPOSITORY_SUFFIX:=-private
+endif
+
 ROOT_DIRECTORY:=$(realpath $(dir $(realpath $(lastword $(MAKEFILE_LIST)))))
 
 REPOSITORY_GOPATH:=$(word 1, $(subst :, ,$(GOPATH)))
 REPOSITORY_PACKAGE:=github.com/tidepool-org/platform
-REPOSITORY_NAME:=$(notdir $(REPOSITORY_PACKAGE))
+REPOSITORY_NAME:=$(notdir $(REPOSITORY_PACKAGE))$(REPOSITORY_SUFFIX)
 
 BIN_DIRECTORY := ${ROOT_DIRECTORY}/_bin
 PATH := ${PATH}:${BIN_DIRECTORY}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -3,11 +3,19 @@ TIMESTAMP ?= $(shell date +%s)
...
 
+ifeq ($(PRIVATE),true)
+  REPOSITORY_SUFFIX:=-private
+endif
+
+SERVICES_SEPARATOR=,
+BUILD_SERVICES?=auth,blob,data,migrations,prescription,task,tools
+BUILD_SERVICES:=$(subst $(SERVICES_SEPARATOR), ,$(BUILD_SERVICES))
+
 ROOT_DIRECTORY:=$(realpath $(dir $(realpath $(lastword $(MAKEFILE_LIST)))))
```
ewollesen, line 14: Is this not just $(CURDIR), which is automatically set by make.

Oh, this isn't a modified line, oh well, ignore.

Makefile
```diff
@@ -286,11 +299,15 @@ endif
...
 endif
+else
+	@echo skipping $(DOCKER_FILE)
+endif
 endif
 endif
 
 docker-push:
 ifdef DOCKER
+ifdef BUILD_SERVICE:
```
ewollesen, line 310:
```suggestion
ifdef BUILD_SERVICE
```
Was the `:` a typo? But if it was a typo, it shouldn't have ever matched, and so should have never worked?

ewollesen (dismissed):
I think this works, but I have some suggestions and a PR.

Makefile
```diff
@@ -3,11 +3,19 @@ TIMESTAMP ?= $(shell date +%s)
 # these can vary by 1 second
 export TIMESTAMP
 
+ifeq ($(PRIVATE),true)
+  REPOSITORY_SUFFIX:=-private
+endif
+
+SERVICES_SEPARATOR=,
+BUILD_SERVICES?=auth,blob,data,migrations,prescription,task,tools
+BUILD_SERVICES:=$(subst $(SERVICES_SEPARATOR), ,$(BUILD_SERVICES))
```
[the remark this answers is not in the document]
toddkazakov: So you can pass a comma separated list in an env var

Makefile
toddkazakov: Definitely typo, I never tested the push

PR #789, opened 2024-11-21
tweak how private is handled

ewollesen:
This uses the PRIVATE variable a little closer to what
matters. Previously, one had to follow the PRIVATE variable to the
setting of REPOSITORY_SUFFIX, then follow that its use in setting
REPOSITORY_NAME, then trace that all around to understand the effect
of PRIVATE.

This way, we read the value of PRIVATE, and use it all in the same
place, and we don't need REPOSITORY_SUFFIX at all.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,6 @@ TIMESTAMP ?= $(shell date +%s)
 # these can vary by 1 second
 export TIMESTAMP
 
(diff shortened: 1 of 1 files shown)


ewollesen:
Discussed with Todd while this isn't a good idea.

PR #790, opened 2024-11-25
fixes a few flappy tests

ewollesen:
A few of these tests assumed that their generated keys were unique, when they weren't, which lead to flappy tests.

I'm not sure I've found them all, but I was able to run them with --repeat 20000 without a failure.


ewollesen:
Here's one example of a flappy failure: https://app.travis-ci.com/github/tidepool-org/platform/builds/272827297#L2284

darinkrauss (approved):
LGTM

PR #792, opened 2024-11-27
[BACK-3055] Decouple RPM reports from clinic client schema

Roukoswarf:


clinics/service.go
```diff
@@ -165,8 +166,25 @@ func (d *defaultClient) getPatient(ctx context.Context, clinicID, patientID stri
 	return response.JSON200, nil
 }
 
-func (d *defaultClient) GetPatients(ctx context.Context, clinicId string, userToken string, params *clinic.ListPatientsParams) ([]clinic.Patient, error) {
+func (d *defaultClient) GetPatients(ctx context.Context, clinicId string, userToken string, params *clinic.ListPatientsParams, injectedParams map[string]any) ([]clinic.Patient, error) {
 	response, err := d.httpClient.ListPatientsWithResponse(ctx, clinicId, params, func(ctx context.Context, req *http.Request) error {
+		if len(injectedParams) != 0 {
```
toddkazakov, line 171: This block doesn't belong to this function. This function is generic and may be used by multiple callers. This block converts a specific shape only known to the caller to raw query params. Instead, the parameters here should be `rawParams map[string][]string` and this conversion should be done by the caller.

Roukoswarf: raw params in this case will be `map[string]string`, array fields in the clinic client are comma separated, not multi-used, so strings would map most generic.

toddkazakov (changes requested):

toddkazakov (approved):

PR #794, opened 2024-12-11
rearrange the test target to point to go-test

ewollesen:
This is to parallel the CI server, which is using go-ci-test.

The old test target is now ginkgo-test.


darinkrauss (approved):
LGTM!

PR #804, opened 2025-01-13
Summary Queue Prometheus Metrics

Roukoswarf:


data/summary/store/store.go
```diff
@@ -260,6 +284,13 @@ func (r *Repo[T, A]) GetOutdatedUserIDs(ctx context.Context, page *page.Paginati
 		response.End = *userSummary.Dates.OutdatedSince
 	}
 
+	QueueLag.WithLabelValues(typ).Observe(time.Now().UTC().Sub(response.Start).Minutes())
```
toddkazakov, line 287: The duration ought to be measured in seconds and combined with the appropriate buckets configuration of the histogram.

data/summary/store/store.go
```diff
@@ -15,6 +18,17 @@ import (
 	storeStructuredMongo "github.com/tidepool-org/platform/store/structured/mongo"
 )
 
+var (
+	QueueLag = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
toddkazakov, line 22: You ought to pass buckets here, because the defaults are not suitable. From the docs:
```
var DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
```

> DefBuckets are the default Histogram buckets. The default buckets are tailored to broadly measure the response time (in seconds) of a network service. Most likely, however, you will be required to define buckets customized to your use case.

Perhaps this is a good starting point:
```
[]float64{0.5, 1, 2.5,  5, 7.5, 10, 25, 50, 75, 100, 150, 250, 500, 1000}
```

toddkazakov (approved):

toddkazakov (changes requested):
You need to expose the metrics endpoint in the data service similar to how it's done in the task service https://github.com/tidepool-org/platform/blob/master/task/service/api/v1/v1.go#L40-L48

toddkazakov (approved):

PR #809, opened 2025-01-16
change egrep to grep -E

ewollesen:
egrep is deprecated now, at least on my Linux box, so this was
spitting out tons of warning messages, which this change stops.


tjotala (approved):
LGTM :shipit:

PR #810, opened 2025-01-22
Disable modal close on oauth connection redirect

clintonium-119:
Serves https://tidepool.atlassian.net/browse/WEB-3275

With the new data source connection UI, we no longer will need to force-close the popup window.  

Rather, we can react to the normal redirect URL, which will allow us to first detect the end result and then close it while setting an appropriate toast message.


jh-bate (approved):
LGTM

PR #814, opened 2025-01-29
upgrade golang.org/x/net package to fix CVE-2024-45338

ewollesen:
My investigation indicates that neither our code, nor our dependencies use the functions in question, but at the same time, the impact of upgrading them is minimal, so it feels like the less risky path to just upgrade and not have to worry.

BACK-3353


darinkrauss (approved):
LGTM!

ewollesen:
updated to include latest go-common

Roukoswarf (approved):

darinkrauss (approved):
LGTM!

PR #815, opened 2025-01-30
allow an access token without a user id to authenticate as a service

ewollesen:
"MethodAccessToken" is a valid way for a service to authenticate, and
could happen for example, when someone uses a oauth library to perform
authentication, rather than using the internal platform client.


darinkrauss (approved):
LGTM!

PR #817, opened 2025-02-07
[BACK-3392] Add OpenID sector for Abbott OAuth

darinkrauss:
Note: This implementation is a temporary placeholder to allow bootstrapping of the Abbott OAuth client workflow. Will need to migrate this to environment variables and add minimal authorization. For now, though, this is acceptable since it isn't revealing anything that is not already available in other locations (i.e. other public repos).

TODO: https://tidepool.atlassian.net/browse/BACK-3394

### DIFF
--- a/data/service/api/v1/v1.go
+++ b/data/service/api/v1/v1.go
@@ -20,6 +20,8 @@ func Routes() []service.Route {
 		service.Put("/v1/data_sets/:dataSetId", DataSetsUpdate, api.RequireAuth),
 		service.Get("/v1/time", TimeGet),
 		service.Post("/v1/users/:userId/data_sets", UsersDataSetsCreate, api.RequireAuth),
+
+		service.Get("/v1/partners/:partner/sector", PartnersSector),
 	}
 
 	routes = append(routes, DataSetsRoutes()...)
--- a/data/service/api/v1/partners.go
+++ b/data/service/api/v1/partners.go
@@ -1,0 +1,46 @@
+package v1
+
+import (
+	"net/http"
+	"os"
+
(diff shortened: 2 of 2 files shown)


ewollesen (approved):
That's a lot of URLs / data to put in an env variable. I wonder if it shouldn't be a in a file, eventually that is.

As a temporary measure, it seems fine though.

darinkrauss:
Thanks. Where would we put the file? I'd like to put it in the infrastructure somewhere so it can be easily configured per environment. I'm probably missing something obvious.

data/service/api/v1/v1.go
```diff
@@ -20,6 +20,8 @@ func Routes() []service.Route {
 		service.Put("/v1/data_sets/:dataSetId", DataSetsUpdate, api.RequireAuth),
 		service.Get("/v1/time", TimeGet),
 		service.Post("/v1/users/:userId/data_sets", UsersDataSetsCreate, api.RequireAuth),
+
+		service.Get("/v1/partners/:partner/sector", PartnersSector),
```
toddkazakov, line 24: Given that in the OIDC this is referred to `Sector Identifier URI`, should this be called `sector_identifier`?

darinkrauss: I could go either way. The way I read the docs was as the sector's identifier url, not the sector identifier's url. Let me know which way you'd like to go.

jh-bate (approved):
LGTM

data/service/api/v1/v1.go
toddkazakov: I think you're right

toddkazakov (approved):

PR #820, opened 2025-03-07
strip kc: token if present before parsing token claims

ewollesen:
Intended to prevent egregious log entries.


ewollesen:
Hmm... having re-read Todd's slack post, I might need to split on ":", strip the kc and refresh token (if any) before continuing...

ewollesen:
Closing in favor of #821

PR #822, opened 2025-03-12
Make trend rate units optional

toddkazakov:
DIY Loop doesn't send trend rate units which prevents CBG data from being uploaded due to too stringent validation. We could potentially try to infer trend rate units from the blood glucose units, but that doesn't seem ideal.


darinkrauss (dismissed):
LGTM!

tjotala (approved):
LGTM

PR #828, opened 2025-03-23
Add client id to OAuth provider and fix authentication issue

darinkrauss:
- Add client id to OAuth provider
- Fix authentication issue with non-Bearer Authorization header

Counterpart to https://github.com/tidepool-org/platform-plugin-abbott/pull/20.


jh-bate (approved):
LGMT

PR #831, opened 2025-03-27
BACK-3471 summary stats

jh-bate:
see https://github.com/tidepool-org/platform-plugin-abbott/pull/24

### DIFF
--- a/data/service/service/standard.go
+++ b/data/service/service/standard.go
@@ -541,6 +541,7 @@ func (s *Standard) initializeWorkCoordinator() error {
 		DataDeduplicatorFactory: s.dataDeduplicatorFactory,
 		DataSetClient:           s.dataClient,
 		DataSourceClient:        s.dataSourceStructuredStore.NewDataSourcesRepository(),
+		DataStoreClient:         s.dataStore,
 		ProviderSessionClient:   s.AuthClient(),
 		DataRawClient:           s.dataRawClient,
 		AbbottClient:            s.abbottClient,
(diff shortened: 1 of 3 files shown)


data/service/service/standard.go
```diff
@@ -477,6 +483,18 @@ func (s *Standard) initializeDataSourceClient() error {
 	return nil
 }
 
+func (s *Standard) initializeSummaryClient() error {
+	s.Logger().Debug("Creating summary client")
+
+	clnt, err := dataSummaryClient.New(s.dataStore)
```
darinkrauss, line 489: What do you think about injecting the summarizer registry into the summary client (instead of the store, which is a fairly large dependency)? Something like:
```suggestion
	s.Logger().Debug("Creating summarizer registry")

	summarizerRegistry := dataSummary.New(
		s.dataStore.NewSummaryRepository().GetStore(),
		s.dataStore.NewDataRepository(),
	)

	s.Logger().Debug("Creating summary client")

	clnt, err := dataSummaryClient.New(summarizerRegistry)
```
Seems like this would be better for testing.

darinkrauss (approved):
One comment/suggestion to think about regarding dependency injection into the summary client. It's fine as-is, but I think it would be easier for testing with only injecting the summarizer registry (rather than the entire data store).

@Roukoswarf - Is the summarizer registry stateless or stateful? That is, does a new summarizer registry need to be created for every user request to ingest data? 

@jh-bate Approved, pending your thoughts on my dependency injection comment and assuming the summary registry is stateless per Alex (i.e. we can create one client at the beginning of time and use it for all ingestion).

PR #836, opened 2025-05-01
[BACK-3605] [BACK-3656] testing of the work system 

jh-bate:
just for initial exploration of test the work system and also load tests and mechanism to drive it


work/load/load_processors.go
```diff
@@ -0,0 +1,175 @@
...
+			if resultStr == "random" {
+				var indexnum int = rand.IntN(len(possibleWorkResults) - 1)
+				return possibleWorkResults[indexnum]
+			}
+			if slices.Contains(possibleWorkResults, resultStr) {
+				return resultStr
+			}
+		}
+	}
+	return work.ResultDelete
```
darinkrauss, line 104: Rather than defaulting to delete, you could default to random. Then no need for the "random" check above. If `MetadataProcessResult` isn't specified then that means random.

work/load/load_processors.go
```diff
@@ -0,0 +1,175 @@
...
+		return work.NewProcessResultSuccess(work.SuccessUpdate{
+			Metadata: wrk.Metadata,
+		})
+	default:
+		return work.NewProcessResultDelete()
+	}
+}
+
+func (p *loadProcessor) chooseDopeyResult(wrk *work.Work) string {
+	possibleWorkResults := []string{
```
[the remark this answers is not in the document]
darinkrauss: Also, note that `failing` is very similar to `pending` in that it will keep repeating forever until either a failed/success is returned.

work/load/load_processors.go
```diff
@@ -0,0 +1,175 @@
...
+			msg = *errMsg
+		}
+
+		return work.NewProcessResultFailing(work.FailingUpdate{
+			FailingError: *errors.NewSerializable(errors.New(msg)),
+			Metadata:     wrk.Metadata,
+		})
+	case work.ResultFailed:
+		return work.NewProcessResultFailed(work.FailedUpdate{
+			FailedError: *errors.NewSerializable(errors.New("failed from load test")),
```
darinkrauss, line 74: Eventually, the error message should be specified in the metadata.

work/load/load_processors.go
```diff
@@ -0,0 +1,175 @@
...
+func (p *loadProcessor) getProcessResult(wrk *work.Work, result any, errMsg *string) *work.ProcessResult {
+	switch result.(string) {
+	case work.ResultFailing:
+
+		msg := "failing from load test"
+		if errMsg != nil {
+			msg = *errMsg
+		}
+
+		return work.NewProcessResultFailing(work.FailingUpdate{
```
darinkrauss, line 68: You will likely want to add a error and retry time to this. Also, I'd recommend incrementing the retry count each time.

work/load/load_processors.go
```diff
@@ -0,0 +1,175 @@
...
+		return *p.getProcessResult(wrk, work.ResultSuccess, nil)
+	case TypeDopey:
+		return *p.getProcessResult(wrk, p.chooseDopeyResult(wrk), nil)
+	default:
+		return *p.getProcessResult(wrk, work.ResultFailed, pointer.FromString(fmt.Sprintf("[%s] not an expected work type", wrk.Type)))
+	}
+}
+
+type LoadItem struct {
+	SecondsOffsetFromStart int64        `json:"secondsOffsetFromStart"`
```
darinkrauss, line 121: Milliseconds?

work/load/load_processors.go
```diff
@@ -0,0 +1,175 @@
...
+
+func NewLoadWorkCreate(create *work.Create) (*work.Create, error) {
+	if create == nil {
+		return nil, errors.New("create is missing")
+	}
+	if create.Type != TypeSleepy && create.Type != TypeDopey {
+		return nil, fmt.Errorf("invalid work type %s", create.Type)
+	}
+
+	availableTime := time.Now().Add(processingAvailableTimeDelay)
```
darinkrauss, line 133: I would avoid a default here and require the `create.ProcessingAvailableTime`.

work/load/load_processors.go
```diff
@@ -0,0 +1,175 @@
...
+		timeout = create.ProcessingTimeout
+	}
+
+	metadata := map[string]any{}
+
+	if create.Metadata != nil && create.Metadata[MetadataProcessResult] != nil {
+		metadata[MetadataProcessResult] = create.Metadata[MetadataProcessResult]
+	}
+
+	return &work.Create{
```
darinkrauss, line 148: Why aren't you just straight-away using the incoming `create` directly? The caller should be setting the `ProcessingAvailableTime` based upon the `SecondsOffsetFromStart`. I suppose if the ProcessingTime is 0, then you could just change it to the default, which rather that 5 minutes, I'd make infinite. We want to see if we can get it stuck.

work/service/work_handlers.go
```diff
@@ -0,0 +1,85 @@
+package service
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	dataService "github.com/tidepool-org/platform/data/service"
+	request "github.com/tidepool-org/platform/request"
```
darinkrauss, line 9: You can remove the `request` alias as it is not needed.

work/service/work_handlers.go
```diff
@@ -0,0 +1,85 @@
...
+	workLoad "github.com/tidepool-org/platform/work/load"
+)
+
+func LoadTestRoutes() []dataService.Route {
+	return []dataService.Route{
+		dataService.Get("/v1/work/load/ok", okHandler),
```
darinkrauss, line 15: I would use `/v1/work/load/status` as it is more intuitive since you are returning the status.

work/service/work_handlers.go
```diff
@@ -0,0 +1,85 @@
...
+	res := dataServiceContext.Response()
+	req := dataServiceContext.Request()
+	responder := request.MustNewResponder(res, req)
+	responder.Data(http.StatusOK, map[string]any{"status": "All good ready for work load"})
+}
+
+func deleteHandler(dataServiceContext dataService.Context) {
+	ctx := dataServiceContext.Request().Context()
+	res := dataServiceContext.Response()
+	req := dataServiceContext.Request()
```
darinkrauss, line 31:
```suggestion
	res := dataServiceContext.Response()
	req := dataServiceContext.Request()
	ctx := req.Context()
```

work/service/work_handlers.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	_, err := dataServiceContext.WorkClient().DeleteAllByGroupID(ctx, groupID)
+
+	if err != nil {
+		responder.Error(http.StatusInternalServerError, err)
+		return
+	}
+
+	responder.Empty(http.StatusNoContent)
```
darinkrauss, line 48: 200 is more typical for a delete operation. 204 is more for a get where this isn't anything to get.

work/load/load_processors.go
darinkrauss: In fact, why do we even need this function at all? Couldn't we send the `create` from the parsed JSON directly into the work system?

work/service/work_handlers.go
```diff
@@ -0,0 +1,85 @@
...
+	for _, item := range items {
+		item.Create.ProcessingAvailableTime = start.Add(time.Second * time.Duration(item.SecondsOffsetFromStart))
+
+		wc, err := workLoad.NewLoadWorkCreate(item.Create)
+		if err != nil {
+			responder.Error(http.StatusBadRequest, fmt.Errorf("error creating work.Create %s", err.Error()))
+			return
+		}
+		if wrk, err := dataServiceContext.WorkClient().Create(ctx, wc); err != nil {
+			responder.Error(http.StatusBadRequest, fmt.Errorf("error creating work.Work %s", err.Error()))
```
darinkrauss, line 77: What happens with any previously created work? Say, the 3rd out of 4 work item fails to created. What will happen to the 1st and 2nd?

work/test/load/cli/cli.go
```diff
@@ -0,0 +1,239 @@
...
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"slices"
+	"time"
+
+	"github.com/tidepool-org/platform/pointer"
+	"github.com/tidepool-org/platform/work"
+	"github.com/tidepool-org/platform/work/load"
```
darinkrauss, line 17: `workLoad` alias?

work/test/load/cli/cli.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/urfave/cli/v3"
```
darinkrauss, line 18: In separate import block above the local `platform` import block?

work/work_system_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"go.uber.org/mock/gomock"
+
+	authTest "github.com/tidepool-org/platform/auth/test"
+	"github.com/tidepool-org/platform/work"
+	workTest "github.com/tidepool-org/platform/work/test"
+	workLoadTest "github.com/tidepool-org/platform/work/test/load"
```
darinkrauss, line 13: `workTestLoad`?

work/test/load/coordinator_client.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	c.coordinator.Start()
+	c.runFilePath = runFilePath
+
+	jsonFile, err := os.Open(runFilePath)
+	if err != nil {
+		return errors.Wrapf(err, "unable open run file %s", runFilePath)
+	}
+	defer jsonFile.Close()
+	jsonData, err := io.ReadAll(jsonFile)
```
darinkrauss, line 62:
```suggestion
	jsonData, err := os.ReadFile(runFilePath)
```

work/test/load/coordinator_client.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "read file %s", runFilePath)
+	}
+	var allData []workLoad.LoadItem
+	json.Unmarshal(jsonData, &allData)
```
darinkrauss, line 67: Check error here?

work/test/load/coordinator_client.go
```diff
@@ -0,0 +1,96 @@
...
+
+	c.runStart = time.Now()
+	for _, data := range allData {
+		data.Create.ProcessingAvailableTime = c.runStart.Add(time.Second * time.Duration(data.SecondsOffsetFromStart))
+		workItem, err := c.workClient.Create(ctx, data.Create)
+		if err != nil {
+			return errors.Wrapf(err, "error creating work %v", data.Create)
+		}
+		c.groupWorkItems[*data.Create.GroupID] = append(c.groupWorkItems[*data.Create.GroupID], workItem)
```
darinkrauss, line 76: What happens if no group id?

work/test/load/coordinator_client.go
```diff
@@ -0,0 +1,96 @@
...
+	"context"
+	"encoding/json"
+	"io"
+	"log"
+	"os"
+	"time"
+
+	"github.com/tidepool-org/platform/auth"
+	"github.com/tidepool-org/platform/errors"
+	"github.com/tidepool-org/platform/log/null"
```
darinkrauss, line 13: `logNull` alias?

work/test/load/cli/cli.go
```diff
@@ -0,0 +1,239 @@
...
+
+					req, err := http.NewRequest("GET", url, nil)
+					if err != nil {
+						return fmt.Errorf("error checking %s", err.Error())
+					}
+
+					res, err := http.DefaultClient.Do(req)
+					if err != nil {
+						return fmt.Errorf("unable to issue work load test API request: %s", err.Error())
+					}
```
darinkrauss, line 216: Consider:
```suggestion
					res, err := http.Get(fmt.Sprintf("%s/v1/work/load/ok", urlBase))
					if err != nil {
						return fmt.Errorf("unable to issue work load test API request: %s", err.Error())
					}
```

work/test/load/cli/cli.go
```diff
@@ -0,0 +1,239 @@
...
+					url := fmt.Sprintf("%s/v1/work/load", urlBase)
+
+					req, err := http.NewRequest("POST", url, bytes.NewReader(testData))
+					if err != nil {
+						return fmt.Errorf("error sending data to load work %s", err.Error())
+					}
+					req.Header.Add("content-type", "application/json")
+					req.Header.Add("accept", "application/json")
+
+					res, err := http.DefaultClient.Do(req)
```
darinkrauss, line 95: Consider:
```suggestion
					testData, err := os.Open(filePath)
					if err != nil {
						return fmt.Errorf("error opening %s %s", filePath, err.Error())
					}
                    defer testData.Close()
                    
					res, err := http.Post(fmt.Sprintf("%s/v1/work/load", urlBase), "application/json", testData)
```

work/test/load/cli/cli.go
```diff
@@ -0,0 +1,239 @@
...
+					if err != nil {
+						return fmt.Errorf("unable to issue work load test API request: %s", err.Error())
```
darinkrauss, line 97: For all of the code where you use `fmt.Errorf("... %s", err.Error())` consider instead `fmt.Errorf("... %w", err)`.

darinkrauss: Note using `%w` for the `err`.

darinkrauss (commented):
A bunch of comments for your consideration.

PR #844, opened 2025-05-22
[BACK-3735] Update dosing decision recommended and requested bolus to allow extended boluses

darinkrauss:
- Update dosing decision recommended and requested bolus to allow extended boluses
- Refactor separate dosing decision bolus structs into common struct
- Add duration, extended, and normal properties to dosing decision bolus
- Deprecate amount in dosing decision bolus
- https://tidepool.atlassian.net/browse/BACK-3735


jh-bate (dismissed):
LGTM

darinkrauss:
@jh-bate Can I get a re-review of the latest commit? The amount is no longer deprecated. It is necessary in certain use cases. Namely, when the recommended bolus is generalized amount where the requested bolus can be split into normal and extended.

toddkazakov (approved):

PR #846, opened 2025-05-22
[BACK-3394] Update Abbott sector identifiers

darinkrauss:
- Update Abbott sector identifiers
- https://tidepool.atlassian.net/browse/BACK-3394


jh-bate (approved):
LGTM

PR #850, opened 2025-05-23
[BACK-3727] Refactor data client and repository CreateUserDataSet, GetDataSet, and UpdateDataSet

darinkrauss:
- Refactor data client and repository CreateUserDataSet, GetDataSet, and UpdateDataSet
- Add CreateUserDataSet with user id and data.DataSetCreate
- Replace DataSetRepository GetDataSetByID with GetDataSet
- Connect data service Client UpdateDataSet to repository
- https://tidepool.atlassian.net/browse/BACK-3727

### DIFF
--- a/data/store/mongo/mongo_data_set.go
+++ b/data/store/mongo/mongo_data_set.go
@@ -96,40 +96,12 @@ func (d *DataSetRepository) EnsureIndexes() error {
 	})
 }
 
-func (d *DataSetRepository) GetDataSetByID(ctx context.Context, dataSetID string) (*data.DataSet, error) {
+func (d *DataSetRepository) createDataSet(ctx context.Context, dataSet *data.DataSet, now time.Time) (*data.DataSet, error) {
 	if ctx == nil {
 		return nil, errors.New("context is missing")
 	}
-	if dataSetID == "" {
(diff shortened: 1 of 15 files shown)


data/store/mongo/mongo_data_set.go
```diff
@@ -296,6 +269,30 @@ func (d *DataSetRepository) ListUserDataSets(ctx context.Context, userID string,
 	return dataSets, nil
 }
 
+func (d *DataSetRepository) CreateUserDataSet(ctx context.Context, userID string, create *data.DataSetCreate) (*data.DataSet, error) {
+	dataSet := data.NewDataSet()
+	dataSet.Client = create.Client
```
toddkazakov, line 274: Nit: A common pattern for platform services is to have a constructor which sets all fields of a struct given a "create" struct. Is there a reason why this logic is in the store and not in a separate constructor?

jh-bate (approved):
LGTM

data/store/mongo/mongo_data_set.go
darinkrauss: I looked at the platform code and it is a bit hit-or-miss for which way it is done. For example, the task store does it this way.

PR #852, opened 2025-05-29
[BACK-3750] Increase maximum volume for prime device event data

darinkrauss:
- Increase maximum volume for prime device event data
- Update tests and imports
- https://tidepool.atlassian.net/browse/BACK-3750


jh-bate (approved):
LGTM

PR #859, opened 2025-06-04
[BACK-3731] Fix variety of minor issues

darinkrauss:
- Fix variety of minor issues
- Promote refresh token logic from dexcom client to oauth client package
- Refactor pointer Default functions to return value, not pointer
- Add client Config.LoadFromConfigReporter helper function
- Add config Reporter debug functionality
- Use string Mutator not string array Mutator for DeviceLogsFilter
- Add request.HeadersMutator
- Remove unnecessary logger from request.HeadersInspector
- Move DebuggingTransport into net package
- Rename palmtree_secrets.go to palm_tree_secrets.go for consistency
- Use http.DefaultClient rather than creating a new http.Client
- Update service middleware authentication to allow Basic authentication
- Add structured store Condition with revision
- Allow more flexible structure parsing of time values
- Use RAM disk for Mongo in Travis to speed up builds
- Add and update tests
- https://tidepool.atlassian.net/browse/BACK-3731


structure/parser/object_parser.go
```diff
@@ -202,28 +202,30 @@ func (o *Object) StringArray(reference string) *[]string {
...
-	switch rawValue.(type) {
-	case time.Time:
-		timeValue = rawValue.(time.Time)
-	default:
-		stringValue, ok := rawValue.(string)
-		if !ok {
-			o.base.WithReference(reference).ReportError(ErrorTypeNotTime(rawValue))
-			return nil
-		}
+	timeValue, timeValueOk := rawValue.(time.Time)
```
jh-bate, line 210: should the object and array parser just use the same time parsing logic?

jh-bate (approved):
LGTM - just a couple of questions but nothing that needs addressing

structure/parser/object_parser.go
darinkrauss: Yea, probably. Perhaps with a common parsing implementation between the two, but that is a wider ranging change.

PR #867, opened 2025-06-09
use a default timeout on requests that don't specify one

ewollesen:
This behavior was inadvertently lost in
e282fb3cfb0004c9a52e014003a00c42b9fd462e, but is restored here, while
still allowing for per-request timeouts to be specified.


darinkrauss (commented):
I think we can probably abandoned this PR.

PR #871, opened 2025-06-20
round glucose values via "round-to-even" before classifying

ewollesen:
The aim is to maintain consistency with the front end, and be easy to
explain when values don't align with other interpretations due to
rounding differences.

BACK-3800

### DIFF
--- a/data/types/blood/glucose/glucose.go
+++ b/data/types/blood/glucose/glucose.go
@@ -1,7 +1,9 @@
 package glucose
 
 import (
+	"cmp"
 	"math"
+	"slices"
 
 	"github.com/tidepool-org/platform/data"
 	dataBloodGlucose "github.com/tidepool-org/platform/data/blood/glucose"
@@ -78,6 +80,7 @@ func (c Classifier) Classify(g *Glucose) (Classification, error) {
 	// inconsistency between frontend, backend, and/or other reports. See BACK-3800 for
 	// details.
 	rounded := roundToEvenWithDecimalPlaces(normalized, 1)
+	sortThresholds(c)
 	for _, threshold := range c {
 		if threshold.Includes(rounded) {
 			return threshold.Name, nil
@@ -85,6 +88,21 @@ func (c Classifier) Classify(g *Glucose) (Classification, error) {
 	}
 	// Ensure your highest threshold has a value like math.MaxFloat64 to avoid this.
 	return ClassificationInvalid, errors.Newf("unable to classify value: %v", *g)
+}
+
+func sortThresholds(ts []classificationThreshold) {
+	slices.SortFunc(ts, func(i, j classificationThreshold) int {
+		if valueCmp := cmp.Compare(i.Value, j.Value); valueCmp != 0 {
+			return valueCmp
+		}
+		if !i.Inclusive && j.Inclusive {
+			return -1
+		} else if i.Inclusive == j.Inclusive {
+			return 0
+		} else {
+			return 1
+		}
+	})
 }
 
 // Config helps summaries report the configured thresholds.
@@ -104,9 +122,6 @@ func (c Classifier) Config() map[Classification]float64 {
 //
 // In addition to the standard ADA ranges, the Tidepool-specifiic "extremely high" range is
 // added.
-//
-// It is the author's responsibility to ensure the thresholds remain sorted from smallest to
-// largest.
 var TidepoolADAClassificationThresholdsMmolL = Classifier([]classificationThreshold{
 	{Name: VeryLow, Value: 3, Inclusive: false},                    // Source: https://doi.org/10.2337/dc24-S006
 	{Name: Low, Value: 3.9, Inclusive: false},                      // Source: https://doi.org/10.2337/dc24-S006


summary/types/glucose.go
```diff
@@ -162,31 +162,38 @@ func (rs *GlucoseRanges) Finalize(days int) {
 	}
 }
 
-func (rs *GlucoseRanges) Update(record *glucoseDatum.Glucose) {
-	normalizedValue := *glucose.NormalizeValueForUnits(record.Value, record.Units)
+func (rs *GlucoseRanges) Update(record *glucoseDatum.Glucose) error {
+	classification, err := glucoseDatum.TidepoolADAClassificationThresholdsMmolL.Classify(record)
```
toddkazakov, line 166: I think you. need to replace the ["config" values in the summary](https://github.com/tidepool-org/platform/blob/master/summary/types/summary.go#L23-L27) with the new thresholds

data/types/blood/glucose/glucose.go
```diff
@@ -33,3 +36,75 @@ func (g *Glucose) Normalize(normalizer data.Normalizer) {
...
+
+func (c Classifier) Classify(g *Glucose) (Classification, error) {
+	normalized, err := dataBloodGlucose.NormalizeValueForUnitsSafer(g.Value, g.Units)
+	if err != nil {
+		return ClassificationInvalid, errors.Wrap(err, "unable to classify")
+	}
+	// Rounded values are used for all classifications. To not do so risks introducing
+	// inconsistency between frontend, backend, and/or other reports. See BACK-3800 for
+	// details.
+	rounded := roundToEvenWithDecimalPlaces(normalized, 1)
```
darinkrauss, line 77: Hmm, if you are going to use the datum units (above `g.Units`), then I don't think you should hard-code the `1` here (which is, by extension, hard-coding mmol/L).

data/types/blood/glucose/glucose.go
```diff
@@ -33,3 +36,75 @@ func (g *Glucose) Normalize(normalizer data.Normalizer) {
...
+	for _, threshold := range c {
+		if threshold.Includes(rounded) {
+			return threshold.Name, nil
+		}
+	}
+	// Ensure your highest threshold has a value like math.MaxFloat64 to avoid this.
+	return ClassificationInvalid, errors.Newf("unable to classify value: %v", *g)
```
darinkrauss, line 84: To remove this edge case (and remove the `math.MaxFloat64`) can we just remove the last `classificationThreshold` below and assume that if it isn't yet classified then in must be `ExtremelyHigh`?

summary/types/glucose.go
```diff
@@ -237,7 +244,9 @@ func (b *GlucoseBucket) Update(r data.Datum, lastData *time.Time) (bool, error)
 		return false, nil
 	}
 
-	b.GlucoseRanges.Update(record)
+	if err := b.GlucoseRanges.Update(record); err != nil {
+		return false, err
```
darinkrauss, line 248: Does this mean that if a single value has an error then it will cause the entire summary to fail?

data/types/blood/glucose/glucose.go
```diff
@@ -33,3 +36,75 @@ func (g *Glucose) Normalize(normalizer data.Normalizer) {
...
+// TidepoolADAClassificationThresholdsMmolL for classifying glucose values.
+//
+// All values are normalized to MmolL before classification.
+//
+// In addition to the standard ADA ranges, the Tidepool-specifiic "extremely high" range is
+// added.
+//
+// It is the author's responsibility to ensure the thresholds remain sorted from smallest to
+// largest.
+var TidepoolADAClassificationThresholdsMmolL = Classifier([]classificationThreshold{
```
darinkrauss, line 96: Hmm, including the `Name` in the `classificationThreshold` could lead to some future errors. For example, not realizing that the order was important, someone could reorder these and cause havoc. Consider a map and explicitly classifying in the correct order (i.e. [VeryLow, Low, InRange, High, VeryHigh] and anything else is ExtremelyHigh)?

data/types/blood/glucose/glucose.go
```diff
@@ -33,3 +36,75 @@ func (g *Glucose) Normalize(normalizer data.Normalizer) {
...
+	{Name: VeryLow, Value: 3, Inclusive: false},                    // Source: https://doi.org/10.2337/dc24-S006
+	{Name: Low, Value: 3.9, Inclusive: false},                      // Source: https://doi.org/10.2337/dc24-S006
+	{Name: OnTarget, Value: 10, Inclusive: true},                   // Source: https://doi.org/10.2337/dc24-S006
```
darinkrauss, line 99: Will we ever have different values for `Inclusive` for a specific classifier or will the classification itself determine the inclusiveness. That is, is `High` always inclusive, or can I have a classifier where it is not?

data/types/blood/glucose/glucose.go
```diff
@@ -33,3 +36,75 @@ func (g *Glucose) Normalize(normalizer data.Normalizer) {
 		g.Value = dataBloodGlucose.NormalizeValueForUnits(g.Value, units)
 	}
 }
+
+func roundToEvenWithDecimalPlaces(v float64, decimals int) float64 {
+	if decimals < 1 {
+		return math.RoundToEven(v)
+	}
+	coef := math.Pow(10, float64(decimals))
```
darinkrauss, line 44: [nit] `math.Pow10`?

data/types/blood/glucose/glucose.go
```diff
@@ -33,3 +36,75 @@ func (g *Glucose) Normalize(normalizer data.Normalizer) {
 		g.Value = dataBloodGlucose.NormalizeValueForUnits(g.Value, units)
 	}
 }
+
+func roundToEvenWithDecimalPlaces(v float64, decimals int) float64 {
+	if decimals < 1 {
```
darinkrauss, line 41: Consider just allowing 0 and negative numbers? The math still works out just fine.

data/types/blood/glucose/glucose.go
```diff
@@ -33,3 +36,75 @@ func (g *Glucose) Normalize(normalizer data.Normalizer) {
...
+		return math.RoundToEven(v)
+	}
+	coef := math.Pow(10, float64(decimals))
+	return math.RoundToEven(v*coef) / coef
+}
+
+type Classification string
+
+const (
+	ClassificationInvalid Classification = "invalid"
```
darinkrauss, line 51: Can we get rid of this edge case?

data/types/blood/glucose/glucose.go
darinkrauss: Oh, and if a map doesn't include a classification (e.g. 'Low') skip it and go to the next one (e.g. 'In Range').

data/types/blood/glucose/glucose.go
ewollesen: > Will we ever have different values for `Inclusive` for a specific classifier or will the classification itself determine the inclusiveness. That is, is `High` always inclusive, or can I have a classifier where it is not?

I don't know. I know there are "custom" ranges coming down the pipe, which is why I've kept the possibility.

data/types/blood/glucose/glucose.go
ewollesen: Nope. Go's type system isn't expressive enough.

data/types/blood/glucose/glucose.go
ewollesen: > Hmm, if you are going to use the datum units (above `g.Units`), then I don't think you should hard-code the `1` here (which is, by extension, hard-coding mmol/L).

I don't understand why not? NormalizeValueForUnits can only return units in mmol/L. It doesn't return any other units.

data/types/blood/glucose/glucose.go
```diff
@@ -33,3 +36,75 @@ func (g *Glucose) Normalize(normalizer data.Normalizer) {
...
+
+	VeryLow       = "very low"
+	Low           = "low"
+	OnTarget      = "on target"
```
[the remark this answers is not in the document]
darinkrauss: Okay. For future reference then, if `InRange` is not desired, then perhaps just `Target` or `TargetRange` (as I've never heard "on target" used in relation to Diabetes).

data/types/blood/glucose/glucose.go
darinkrauss: Ah, okay, makes sense.

data/types/blood/glucose/glucose.go
ewollesen: I'll take a look at what I can do. Using a map doesn't prevent errors (the user could miss a key, or leave a key out of the explicit ordering), and I don't wanna overcomplicate it too much.

If someone doesn't read the comment *and* misses the fact that it's using an ordered data structure *and* ignores all the unit tests *and* gets the change past code review... Forgive me, but that's on them, don't you think? There are limits to what I can prevent. I feel like I've taken reasonable steps.

summary/types/glucose.go
ewollesen: My fix for this is somewhat ... awkward maybe? I welcome other ideas, but I think this is going to all be changing soonish with the introduction of alternate ranges, so I'm not sure it's worth spending a ton of time on.

data/types/blood/glucose/glucose.go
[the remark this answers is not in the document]
ewollesen: I was in the area making changes, so I renamed "on target" => "in range".

data/types/blood/glucose/glucose.go
ewollesen: I've applied a sort to the thresholds. I dunno that its worth it, but... it's... something. Let me know what you think.

PR #873, opened 2025-06-26
fix GetBucketsRange query with both start and end time

ewollesen:
Previously, if both a start and end time were present, the end time
would overwrite the start time in the query filter, causing more data
to be retrieved than expected.

This showed up in EHR Reports where a report over the past 30 days
would indicate that 61 days of data were in scope.

BACK-3777


tjotala (approved):
LGTM!

PR #891, opened 2025-08-15
simplify consent record creation and updating

ewollesen:
Having separate types to juggle seemed overly complicated to me, so I wanted to try a simpler approach. I feel like this is an improvement, but maybe there are things I didn't see...?

### DIFF
--- a/consent/consent_record.go
+++ b/consent/consent_record.go
@@ -2,19 +2,14 @@ package consent
 
 import (
 	"context"
-	"encoding/json"
 	"time"
 
 	storeStructuredMongo "github.com/tidepool-org/platform/store/structured/mongo"
 
 	"github.com/tidepool-org/platform/auth"
-	"github.com/tidepool-org/platform/errors"
-	"github.com/tidepool-org/platform/id"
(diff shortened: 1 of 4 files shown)


consent/consent_record.go
```diff
@@ -162,7 +120,6 @@ func (r *RecordCreate) Parse(parser structure.ObjectParser) {
 
 func (r *RecordCreate) Validate(validator structure.Validator) {
 	validator.String("ageGroup", structure.ValueAsString(&r.AgeGroup)).OneOf(structure.ValuesAsStringArray(AgeGroups())...)
-	validator.Time("createdTime", &r.CreatedTime).NotZero().BeforeNow(time.Second)
```
ewollesen, line 165: This doesn't come from the user, so doesn't need to be validated here.

I've moved responsibility for this to the repository, which checks when creating if it exists, and if not, sets it to time.Now().

toddkazakov: This validator makes sure the model that's inserted in the database is valid right before it's persisted.

PR #918, opened 2026-03-17
[BACK-3013] Log work system email notifications event history. 

lostlevels:
Adds a log of email notification related events. Used by related script to send, and avoid re-sending, to existing users w/ device issues / expired connection requests.


notifications/logs/logs.go
```diff
@@ -0,0 +1,51 @@
...
+	NotificationEmailError        NotificationEventType = "email-error"
+	NotificationEmailSent         NotificationEventType = "email-sent"
+)
+
+type Logger interface {
+	Add(ctx context.Context, entry LogEntry) error
+	List(ctx context.Context, filter Filter) ([]LogEntry, error)
+}
+
+type Filter struct {
```
darinkrauss: Consider `Validate` function for `Filter`.

darinkrauss (changes requested):
First pass with a number of questions/comments.

notifications/history/history.go
darinkrauss: Is having a separate type worth it here? You'll have to map back and forth to `string`. OTOH, it will help enforce type safety.

Another choice is to make it just a type alias. So:
```
type NotificationEventType = string
```
You'll lose the type safety, but it will be just a string.

notifications/work/claims/processor.go
```diff
@@ -127,48 +126,44 @@ func (p *processor) Process(ctx context.Context, wrk *work.Work, updater work.Pr
...
 		ProcessorType: Type,
 		Email:         pointer.DefaultString(patient.Email, ""),
 		GroupID:       pointer.DefaultString(wrk.GroupID, ""),
-		EventType:     logs.NotificationEmailSent,
+		EventType:     history.NotificationEmailSent,
 		UserID:        data.UserID,
-		CreatedAt:     time.Now(),
 	}
-	_ = p.dependencies.Logs.Add(ctx, entry)
+	_ = p.dependencies.Recorder.Create(ctx, entry)
```
darinkrauss: Please do not bury the error here or we won't have any idea anything failed. (I don't see any error logging in the recorder itself.) Either log it here or mark the work item as Failing (if you want to retry) or Failed (no retry).

notifications/work/connections/issues/processor.go
```diff
@@ -143,28 +141,26 @@ func (p *processor) Process(ctx context.Context, wrk *work.Work, updater work.Pr
...
 		ProcessorType: Type,
 		GroupID:       pointer.DefaultString(wrk.GroupID, ""),
 		DataSourceID:  data.DataSourceID,
-		EventType:     logs.NotificationAttempted,
+		EventType:     history.NotificationAttempted,
 		UserID:        data.UserID,
-		CreatedAt:     time.Now(),
 	}
-	_ = p.dependencies.Logs.Add(ctx, entry)
+	_ = p.dependencies.Recorder.Create(ctx, entry)
```
darinkrauss: Please do not bury the error here or we won't have any idea anything failed. (I don't see any error logging in the recorder itself.) Either log it here or mark the work item as Failing (if you want to retry) or Failed (no retry).

notifications/work/connections/requests/processor.go
```diff
@@ -174,36 +172,33 @@ func (p *processor) Process(ctx context.Context, wrk *work.Work, updater work.Pr
...
 		ProcessorType: Type,
 		Email:         pointer.DefaultString(user.Username, ""),
 		GroupID:       pointer.DefaultString(wrk.GroupID, ""),
-		EventType:     logs.NotificationEmailSent,
+		EventType:     history.NotificationEmailSent,
 		UserID:        data.UserID,
-		CreatedAt:     time.Now(),
 	}
-	_ = p.dependencies.Logs.Add(ctx, entry)
+	_ = p.dependencies.Recorder.Create(ctx, entry)
```
darinkrauss: Please do not bury the error here or we won't have any idea anything failed. (I don't see any error logging in the recorder itself.) Either log it here or mark the work item as Failing (if you want to retry) or Failed (no retry).

darinkrauss (changes requested):
A number of comments and a question or two. Does need some changes, mostly handling the error from `Recorder.Create`.

darinkrauss (dismissed):
I'll approve, but if the List API has the potential to return many documents (thousands+), then this PR needs changes to support a pagination parameter and a stable sort on the Mongo query.

lostlevels:
> I'll approve, but if the List API has the potential to return many documents (thousands+), then this PR needs changes to support a pagination parameter and a stable sort on the Mongo query.

Addressed.

darinkrauss (dismissed):
LGTM!

lostlevels:
Force pushed to squash last 7 commits to make merging in master easier.

lostlevels:
@darinkrauss Latest commit always adds metadata to notification events history and forgot to add notification history event for device issues.

lostlevels:
Last commit uses the work item's metadata since that's serialized already w/ correct capitalization.

notifications/history/history.go
```diff
@@ -62,6 +62,7 @@ type Entry struct {
 	Email         string    `bson:"email,omitempty"`
 	GroupID       string    `bson:"groupId,omitempty"`
 	Metadata      bson.M    `bson:"metadata,omitempty"`
+	Error         error     `bson:"error,omitempty"`
```
darinkrauss, line 65: If you capture the error within an `errors.Serializable` it will capture the full  internal error context and information. So:
```
	Error         errors.Serializable     `bson:"error,omitempty"`
```

darinkrauss (dismissed):
LGTM! One comment about serializing the error.

darinkrauss (approved):
LGTM!

PR #926, opened 2026-03-24
Add Oura webhook subscriptions

darinkrauss:
- Add Oura webhook subscriptions
- Add Oura partner endpoints
- Add webhook work to periodically check and renew subscriptions
- Add custom error parser for Oura response with status 422
- Add UpdateSubscription struct and function
- Add OuraClient to data service context
- Add bson tags throughout in case struct is captured in persisted error
- Use time.ParseInLocation with default location time.UTC
- Add various Oura data validations
- Fix bugs
- Add and update tests
- https://tidepool.atlassian.net/browse/BACK-4027
- https://tidepool.atlassian.net/browse/BACK-4028
- https://tidepool.atlassian.net/browse/BACK-4029

### DIFF
--- a/auth/provider_session.go
+++ b/auth/provider_session.go
@@ -36,10 +36,10 @@ type ProviderSessionClient interface {
 }
 
 type ProviderSessionFilter struct {
-	UserID     *string `json:"userId,omitempty"`
-	Type       *string `json:"type,omitempty"`
-	Name       *string `json:"name,omitempty"`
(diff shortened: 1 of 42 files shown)


toddkazakov (commented):
What is reason for adding bson struct tags to structs that are not being serialized as bson anywhere in the code base except in tests?

Copilot comments seems valid and good to be addressed. Everything else looks ok.

toddkazakov (approved):

PR #928, opened 2026-03-26
Update mock controller usage

darinkrauss:


auth/client/client_test.go
```diff
@@ -49,10 +49,6 @@ var _ = Describe("Client", func() {
 		ctx = auth.NewContextWithServerSessionTokenProvider(log.NewContextWithLogger(context.Background(), logTest.NewLogger()), serverSessionTokenProvider)
 	})
 
-	AfterEach(func() {
```
toddkazakov, line 52: TIL: you don't need to call Finish()

toddkazakov (approved):

PR #935, opened 2026-04-14
Mark surveys as inelegible if validation of consent record data fails

toddkazakov:


oura/jotform/processor.go
darinkrauss: Is this comment correct? It implies that the `SendEvent` would only be called to indicate that the sizing kit was delivered (or at least the discount code created), but if the `OuraEligibilitySurveyEligible` is `false`, which prevents `CreateDiscountCode` from being invoked, this still gets invoked.

From what I can see, the main difference with this commit change is that if the survey doesn't `Validate` then it no longer errors and exits, but instead just warns and skips the `CreateDiscountCode` call, but everything else completes normally.

What am I missing?

darinkrauss (commented):
Just one question/comment for clarification.

oura/jotform/processor.go
toddkazakov: This comment is not correct. We're sending a "oura_eligibility_survey_completed` event - I'll fix this. 

With this change, If the survey data fails validation, we don't create consent records and discount code, but we still want to notify customer.io that the user is not eligible so they are taken out of the active campaign and we don't send reminders.

PR #944, opened 2026-05-07
require that tasks advance their availableTime

ewollesen:
During the investigation of a bug, this race condition was found where availableTime, if it were in the past would cause a task to be marked failed and therefore not run any more.

Now we track the previous availableTime for each task, and only when they don't advance is the task marked failed.

BACK-4452


ewollesen:
Closing in favor for #945

PR #945, opened 2026-05-08
don't error a task out when its availableTime is in the past

ewollesen:
During the investigation of a bug, this race condition was found where
availableTime, if it were in the past would cause a task to be marked
failed and therefore not run any more.

Now we use time.Now() when we find an availableTime in the past.

BACK-4452


toddkazakov (approved):

darinkrauss (approved):
LGTM!

PR #951, opened 2026-05-18
Add oura periodic and personal work processors

darinkrauss:
- Add oura periodic and personal work processors
- Update tests
- https://tidepool.atlassian.net/browse/BACK-4443
- https://tidepool.atlassian.net/browse/BACK-4447

### DIFF
--- a/oura/data/work/periodic/processor.go
+++ b/oura/data/work/periodic/processor.go
@@ -1,0 +1,275 @@
+package periodic
+
+import (
+	"context"
+	"time"
+
+	providerSessionWork "github.com/tidepool-org/platform/auth/providersession/work"
+	"github.com/tidepool-org/platform/compress"
+	dataRaw "github.com/tidepool-org/platform/data/raw"
+	dataRawWork "github.com/tidepool-org/platform/data/raw/work"
+	dataSourceWork "github.com/tidepool-org/platform/data/source/work"
+	dataWork "github.com/tidepool-org/platform/data/work"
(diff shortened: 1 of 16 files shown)


oura/data/work/periodic/processor.go
```diff
@@ -0,0 +1,275 @@
...
+	"github.com/tidepool-org/platform/pointer"
+	"github.com/tidepool-org/platform/structure"
+	structureValidator "github.com/tidepool-org/platform/structure/validator"
+	"github.com/tidepool-org/platform/times"
+	"github.com/tidepool-org/platform/work"
+	workBase "github.com/tidepool-org/platform/work/base"
+)
+
+const (
+	PendingAvailableDuration   = 12 * time.Hour // Data returned is for previous 24 hours, using 12 hours ensures we do not miss data
```
toddkazakov, line 30: What will happen if a task is in failing state for more than 24 hours?

toddkazakov (approved):
Nothing blocking, just one question

oura/data/work/periodic/processor.go
darinkrauss: We lose the opportunity to ingest the data beyond 24 hours old. Unfortunately, that's just the way the Oura API works for these two data types (heartrate, ring_battery_level). The API only ever returns the last 24 hours from the time of the API call.

The reason I use 12 hours here is to avoid the possibility of a delay or jitter causing the API to fire after 24 hours plus. This could probably be 23 hours and yield the same result.

PR #955, opened 2026-05-26
[BACK-3938] Add has data method to data client.

lostlevels:
Add HasAnyData for platform-data clients and corresponding route for use by clinic-worker to query for before potentially deleting a user.

### DIFF
--- a/data/service/api/v1/users_data_exists.go
+++ b/data/service/api/v1/users_data_exists.go
@@ -1,0 +1,38 @@
+package v1
+
+import (
+	"net/http"
+
+	dataService "github.com/tidepool-org/platform/data/service"
+	"github.com/tidepool-org/platform/request"
+)
+
+func GetHasAnyData(dataServiceContext dataService.Context) {
+	res := dataServiceContext.Response()
+	req := dataServiceContext.Request()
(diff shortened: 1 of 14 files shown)


data/service/api/v1/users_data_exists.go
```diff
@@ -0,0 +1,38 @@
+package v1
+
+import (
+	"net/http"
+
+	dataService "github.com/tidepool-org/platform/data/service"
+	"github.com/tidepool-org/platform/request"
+)
+
+func GetHasAnyData(dataServiceContext dataService.Context) {
```
toddkazakov, line 10: I don't think adding a new endpoint is required. The common denominator is a dataset - if the user doesn't have a data set, they don't have any data. To check whether a user has an existing data set, you can use the existing `/v1/users/:userId/datasets` endpoint

toddkazakov (changes requested):
As a whole, I don't think this PR is required. You can use the existing `/v1/users/:userId/datasets` endpoint to check if custodial users have data. If they have a dataset it was either created by an upload in the clinic (using Tidepool Uploader) or by explicitly linking a cloud provider.

PR #959, opened 2026-06-26
use go tool for mockgen

ewollesen:
This ensures we're using the same version across the team, which previously wasn't the case (as evidenced by errors running make generate).


darinkrauss (approved):
LGTM, but *please* do not merge until after the Oura work is merged.

ewollesen:
Oura updates obliterated this. I'll back this change into my upcoming device issues dashboard work.

PR #962, opened 2026-07-13
fix for sparse data showing up in the wrong summaries

ewollesen:
The change fixes a bucketing bug in the newest-first summary walk. Previously the period-boundary check was a single if, so each hourly bucket could advance nextStopPoint by at most one. When a large time gap sits between two buckets, one bucket can be past several period boundaries (1d/7d/14d/30d) at once, but only one got closed off, so the gapped bucket leaked into shorter periods it doesn't belong to.

BACK-4510

BACK-4525 is also included in this PR for ease of QA testing

### DIFF
--- a/summary/types/glucose_test.go
+++ b/summary/types/glucose_test.go
@@ -1388,6 +1388,19 @@ var _ = Describe("Glucose", func() {
 				Expect(s).To(BeEmpty())
 			})
 
+			It("handles cursors that fail mid-iteration", func() {
+				s := GlucosePeriods{}
+				s.Init()
+
+				buckets := CreateGlucoseBuckets(bucketTime, 24, 1, true)
+				bucketsCursor, err := mongo.NewCursorFromDocuments(
+					ConvertToIntArray(buckets), fmt.Errorf("batch fetch failed"), nil)
+				Expect(err).ToNot(HaveOccurred())
+
+				err = s.Update(ctx, bucketsCursor)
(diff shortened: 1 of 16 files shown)


summary/types/glucose_test.go
```diff
@@ -1484,6 +1484,35 @@ var _ = Describe("Glucose", func() {
...
+				old := CreateGlucoseBuckets(oldStart, 1, 1, true)
+				buckets := append(recent, old...)
+
+				bucketsCursor, err := mongo.NewCursorFromDocuments(ConvertToIntArray(buckets), nil, nil)
+				Expect(err).ToNot(HaveOccurred())
+
+				err = s.Update(ctx, bucketsCursor)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(s["1d"].Total.Records).To(Equal(24))
```
toddkazakov, line 1546: Glucose summares calculate offset periods as well. Since that code fragment that deals with the stop points is duplicated we should probably add some tests too.

toddkazakov (changes requested):

summary/types/glucose_test.go
ewollesen: @toddkazakov is the newly added test what you had in mind?

toddkazakov (dismissed):

toddkazakov (dismissed):

toddkazakov (approved):

PR #963, opened 2026-07-14
upgrade go-google.golang.org/grpc to avoid a CVE

ewollesen:
The CVE only affects GRPC servers (we use GRPC as a client), but I've
upgraded it anyway, as it's a very low risk operation, and this way we
won't have to worry about it being flagged again.

BACK-4329


darinkrauss (approved):
LGTM!

PR #964, opened 2026-07-20
Refactor task queue to resolve task-related failures

darinkrauss:
- Refactor task queue to resolve task-related failures
- Remove priority and expiration time from task as unused
- Add task timeout watchdog
- Add name to queue for debugging purposes
- Add revision condition to task Client and Repository API
- Set default available time for a new task to now
- Update runner deadline to return duration, not time
- Add Prometheus metrics to Dexcom client
- Add Prometheus metrics to task queue
- Add serialization mutex to test logger
- Require OAuth client to specify configured http client
- Capture response metrics on all outgoing OAuth client requests
- Add Prometheus helpers for outgoing clients
- Add common CloseCursor function to properly close Mongo cursors and log errors
- Add and update tests
- Fix typos
- https://tidepool.atlassian.net/browse/BACK-4523
- https://tidepool.atlassian.net/browse/BACK-4553


darinkrauss:
### Code review

Found 1 issue:

1. `AvailableTime` is never cleared when a task completes or fails, leaving a stale value on terminal tasks in the database. The previous queue code explicitly set `tsk.AvailableTime = nil` on dispatch and before persisting non-pending tasks, but in the refactor: `computeState` has an empty case for `TaskStateFailed`/`TaskStateCompleted`, `completeTask` passes the stale pre-run `AvailableTime` into the `TaskUpdate`, and `parseUpdate` has no unset branch for `availableTime` (unlike `Data`/`Error`, which use non-nil wrapper semantics precisely so they can be unset). Since `NewTask` always sets a non-nil `AvailableTime`, every completed/failed task will retain it. Notably, `dexcom_analyze` in this same PR still flags `Issue_Task_With_State_Failed_AvailableTime_Present` as an anomaly, so this state will trip your own diagnostic tooling.

https://github.com/tidepool-org/platform/blob/9d0e3b15d502e07a550e86f2c3b8d238145cba78/task/queue/queue.go#L631-L638

https://github.com/tidepool-org/platform/blob/9d0e3b15d502e07a550e86f2c3b8d238145cba78/task/queue/queue.go#L663-L665

https://github.com/tidepool-org/platform/blob/9d0e3b15d502e07a550e86f2c3b8d238145cba78/task/store/mongo/mongo.go#L607-L611

https://github.com/tidepool-org/platform/blob/9d0e3b15d502e07a550e86f2c3b8d238145cba78/services/tools/dexcom_analyze/dexcom_analyze.go#L93-L96

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

darinkrauss:
Resolve code review issue above.

toddkazakov (dismissed):

auth/service/api/v1/metrics.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/ant0ine/go-json-rest/rest"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+func (r *Router) MetricsRoutes() []*rest.Route {
+	return []*rest.Route{
+		rest.Get("/v1/metrics", r.PrometheusMetrics),
+	}
+}
```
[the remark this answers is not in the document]
darinkrauss: Not an issue. This path is not routed external to k8.

dexcom/fetch/runner.go
```diff
@@ -170,21 +172,29 @@ func NewTaskRunner(provider Provider, tsk *task.Task) (*TaskRunner, error) {
...
 	t.task.ClearError()
-	if err := t.run(); err == nil {
-		t.rescheduleTask()
-	} else if !t.task.HasError() {
-		t.rescheduleTaskWithResourceError(err)
+
+	if err := t.run(); err != nil {
+		t.task.AppendError(err)
+	}
+
```
toddkazakov, line 182: Claude flagged this - should there be a check here that returns early without executing the code below if the error returned above is `queue.ErrClaimLost`?

dexcom/fetch/runner.go
```diff
@@ -202,24 +212,23 @@ func (t *TaskRunner) run() error {
 	if err := t.fetchSinceLatestDataTime(); err != nil {
 		return err
 	}
-	if err := t.updateDataSourceWithLastImportTime(); err != nil {
```
toddkazakov, line 205: Partial fetch no longer updates the last import time of the data source. Just wanted to make sure this is intentional.

toddkazakov (dismissed):
Left two inline comments for your consideration

dexcom/fetch/runner.go
darinkrauss: Yes, this is missing. I fixed it.

dexcom/fetch/runner.go
darinkrauss: Yes, intentional. Only under rare circumstances would this even happen, and updating it **without** ingesting data and **without** finishing the import is somewhat misleading. Either way, going to leave it as.is.

darinkrauss:
@toddkazakov Just need a final review on the last two commits I just added - https://github.com/tidepool-org/platform/pull/964/changes/ca3d2b0faffea9bc06408065a21c6af7c306c0dd..355b240bdbcdde5b3dc1630f40baa69e747ac385.

toddkazakov (approved):

PR #965, opened 2026-07-28
Add checks for task claims that have been lost

darinkrauss:
- Add checks for task claims that have been lost
- Update Dexcom task runner to correctly update last import time
- Update Dexcom task runner to correctly reschedule
- Fix delay constants
- Update comments
- Update tests

### DIFF
--- a/env.sh
+++ b/env.sh
@@ -1,6 +1,8 @@
+# shellcheck shell=sh
+
 export TIDEPOOL_ENV="local"
 export TIDEPOOL_LOGGER_LEVEL="debug"
-export TIDEPOOL_STORE_ADDRESSES="localhost"
+export TIDEPOOL_STORE_ADDRESSES="127.0.0.1"
 export TIDEPOOL_STORE_DATABASE="tidepool"
 export TIDEPOOL_STORE_TLS="false"
 export TIDEPOOL_SERVER_TLS="false"
--- a/task/queue/queue.go
+++ b/task/queue/queue.go
@@ -444,18 +444,19 @@ func (q *Queue) runTask(ctx context.Context, tsk *task.Task) {
 		return
 	}
 
-	// If the runner left the task running, reconcile its state based on why the run ended.
+	// If the runner left the task running, reconcile its state based on why the run ended. A claim lost after the check
+	// above matches neither branch; the outcome is discarded during completion. Any other run left running is failed
+	// with a missing terminal state error during completion.
 	if tsk.State == task.TaskStateRunning {
-		switch {
-		case context.Cause(ctx) != nil:
+		if context.Cause(ctx) != nil {
 			// The parent (worker) context was canceled by shutdown; make the task available again for retry rather than
 			// treating the interruption as a completion.
 			tsk.RepeatAvailableAfter(0)
-		case context.Cause(runnerContext) != nil:
+		} else if cause := context.Cause(runnerContext); errors.Is(cause, ErrRunnerTimeoutExceeded) {
 			// The runner exceeded its timeout but returned; record the cause so the failure is attributed to the
 			// timeout rather than the generic missing terminal state error.
 			lgr.Warn("Task runner exceeded timeout; task will be failed")
-			tsk.AppendError(context.Cause(runnerContext))
+			tsk.AppendError(cause)
 			RunnerTimeoutExceededTotal.WithLabelValues(runner.GetRunnerType(), "recovered").Inc()
 		}
 	}
--- a/task/queue/queue.go
+++ b/task/queue/queue.go
@@ -21,53 +21,68 @@ import (
 )
 
 const (
-	WorkersDefault      = 5
-	DelayDefault        = 1 * time.Minute
-	DelayInitialDefault = 1 * time.Minute
-	DelayUnstickDefault = 5 * time.Minute
+	// WorkersDefault is the default number of workers for the queue.
+	WorkersDefault = 5
 
-	// StopWaitTimeoutDefault bounds how long Stop waits for in-flight tasks to observe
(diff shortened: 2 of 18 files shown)


task/queue/queue.go
```diff
@@ -76,46 +91,60 @@ func (c *Config) Load(configReporter config.Reporter) error {
...
 		} else {
-			c.StopWaitTimeout = time.Duration(stopWaitTimeout) * time.Second
+			c.RunnerWatchdogGracePeriod = time.Duration(value) * time.Second
 		}
 	}
-	if runnerWatchdogGracePeriodString, err := configReporter.Get("runner_watchdog_grace_period"); err == nil {
-		if runnerWatchdogGracePeriod, parseErr := strconv.ParseInt(runnerWatchdogGracePeriodString, 10, 0); parseErr != nil {
-			return errors.New("runner watchdog grace period is invalid")
+	if valueString, err := configReporter.Get("unstick_tasks_delay"); err == nil {
+		if value, parseErr := strconv.ParseInt(valueString, 10, 0); parseErr != nil {
```
ewollesen, line 130: If the source you're reading from doesn't use durations, like here, a fun trick sometimes is to do this:

```go
value, err := time.ParseDuration(valueString + "s")
```

It just saves you a time.Duration(value), but I find it reads a little bit simpler and has fewer if-then indent-dedent, which it think makes it all a little more readable.

task/queue/queue.go
```diff
@@ -76,46 +91,60 @@ func (c *Config) Load(configReporter config.Reporter) error {
...
-			c.DelayUnstick = time.Duration(delayUnstick) * time.Second
+			c.MonitorTaskDelay = time.Duration(value) * time.Second
 		}
 	}
-	if stopWaitTimeoutString, err := configReporter.Get("stop_wait_timeout"); err == nil {
-		if stopWaitTimeout, parseErr := strconv.ParseInt(stopWaitTimeoutString, 10, 0); parseErr != nil {
-			return errors.New("stop wait timeout is invalid")
+	if valueString, err := configReporter.Get("runner_watchdog_grace_period"); err == nil {
+		if value, parseErr := strconv.ParseInt(valueString, 10, 0); parseErr != nil {
+			return errors.New("runner watchdog grace period is invalid")
```
ewollesen, line 124: for debugging, you might consider if you want to include the valueString in the error, so the log will show what the "bad" value was. (Several places in this change that this might be useful, but maybe you have a good reason for not including it)

task/queue/queue.go
```diff
@@ -126,28 +155,34 @@ func (c *Config) Validate() error {
...
+	}
 	return nil
 }
 
-// The Queue's fields are all immutable after New, except the lifecycle fields, which are
-// guarded by the lifecycle mutex, and workersAvailable, which is owned exclusively by the
-// manager goroutine. The workers and manager therefore read the channels and runners map
-// freely, without synchronization.
+// The Queue's fields are all immutable after New, except the lifecycle fields, which are guarded by the lifecycle
+// mutex, and workersAvailable, which is initialized by Start before the manager exists and thereafter owned exclusively
```
ewollesen:
```suggestion
// mutex, and workersAvailable, which is initialized by Start() before the manager exists and is thereafter owned exclusively
```

env.sh
```diff
@@ -1,6 +1,8 @@
+# shellcheck shell=sh
+
 export TIDEPOOL_ENV="local"
 export TIDEPOOL_LOGGER_LEVEL="debug"
-export TIDEPOOL_STORE_ADDRESSES="localhost"
+export TIDEPOOL_STORE_ADDRESSES="127.0.0.1"
```
ewollesen, line 5: Heh, cant say I haven't been bit by similar bugs.

task/queue/queue_test.go
```diff
@@ -480,34 +518,181 @@ var _ = Describe("Queue", func() {
...
+					defer func() { _ = recover() }()
+					lgr.AssertError("Unable to stop task; no running task matched the id and claim token")
+					return true
+				}, "5s", "100ms").To(BeTrue())
+
+				actualTask := test.Must(str.NewTaskRepository().GetTask(ctx, createdTask.ID, nil))
+				Expect(actualTask.State).To(Equal(task.TaskStateRunning))
+			})
+
+			It("does not complete a task was deleted while it was running", func() {
```
ewollesen:
```suggestion
			It("does not complete a task that was deleted while it was running", func() {
```

ewollesen (approved):

task/queue/queue.go
darinkrauss: Yea, I think for the "next version" we should stick to using full duration strings (e.g. "1s") for the configuration of durations. They have to be a string anyhow since they go through environment variables and using duration strings just makes it so much clearer.

task/queue/queue.go
darinkrauss: I suppose I could go either way. 

Two reasons why historically I avoided embedding the invalid values in error messages: 1) finding errors in logs (very weak, just requires a more complicated regex query), and 2) accidentally exposing something in the API or logs that shouldn't be, like secrets (also weak, since this is just a duration, but generalizing it is a moderate concern). Both weak arguments, in this case, I suppose.

For item 2, I "fear" habituation. That is, you start doing something by habit and, since it is relatively rare occurrence (like adding a new secret), fail to ask whether it should be done in every case. And, with the current state of AI code completion, I've found the AI repeats existing patterns found in the code, which is generally good. However, it is easy to see how a new secret might get accidentally exposed and not caught.

I 100% agree with the usefulness of knowing what the invalid value actually is for debugging purposes.

I'm not sure how to reconcile this "fear" with the usefulness. Ideas?

PR #966, opened 2026-07-28
fixes a small validation typo in the work system

ewollesen:
The validation of success time was accidentally looking at the
FailedTime field instead of SuccessTime.


darinkrauss (approved):
Fortunately, no current work item returns the success state (most are delete), but good catch!

PR #968, opened 2026-08-19
[BACK-4528] Calculate summaries using the work system

toddkazakov:


data/service/api/v1/datasets_update.go
```diff
@@ -93,13 +91,9 @@ func DataSetsUpdate(dataServiceContext dataService.Context) {
 			return
 		}
 
-		// create map of all types, this will create redundant summaries, but will be cleaned up upon processing
-		updatesSummary := make(map[string]struct{})
-		for _, typ := range types.AllSummaryTypes {
-			updatesSummary[typ] = struct{}{}
+		if err = dataWorkPostprocess.Enqueue(ctx, dataServiceContext.WorkClient(), *dataSet.UserID, dataWorkPostprocess.ReasonUploadCompleted); err != nil {
+			lgr.WithError(err).Error("Unable to report upload completed")
 		}
```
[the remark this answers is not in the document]
toddkazakov: @darinkrauss should we move the dataset update logic after the deduplication block? Is there a reason why we might not want this?

data/work/postprocess/enqueue_test.go
```diff
@@ -0,0 +1,145 @@
...
+	var controller *gomock.Controller
+	var workClient *workTest.MockClient
+	var ctx context.Context
+	var userID string
+	var id string
+
+	BeforeEach(func() {
+		controller = gomock.NewController(GinkgoT())
+		workClient = workTest.NewMockClient(controller)
+		ctx = log.NewContextWithLogger(context.Background(), logTest.NewLogger())
```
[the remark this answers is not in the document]
toddkazakov: This is a pretty standard pattern, ignoring.

data/service/api/v1/datasets_update.go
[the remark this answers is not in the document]
darinkrauss: Hmm, not sure. I have a vague memory that this was here to prevent a repeated request from causing issues. The way it is now, the second request will fail because the data set is already closed. If we switched the order, then both requests could go through the deduplication code at the same time, which might be bad.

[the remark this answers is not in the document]
toddkazakov: Leaving this as is - enqueue must run after deduplication

clinics/service_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	Context("DeletePatientSummary", func() {
+		var summaryID string
+
+		BeforeEach(func() {
+			responseStatusCode = http.StatusOK
+			summaryID = primitive.NewObjectID().Hex()
+		})
+
+		It("deletes the summary from every patient record holding it", func() {
```
ewollesen, line 128: This test doesn't test what it says it will, and should be renamed to reflect what it's testing.

clinics/service_test.go
```diff
@@ -0,0 +1,150 @@
...
+			Expect(json.Unmarshal(requestBody, decoded)).To(Succeed())
+			Expect(decoded.CgmStats).ToNot(BeNil())
+			Expect(decoded.CgmStats.Id).To(Equal(patientSummary.CgmStats.Id))
+			Expect(decoded.BgmStats).To(BeNil())
+		})
+
+		// The clinic service reports a user who is not a patient of any clinic as no change. Most
+		// users are not, so reporting that as a failure would fail the work of nearly every user.
+		It("returns no error when the user is not a patient of any clinic", func() {
+			responseStatusCode = http.StatusNoContent
```
ewollesen, line 104: FWIW, the clinic service responds with 200 OK, not 204 No Content:

https://github.com/tidepool-org/clinic/blob/master/api/patients.go#L295
https://tidepool.redocly.app/reference/clinic.v1/clinics/updatepatientsummary

Line 232 in clinic/service.go could also be simplified to match.

The ec.NoContent just refers to the body, the status code returned is that which is passed as its argument (in this case, http.StatusOK).

clinics/service_test.go
```diff
@@ -0,0 +1,150 @@
...
+			summaryID = primitive.NewObjectID().Hex()
+		})
+
+		It("deletes the summary from every patient record holding it", func() {
+			Expect(client.DeletePatientSummary(context.Background(), summaryID)).To(Succeed())
+			Expect(requestPath).To(Equal("/v1/summaries/" + summaryID + "/clinics"))
+		})
+
+		It("returns no error when no patient record holds the summary", func() {
+			responseStatusCode = http.StatusNoContent
```
ewollesen, line 134: This also doesn't return 204 NoContent, but rather 200 OK. 

https://github.com/tidepool-org/clinic/blob/master/api/patients.go#L295

data/work/postprocess/processor.go
```diff
@@ -0,0 +1,248 @@
...
+			return p.Failing(errors.Wrap(err, "unable to delete work"))
+		}
+		// The work changed since it was listed - log a warning since this shouldn't happen.
+		// Work items are processed serially and there's no modification path.
+		if deleted == nil {
+			log.LoggerFromContext(p.Context()).WithField("id", wrk.ID).Warn("work absorbed changed before it was deleted")
+		}
+	}
+
+	if len(absorbed) > 0 {
```
ewollesen, line 149: This check can be removed.

It was already performed on line 128, and absorbed isn't modified below there.

data/work/postprocess/processor.go
```diff
@@ -0,0 +1,248 @@
...
+
+	return nil
+}
+
+// deferredPendingBuilder defers work until a time decided during processing
+type deferredPendingBuilder struct {
+	availableTime time.Time
+}
+
+func (d *deferredPendingBuilder) ProcessingAvailableTime(ctx context.Context, wrk *work.Work, tm time.Time) time.Time {
```
ewollesen, line 243: No need for either ctx, nor wrk to be passed here.

work/store/structured/mongo/mongo.go
```diff
@@ -67,9 +67,73 @@ func (s *Store) EnsureIndexes() error {
...
+	}
+	update := bson.A{
+		bson.M{"$set": bson.M{
+			"state":       work.StateFailing,
+			"failingTime": now,
+			// Literal as the serialized error must be stored verbatim rather than evaluated as an expression
+			"failingError":      bson.M{"$literal": errors.NewSerializable(errors.New("processing timeout expired"))},
+			"failingRetryCount": bson.M{"$add": bson.A{bson.M{"$ifNull": bson.A{"$failingRetryCount", 0}}, 1}},
+			// Retry immediately as the work never had the opportunity to report its completion
+			"failingRetryTime": now,
```
ewollesen, line 113: Without a backoff, and no limit to retries, this seems like it could block processing indefinitely. Is that what we want here?

ewollesen (commented):
One of Claude's issues that I thought was interesting, but don't have the background to evaluate. Maybe you'd care to verify that this isn't going to be an issue?


> data/service/api/v1/work.go
>  ●  14 [api-design]     POST /v1/work accepts a raw work.Create from any server token and passes it straight to the store, so the external producer (jellyfish) must reproduce postprocess's internal invariants (Type, GroupID == SerialID == IDFromUserID(userID), metadata userId/reasons, processingTimeout) itself; a mismatch is accepted with 201 and only fails later in validateWork as a permanent failed item, and GetWork maps a malformed id to 500.

ewollesen (dismissed):
~Oh, and are the existing task-based runners disconnected?~ Yes, in the next PR.

Overall, a couple of questions and a few nits, nothing that I think should block progress.

data/work/postprocess/processor.go
toddkazakov: Required to satisfy the interface

toddkazakov:
> One of Claude's issues that I thought was interesting, but don't have the background to evaluate. Maybe you'd care to verify that this isn't going to be an issue?
> 
> > data/service/api/v1/work.go
> > ●  14 [api-design]     POST /v1/work accepts a raw work.Create from any server token and passes it straight to the store, so the external producer (jellyfish) must reproduce postprocess's internal invariants (Type, GroupID == SerialID == IDFromUserID(userID), metadata userId/reasons, processingTimeout) itself; a mismatch is accepted with 201 and only fails later in validateWork as a permanent failed item, and GetWork maps a malformed id to 500.

I couldn't think of a nice way to address this. `/v1/work` accepts work items of any kind. We can't have type specific validation without maintaining a registry or a look up table. It seemed to me that this is an established pattern where we fail the work if the metadata validation fails.

work/store/structured/mongo/mongo.go
toddkazakov: This code will only retry tasks which are orphaned and stuck in processing state. This shouldn't happen unless a pod is forcefully killed. On the next retry orphaned work items are supposed to be picked up by their processor and further retries obey the retry rules defined by the processor for the specific work type.

I think it's highly unlikely to be stuck in this orphaned retrying state. Maybe it's a good idea to have a prometheus counter so we have indication how often this happens.

ewollesen (approved):

PR #969, opened 2026-08-25
[BACK-4528] Retire the summary task runners and the outdated-flag mechanism

toddkazakov:
Single cleanup PR collapsing the previously planned PR2 + PR3 + the legacy field removal, stacked on #968.

**Task service**: deletes `summary/task` (update + migration runners), deregisters them, and replaces the `EnsureSummary*Task` defaults with `DeleteDeprecatedTasks`, which removes the `org.tidepool.summary.*` task rows on startup (kept at least one release).

**Data service**: removes the `GET /v1/summaries/{outdated,migratable}/*` routes and handlers. `POST /v1/summaries/{type}/{userId}` remains as a manual recalc lever: it now enqueues upload postprocess work (`DATA_ADDED`) and returns 202 instead of recalculating synchronously. Shrinks `Summarizer` (drops `SetOutdated`/`GetOutdatedUserIDs`/`GetMigratableUserIDs`), removes the corresponding store methods, the `tidepool_summary_queue_lag`/`queue_length` metrics, the data-client methods and panic stubs, and the two outdated index models (physical index drops stay in the ops runbook).

**Legacy fields**: removes `Dates.OutdatedSince`/`OutdatedReason`/`LastUpdatedReason`, the `OutdatedReason*` constants, `SetOutdated`/`SetNotOutdated`, and `types.OutdatedSummariesResponse`. Grep clearance: blip references the fields only in a Storybook fixture; jellyfish only as local variable names; clinic service carries them in its own API where they simply go empty. A schema-version mismatch now recreates the summary from nothing (previously `SetOutdated(SCHEMA_MIGRATION)` + reset), and the no-new-data path no longer writes (there is no flag left to clear).

**Deploy gates**: PR #968 and jellyfish deployed everywhere; `tidepool_summary_queue_length` at zero and staying there; deploy together with tidepool-org/clinic-worker#123, which removes the CDC EHR-sync trigger that read `lastUpdatedReason`.

Tests: 82 packages green across `summary/...`, `data/...`, `task/...`, `work/...`; the failing-cursor regression test is ported from `GetOutdatedUserIDs` to `ListMigratableUserIDs`.


summary/summary.go
```diff
@@ -161,8 +145,8 @@ func (gs *GlucoseSummarizer[PP, PB, P, B]) UpdateSummary(ctx context.Context, us
 		}
 
 		if userSummary.Config.SchemaVersion != types.SchemaVersion {
-			userSummary.SetOutdated(types.OutdatedReasonSchemaMigration)
-			userSummary.Dates.Reset()
+			// A summary calculated with an outdated schema is recreated from nothing
+			*userSummary = *types.Create[PP, PB](userId)
```
ewollesen, line 149: Claude points out that the existing summary's _id is lost. Create doesn't set one. Later on, the summary is upserted via userId/type, which means that the existing database record will be updated and still have the correct _id, however the summary in Go won't have the correct _id, it'll have 0.

I didn't follow the whole chain of calls to see if/how userSummary's _id is used from there, but it goes through quite a few other methods, so having an invalid _id seems like a time bomb waiting to explode.

Does it make sense to make sure that the existing userSummary's _id is still there? That's how the old code would have worked I believe, since it didn't recreate from scratch.

This might not have been caught, because the tests that covered this branch were removed.

data/service/api/v1/summary.go
```diff
@@ -130,7 +122,9 @@ func GetPatientsWithRealtimeData(dataServiceContext dataService.Context) {
 	responder.Data(http.StatusOK, response)
 }
 
-func UpdateSummary[PP types.PeriodsPt[P, PB, B], PB types.BucketDataPt[B], P types.Periods, B types.BucketData](dataServiceContext dataService.Context) {
+// UpdateSummary reports the data of the user as changed rather than recalculating synchronously,
+// which the retired task runners required. The work created recalculates every summary of the user.
```
ewollesen, line 126: Is there a migration plan to collapsed the three endpoints down to one, since all three do the same work?

ewollesen (changes requested):
Claude also pointed out that some metrics mentioned in the README are no longer collected. It would be good to update that.

Nothing major, but things that I feel like should be addressed.

Nice work!

summary/summary.go
toddkazakov: This is a good catch, thanks

data/service/api/v1/summary.go
toddkazakov: I'll collapse them here - no need to maintain 3 endpoints

summary/summary.go
toddkazakov: Fixed in #968

ewollesen (approved):

PR #970, opened 2026-08-26
[BACK-4442] [BACK-4528] Ignore not parsed fields when requesting data with the http client

toddkazakov:
`client.Client.RequestData` now decodes response bodies with `request.IgnoreNotParsed()`, so a service returning fields a client model does not parse — or gaining fields over time — no longer fails the request for every consumer of the shared http client.

This generalizes the fix first made for `user.Client.Get` (keycloak-migrated users gained `securityProfile`, which failed the strict parse and wedged work items depending on the user fetch). The user client spec asserting tolerance of unrecognized fields moves here to pin the behavior at the shared client.

Split out of #968, which will be rebased on top of this branch.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -141,7 +141,7 @@ func (c *Client) RequestDataWithHTTPClient(ctx context.Context, method string, u
 		return nil
 	}
 
-	return request.DecodeStream(ctx, structure.NewPointerSource(), body, responseBody)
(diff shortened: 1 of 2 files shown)


darinkrauss (approved):
Yes, that will work for any validation coming from a Client-based call, but won't cover any of the other validations. Otherwise, LGTM.

PR #971, opened 2026-09-01
fall back to server session token for recording metrics

ewollesen:
Some metrics aren't being recorded, due to auth issues, and they're
spamming the logs. By falling back to using a server session token,
the metrics should be successfully posted and the logs will be
quieter.

### DIFF
--- a/metric/client/client.go
+++ b/metric/client/client.go
@@ -5,6 +5,7 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/tidepool-org/platform/auth"
 	"github.com/tidepool-org/platform/errors"
 	"github.com/tidepool-org/platform/log"
 	"github.com/tidepool-org/platform/platform"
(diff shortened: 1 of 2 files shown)


metric/client/client.go
```diff
@@ -54,6 +55,16 @@ func (c *Client) RecordMetric(ctx context.Context, metric string, data ...map[st
 	var requestURL string
 	if details := request.GetAuthDetails(ctx); details.IsService() {
 		requestURL = c.client.ConstructURL("metrics", "server", c.name, metric)
+		if !details.HasToken() {
+			// Fall back to the server session token provider, if one is available
+			if provider := auth.ServerSessionTokenProviderFromContext(ctx); provider != nil {
+				serverSessionToken, err := provider.ServerSessionToken()
+				if err != nil {
+					return errors.Wrap(err, "unable to get server session token")
```
ewollesen, line 63: I'm not 100% sure it's the best choice to error here, but without this token, the RequestData call on line 72 will fail anyway, so it's kinda the same either way.

darinkrauss (approved):
🥳 Good call! This has been so annoying.

toddkazakov (approved):
LGTM

PR #974, opened 2026-09-17
Use tidepool-org/appattest for X.509 extended key usage.

lostlevels:
The base-d/appattest package has an issue verifying X.509 certificates due to missing extended key usages which causes attestation to fail.
This updates the imports to use [https://github.com/tidepool-org/appattest] that includes the key usages.


ewollesen:
Am I understanding correctly that the real functional change here is the change to a tidepool-maintained fork of appatest? This PR is 99% concerned with some additional logging?

If so, I question whether all the extra logging is really necessary? I imagine it was helpful while you were debugging this issue, but given that it seems to have the possibility of leaking sensitive information, I wonder if it's something we should really be committing?

PR #981, opened 2026-09-23
Generate unique manufacturers in the pump settings test helper

toddkazakov:
Pump settings validation requires unique manufacturers, but the test helper `NewManufacturers` drew independent random strings, so about 1 in 50,000 generated arrays contained a duplicate and the pump suite failed intermittently with `value is a duplicate` at `/manufacturers/N` (for example Travis build 8905, private plugins job).

Use the existing duplicate-free array generator instead. Verified with 200,000 draws producing no duplicates and five race-detector runs of the suite.


lostlevels (approved):

ewollesen (approved):
I wouldn't be surprised if there are a few other places that need this treatment.

Nice job!

